PR #6, opened 2026-02-12
Add .NET performance analysis skill and optimizing agent

artl93:
## .NET Performance Analysis Skill & Agent

Adds the **analyzing-dotnet-performance** skill and **optimizing-dotnet-performance** agent — a distilled, actionable set of ~50 .NET performance patterns designed to detect and fix anti-patterns in real codebases.

### Where this comes from

Compiled from [Stephen Toub's .NET performance blog series](https://gist.github.com/artl93/621797f3cd8a06ba5c6c3f18705b251e) spanning .NET Core 2.0 through .NET 10. The full reference was shrunk to fit a skill context window, organized into 7 topic-specific reference files loaded on-demand based on code signals.

### What's in the skill

| Reference Guide | Patterns | Covers |
|---|---|---|
| `critical-patterns.md` | 17 | Deadlocks, crashes, ReDoS, >10x regressions — always loaded |
| `async-patterns.md` | 5 | ValueTask, Channels, false sharing |
| `memory-and-strings.md` | 9 | Span\<T\>, stackalloc, UTF8 literals, string allocation chains |
| `collections-and-linq.md` | 9 | FrozenDictionary, CollectionsMarshal, params span |
| `io-and-serialization.md` | 6 | JSON source generators, async FileStream, static readonly |
| `regex-patterns.md` | 4 | GeneratedRegex, NonBacktracking, EnumerateMatches |
| `structural-patterns.md` | 1 | Sealed classes (absence-based detection) |

### Agent: Two-Pass Analysis

1. **Pass 1 (LLM-Only)**: Fast, intuition-driven review using the agent's own .NET knowledge — no skills loaded.
2. **Pass 2 (Skill-Based Deep Scan)**: Loads the skill for systematic pattern matching. Deduplicates against Pass 1 and reports only new discoveries.

### Key features

- **Progressive disclosure**: Critical patterns always loaded; topic files loaded based on code signals
- **Scan recipes**: grep-based detection with mandatory execution checklist
- **Verify-the-Inverse**: Absence patterns report ratios ("N of M classes are sealed")
- **Cross-file consistency**: Flags sibling files using un-optimized equivalents
- **Scale-based severity**: Same anti-pattern at 50+ sites escalates to codebase-wide issue
- **Common Pitfalls**: Guards against over-optimization (e.g., unsafe for micro-opts, LINQ removal outside hot paths)

### Real-world validation

Validated against [Humanizer](https://github.com/Humanizr/Humanizer): **+8.9% average speed, +5.1% memory reduction** across 81 BenchmarkDotNet benchmarks with zero functional regressions. [PR example](https://github.com/artl93/Humanizer/pull/1).

[Comparison across runs](https://gist.github.com/artl93/d9a608d33d8fc6b47a70a1fc9311a320) — Run "B" is Claude Opus with no skills, Run "H" is this agent + skill.

### Files

```
agents/optimizing-dotnet-performance.agent.md
skills/analyzing-dotnet-performance/
├── SKILL.md
└── references/
    ├── critical-patterns.md
    ├── async-patterns.md
    ├── collections-and-linq.md
    ├── io-and-serialization.md
    ├── memory-and-strings.md
    ├── regex-patterns.md
    └── structural-patterns.md
```

### DIFF
--- a/src/dotnet/skills/analyzing-dotnet-performance/SKILL.md
+++ b/src/dotnet/skills/analyzing-dotnet-performance/SKILL.md
@@ -70,7 +70,7 @@ For each relevant pattern category, run the detection recipes below. Report exac
 # Strings & memory
 grep -n '\.IndexOf(\"' FILE                    # Missing StringComparison
 grep -n '\.Substring(' FILE                    # Substring allocations
-grep -n '\.StartsWith\|\.EndsWith\|\.Contains' FILE  # Missing StringComparison
+grep -En '\.(StartsWith|EndsWith|Contains)\s*\(' FILE  # Missing StringComparison
 grep -n '\.ToLower()\|\.ToUpper()' FILE        # Culture-sensitive + allocation
 grep -n '\.Replace(' FILE                      # Chained Replace allocations
 grep -n 'params ' FILE                         # params array allocation
@@ -126,7 +126,7 @@ Assign each finding a severity:
 **Prioritization rules:**
 1. If the user identified hot-path code, elevate all findings in that code to their maximum severity
 2. If hot-path context is unknown, report 🔴 Critical findings unconditionally; report 🟡 Moderate findings with a note: _"Impactful if this code is on a hot path"_
-3. Never suggest micro-optimizations on code that is clearly not performance-sensitive (startup, configuration, one-time initialization)
+3. Never suggest micro-optimizations on code that is clearly not performance-sensitive
 
 **Scale-based severity escalation:**
 When the same pattern appears across many instances, escalate severity:
@@ -186,7 +186,7 @@ Before delivering results, verify:
 |---------|-----------------|
 | Flagging every `Dictionary` as needing `FrozenDictionary` | Only flag if the dictionary is never mutated after construction |
 | Suggesting `Span<T>` in async methods | Use `Memory<T>` in async code; `Span<T>` only in sync hot paths |
-| Reporting LINQ outside hot paths | Only flag LINQ in identified hot paths or tight loops; LINQ is fine in startup, config, and one-time code. Since .NET 7, LINQ Min/Max/Sum/Average are vectorized — blanket bans on LINQ are misguided |
+| Reporting LINQ outside hot paths | Only flag LINQ in identified hot paths or tight loops; LINQ is acceptable in code that runs infrequently. Since .NET 7, LINQ Min/Max/Sum/Average are vectorized — blanket bans on LINQ are misguided |
 | Suggesting `ConfigureAwait(false)` in app code | Only applicable in library code; not primarily a performance concern |
 | Recommending `ValueTask` everywhere | Only for hot paths with frequent synchronous completion |
 | Flagging `new HttpClient()` in DI services | Check if `IHttpClientFactory` is already in use |
--- a/src/dotnet/skills/analyzing-dotnet-performance/references/collections-and-linq.md
+++ b/src/dotnet/skills/analyzing-dotnet-performance/references/collections-and-linq.md
@@ -54,27 +54,6 @@ count++;
 
 **Impact: ~48% faster for lookup-and-update patterns (95µs → 49µs).**
 
-### Use CollectionsMarshal.AsSpan for Direct List\<T\> Access
-🟡 **CONSIDER** using `CollectionsMarshal.AsSpan` for direct span access over list internals in ultra-hot paths with benchmarked evidence | .NET 5+
-
-This is an advanced technique for specialized hot paths only. Do not apply broadly — it adds complexity and fragility (the span is invalidated by any mutation to the list). Only use when profiling confirms the enumerator or element-by-element access is a bottleneck.
-
-❌
-```csharp
-List<int> list = new(count);
-for (int i = 0; i < count; i++)
-    list.Add(source[i]);
-```
-✅
-```csharp
-List<int> list = new();
-CollectionsMarshal.SetCount(list, count);
-Span<int> span = CollectionsMarshal.AsSpan(list);
-source.AsSpan(0, count).CopyTo(span);
-```
-
-**Impact: Avoids per-element overhead and enumerator allocation. Enables vectorized operations on list contents.**
-
 ### Use Collection Expressions [] for Zero-Allocation Span Creation
 🟡 **DO** use collection expressions for `Span<T>` targets | C# 12 / .NET 8+
 
--- a/src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
+++ b/src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
@@ -123,7 +123,7 @@ int val = int.Parse(str.AsSpan(5, 3));
 ## Regular Expressions
 
 ### Use Source-Generated Regex [GeneratedRegex]
-🔴 **DO** | .NET 7+
+🔴 **ALWAYS** use `[GeneratedRegex]` for all static regex patterns | .NET 7+
 
 ❌
 ```csharp
@@ -135,7 +135,7 @@ private static readonly Regex s_re =
 [GeneratedRegex(@"\w+@\w+\.\w+")]
 private static partial Regex EmailRegex();
 ```
-**Impact: Near-zero startup; required for AOT/trimming scenarios.**
+**Impact: Always beneficial or neutral for static patterns — near-zero startup, better throughput, and required for AOT/trimming scenarios.**
 
 ### Avoid Nested Quantifiers (Catastrophic Backtracking)
 🔴 **AVOID** | .NET Core+
@@ -263,7 +263,7 @@ int pos = text.AsSpan().IndexOfAny(s_hex);
 
 ## Detection
 
-Scan recipes for critical anti-patterns. Run these and report exact counts.
+Scan recipes for critical anti-patterns. Run these and report exact counts of issues found in each case.
 
 ```bash
 # .IndexOf(string) without StringComparison (culture-aware, 2-3x slower)
@@ -272,6 +272,9 @@ grep -rn --include='*.cs' -E '\.IndexOf\("[^"]+"\)' --exclude-dir=bin --exclude-
 # .Substring( calls (allocates new string — consider AsSpan)
 grep -rn --include='*.cs' '\.Substring(' --exclude-dir=bin --exclude-dir=obj . | wc -l
 
-# .StartsWith/.EndsWith/.Contains without StringComparison (culture-aware, 2-3x slower)
-grep -rn --include='*.cs' -E '\.(StartsWith|EndsWith|Contains)\("[^"]+"\)' --exclude-dir=bin --exclude-dir=obj . | wc -l
+# .StartsWith/.EndsWith without StringComparison (culture-aware, 2-3x slower)
+grep -rn --include='*.cs' -E '\.(StartsWith|EndsWith)\("[^"]+"\)' --exclude-dir=bin --exclude-dir=obj . | wc -l
+
+# .Contains(string) without StringComparison — NOTE: will also match collection .Contains() calls; filter to string receivers
+grep -rn --include='*.cs' -E '\.Contains\("[^"]+"\)' --exclude-dir=bin --exclude-dir=obj . | wc -l
 ```
--- a/src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
+++ b/src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
@@ -1,7 +1,7 @@
 # Regex Patterns
 
 ### Choose the Right Regex Engine Mode
-🟡 **DO** select the regex engine based on your scenario's trade-offs | .NET 7+
+🟡 **DO** use `[GeneratedRegex]` for all static regex patterns, but never remove `NonBacktracking` if present | .NET 7+
 
 ❌
 ```csharp
@@ -17,7 +17,7 @@ var safe = new Regex(untrustedPattern, RegexOptions.NonBacktracking);
 var oneOff = new Regex("pattern");
 ```
 
-**Impact: Source generator gives best throughput with near-zero startup. NonBacktracking prevents O(2^N) worst case at the cost of ~128x slower best case.**
+**Impact: Source generator is always beneficial for static patterns. NonBacktracking prevents O(2^N) worst case — never remove it if present.**
 
 ### Use IsMatch When You Only Need a Boolean Result
 🟡 **DO** use `IsMatch` instead of `Match(...).Success` | .NET 7+
--- a/src/dotnet/skills/analyzing-dotnet-performance/references/structural-patterns.md
+++ b/src/dotnet/skills/analyzing-dotnet-performance/references/structural-patterns.md
@@ -11,7 +11,7 @@ Sealing lets the JIT devirtualize/inline virtual calls and use pointer compariso
 
 ```bash
 # Count unsealed (non-abstract, non-static) classes
-grep -rn --include='*.cs' -E '^\s*(public |internal |private )?(partial )?class ' --exclude-dir=bin --exclude-dir=obj . | grep -v 'sealed' | grep -v 'abstract' | grep -v 'static' | wc -l
+grep -rn --include='*.cs' -E '^\s*((public|internal|private|protected|file)\s+)?(partial\s+)?class ' --exclude-dir=bin --exclude-dir=obj . | grep -v 'sealed' | grep -v 'abstract' | grep -v 'static' | wc -l
 
 # Count already-sealed classes (verify the inverse)
 grep -rn --include='*.cs' 'sealed class' --exclude-dir=bin --exclude-dir=obj . | wc -l
--- a/src/dotnet/tests/analyzing-dotnet-performance/fixtures/inflector-and-regex.cs
+++ b/src/dotnet/tests/analyzing-dotnet-performance/fixtures/inflector-and-regex.cs
@@ -16,7 +16,7 @@ public static partial class CaseTransformExtensions
     [GeneratedRegex(@"(?<=[a-z])(?=[A-Z])")]
     private static partial Regex CamelCaseBoundaryGenerated();
 
-    [GeneratedRegex(@"[\p{Lu}]+(?=\p{Lu}[\p{Ll}}])|[\p{Lu}]?[\p{Ll}]+|\d+")]
+    [GeneratedRegex(@"[\p{Lu}]+(?=\p{Lu}[\p{Ll}])|[\p{Lu}]?[\p{Ll}]+|\d+")]
     private static partial Regex WordPartsRegexGenerated();
 
     private static Regex SeparatorRegex() => SeparatorRegexGenerated();
@@ -25,7 +25,7 @@ public static partial class CaseTransformExtensions
 #else
     private static readonly Regex SeparatorRegexField = new(@"[-\s]", RegexOptions.Compiled);
     private static readonly Regex CamelCaseBoundaryField = new(@"(?<=[a-z])(?=[A-Z])", RegexOptions.Compiled);
-    private static readonly Regex WordPartsRegexField = new(@"[\p{Lu}]+(?=\p{Lu}[\p{Ll}}])|[\p{Lu}]?[\p{Ll}]+|\d+", RegexOptions.Compiled);
+    private static readonly Regex WordPartsRegexField = new(@"[\p{Lu}]+(?=\p{Lu}[\p{Ll}])|[\p{Lu}]?[\p{Ll}]+|\d+", RegexOptions.Compiled);
 
     private static Regex SeparatorRegex() => SeparatorRegexField;
     private static Regex CamelCaseBoundary() => CamelCaseBoundaryField;
--- a/src/dotnet/tests/analyzing-dotnet-performance/fixtures/metric-numeral-extensions.cs
+++ b/src/dotnet/tests/analyzing-dotnet-performance/fixtures/metric-numeral-extensions.cs
@@ -121,7 +121,8 @@ public static class MetricNumeralExtensions
         var symbol = GetSymbolForExponent(metricExponent);
         if (symbol == null)
         {
-            return input.ToString("G");
+            var fallback = input.ToString("G");
+            return isNegative ? $"-{fallback}" : fallback;
         }
 
         var scaledValue = input / Math.Pow(10, metricExponent);
@@ -129,7 +130,7 @@ public static class MetricNumeralExtensions
 
         var suffix = formats.HasFlag(MetricFormat.UseSymbol)
             ? symbol.Value.ToString()
-            : $" {GetPrefixName(symbol.Value, formats)}";;
+            : $" {GetPrefixName(symbol.Value, formats)}";
 
         return isNegative ? $"-{valueStr}{suffix}" : $"{valueStr}{suffix}";
     }
--- a/src/dotnet/tests/analyzing-dotnet-performance/fixtures/number-converter-farsi-kurdish.cs
+++ b/src/dotnet/tests/analyzing-dotnet-performance/fixtures/number-converter-farsi-kurdish.cs
@@ -62,9 +62,12 @@ public class FarsiNumberToWordsConverter : INumberToWordsConverter
             { (long)Math.Pow(10, 2), n => FarsiHundredsMap[n] }
         };
 
+        // Use descending order explicitly — dictionary enumeration order is not guaranteed
+        var groupKeys = new long[] { (long)Math.Pow(10, 15), (long)Math.Pow(10, 12), (long)Math.Pow(10, 9), (long)Math.Pow(10, 6), (long)Math.Pow(10, 3), (long)Math.Pow(10, 2) };
+
         // Anti-pattern: new List<string>() without initial capacity
         var parts = new List<string>();
-        foreach (var group in farsiGroupsMap.Keys)
+        foreach (var group in groupKeys)
         {
             if (number / group > 0)
             {
@@ -140,8 +143,11 @@ public class CentralKurdishNumberToWordsConverter : INumberToWordsConverter
             { (long)Math.Pow(10, 2), n => KurdishHundredsMap[n] }
         };
 
+        // Use descending order explicitly — dictionary enumeration order is not guaranteed
+        var groupKeys = new long[] { (long)Math.Pow(10, 15), (long)Math.Pow(10, 12), (long)Math.Pow(10, 9), (long)Math.Pow(10, 6), (long)Math.Pow(10, 3), (long)Math.Pow(10, 2) };
+
         var parts = new List<string>();
-        foreach (var group in kurdishGroupsMap.Keys)
+        foreach (var group in groupKeys)
         {
             if (number / group > 0)
             {
--- a/src/dotnet/tests/analyzing-dotnet-performance/fixtures/number-converter-multi-locale.cs
+++ b/src/dotnet/tests/analyzing-dotnet-performance/fixtures/number-converter-multi-locale.cs
@@ -320,6 +320,7 @@ public class HungarianNumberToWordsConverter : INumberToWordsConverter
             return cached;
         }
 
+        var originalNumber = number;
         var word = "";
 
         if (number / 1_000_000 > 0)
@@ -361,7 +362,7 @@ public class HungarianNumberToWordsConverter : INumberToWordsConverter
         }
 
         var result = word.Trim();
-        _runtimeCache[number] = result;
+        _runtimeCache[originalNumber] = result;
         return result;
     }
 
--- a/src/dotnet/tests/analyzing-dotnet-performance/fixtures/truncation-and-span.cs
+++ b/src/dotnet/tests/analyzing-dotnet-performance/fixtures/truncation-and-span.cs
@@ -52,6 +52,13 @@ public class FixedLengthTruncator : ITruncator
 
         truncationString ??= string.Empty;
 
+        if (truncationString.Length >= length)
+        {
+            return truncateFrom == TruncateFrom.Left
+                ? truncationString[^length..]
+                : truncationString[..length];
+        }
+
         if (truncateFrom == TruncateFrom.Left)
         {
             // Correct: uses AsSpan for efficient concatenation


skills/analyzing-dotnet-performance/references/async-patterns.md
```diff
@@ -0,0 +1,166 @@
+# Async & Concurrency Patterns
+
+### ConfigureAwait(false) in Library Code
+🟡 **DO** use `ConfigureAwait(false)` on every `await` in library code | .NET Core+
```
stephentoub: I would remove this one. a) there are already analyzers for this and b) its not primarily a perf thing

skills/analyzing-dotnet-performance/references/async-patterns.md
```diff
@@ -0,0 +1,166 @@
...
+{
+    if (_bufferedCount > 0)
+        return new ValueTask<int>(ReadFromBuffer(buffer.Span));
+    return new ValueTask<int>(ReadAsyncCore(buffer));
+}
+```
+
+**Impact: Eliminates Task\<T\> allocation on synchronous completion — the struct stores results inline.**
+
+### Use Parallel.ForEachAsync for Async Parallel Processing
```
stephentoub: Remove this one. It's about convenience more than anything else.

skills/analyzing-dotnet-performance/references/async-patterns.md
```diff
@@ -0,0 +1,166 @@
...
+
+// Consumer
+await foreach (var item in channel.Reader.ReadAllAsync())
+    Process(item);
+```
+
+**Impact: ~25% faster, ~95% fewer GC collections vs manual approaches.**
+
+### Avoid Async Lambda with Action-Accepting Methods
+🟡 **AVOID** passing async lambdas to methods expecting `Action` | .NET Core+
```
stephentoub: This is about correctness, not perf

src/dotnet/skills/analyzing-dotnet-performance/references/collections-and-linq.md
```diff
@@ -0,0 +1,235 @@
...
+```
+✅
+```csharp
+ref int count = ref CollectionsMarshal.GetValueRefOrAddDefault(_counts, key, out _);
+count++;
+```
+
+**Impact: ~48% faster for lookup-and-update patterns (95µs → 49µs).**
+
+### Use CollectionsMarshal.AsSpan for Direct List\<T\> Access
```
stephentoub: This is going to lead to a monstrous amount of churn for little benefit. It _can_ make a meaningful difference, but only in really specialized hot paths. We either need to tweak this to be much more specific or remove it.

skills/analyzing-dotnet-performance/references/collections-and-linq.md
```diff
@@ -0,0 +1,235 @@
...
+```csharp
+var list = new List<int>();
+list.EnsureCapacity(10000);
+for (int i = 0; i < 10000; i++)
+    list.Add(i);
+```
+
+**Impact: Reduces reallocations and array copies during bulk operations.**
+
+### LINQ Is Fine Outside Hot Paths — Don't Over-Optimize
```
stephentoub: This is good advice, but I don't know what place it has in this skill... is it trying to counteract some other guidance in place in the skill? Switch existing open-coded loops over to using LINQ instead?

skills/analyzing-dotnet-performance/references/collections-and-linq.md
```diff
@@ -0,0 +1,235 @@
...
+public string Convert(long number)
+{
+    // ... use s_groupsMap
+}
+```
+
+**Impact: Eliminates collection + internal storage + closure allocations per call. For a Dictionary with N entries, saves ~N+3 allocations per invocation.**
+
+### Add Overloads to Avoid params Array Allocation
+🟡 **DO** add 1- and 2-argument overloads for methods that accept `params T[]` | .NET Core+
```
stephentoub: Could also use params span now

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
+# Critical .NET Performance Anti-Patterns
+
+24 patterns that cause deadlocks, crashes, order-of-magnitude regressions, or security vulnerabilities.
+
+## Async / Tasks
+
+### 1. Avoid async void
```
stephentoub: This is correctness rather than perf

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+int result = await SomeMethodAsync();
+```
+**Impact: Undefined behavior — silent data corruption or exceptions.**
+
+### 4. Don't Use lock with Async Code
+🔴 **AVOID** | .NET Core+
+
+❌
+```csharp
+lock (_sync) { await DoWorkAsync(); }
```
stephentoub: This won't even compile; we don't need AI for that

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+✅
+```csharp
+private readonly SemaphoreSlim _sem = new(1, 1);
+await _sem.WaitAsync();
+try { await DoWorkAsync(); } finally { _sem.Release(); }
+```
+**Impact: Compile error, or if bypassed, deadlocks and thread corruption.**
+
+### 5. Async All the Way Down
+🔴 **DO** | .NET Core+
```
stephentoub: How is this different from the "don't block" advice earlier?

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+**Impact: A single blocking call can cause cascading thread pool starvation.**
+
+## Memory / Allocation
+
+### 6. Use Span\<T\> Instead of Arrays for Slicing
+🔴 **DO** | .NET Core 2.1+
+
+❌
+```csharp
+string sub = input.Substring(5, 10);
```
stephentoub, line 44: Nit: this isn't an array, it's a string. The same advice applies, but the title talks about arrays

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+
+❌
+```csharp
+byte[] buf = new byte[4096];
+```
+✅
+```csharp
+byte[] buf = ArrayPool<byte>.Shared.Rent(4096);
+try { Process(buf); } finally { ArrayPool<byte>.Shared.Return(buf); }
```
stephentoub: I'd remove the try/finally

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+```
+**Impact: Dramatically reduces GC pressure for buffer-heavy workloads.**
+
+### 8. Avoid stackalloc in Loops
```
stephentoub: There's an on-by-default analyzer for this

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+
+❌
+```csharp
+string s = string.Format("{0}.{1}", major, minor);
+```
+✅
+```csharp
+string s = $"{major}.{minor}";
+```
+**Impact: C# 10 interpolation ~40% faster, ~5x less allocation.**
```
stephentoub: What are these "impact" statements supposed to do? These numbers do not apply to all such transformations.

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+```
+✅
+```csharp
+int val = int.Parse(str.AsSpan(5, 3));
+```
+**Impact: Eliminates one string allocation per parse operation.**
+
+## Regular Expressions
+
+### 12. Use Source-Generated Regex [GeneratedRegex]
```
stephentoub: There's an analyzer for this

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+    new(@"\w+@\w+\.\w+", RegexOptions.Compiled);
+```
+✅
+```csharp
+[GeneratedRegex(@"\w+@\w+\.\w+")]
+private static partial Regex EmailRegex();
+```
+**Impact: Near-zero startup; required for AOT/trimming scenarios.**
+
+### 13. Use NonBacktracking for Untrusted Input
```
stephentoub: This isn't about perf

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+
+❌
+```csharp
+bool ok = new Regex(@"\d{3}-\d{4}").IsMatch(s);
+```
+✅
+```csharp
+private static readonly Regex s_re = new(@"\d{3}-\d{4}", RegexOptions.Compiled);
+bool ok = s_re.IsMatch(s);
+```
```
stephentoub: This guidance conflicts with the guidance to use the source generator

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+**Impact: Construction cost is orders of magnitude higher than matching.**
+
+### 16. Set Timeouts on Regex with Untrusted Input
```
stephentoub: This is about reliability / safety / correctness rather than perf

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+```csharp
+var r = new Regex(userPattern);
+```
+✅
+```csharp
+var r = new Regex(userPattern, RegexOptions.None, TimeSpan.FromSeconds(2));
+```
+**Impact: Prevents unbounded CPU consumption; no cost on normal matches.**
+
+### 17. Use TryGetValue Instead of ContainsKey + Indexer
```
stephentoub: There's an analyzer for this

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+```
+✅
+```csharp
+bool found = false;
+foreach (var item in items)
+    if (item.Name == target) { found = true; break; }
+```
+**Impact: Eliminates 1-3 allocations per call; measurable in tight loops.**
+
+### 19. Don't Iterate IEnumerable Multiple Times
```
stephentoub: There's an analyzer for this

skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,368 @@
...
+```
+✅
+```csharp
+[JsonSerializable(typeof(BlogPost))]
+internal partial class AppJsonCtx : JsonSerializerContext { }
+string json = JsonSerializer.Serialize(post, AppJsonCtx.Default.BlogPost);
+```
+**Impact: 37-44% faster; enables trimming and Native AOT.**
+
+### 21. Cache JsonSerializerOptions
```
stephentoub: There's an analyzer for this

skills/analyzing-dotnet-performance/references/io-and-serialization.md
```diff
@@ -0,0 +1,160 @@
+# I/O, Serialization & General Patterns
+
+### Use JsonSerializerContext for Trimming/AOT
```
stephentoub: This isn't about perf

skills/analyzing-dotnet-performance/references/io-and-serialization.md
```diff
@@ -0,0 +1,160 @@
...
+```csharp
+using var writer = new Utf8JsonWriter(stream);
+writer.WriteStartObject();
+writer.WriteString("Title", blogPost.Title);
+writer.WriteNumber("PublicationYear", blogPost.PublicationYear);
+writer.WriteEndObject();
+writer.Flush();
+```
+
+**Impact: ~44% faster than reflection-based JsonSerializer.**
```
stephentoub: I don't think we should be giving this advice, especially not since JsonSerializer has a fast path that effectively does the latter thing

src/dotnet/skills/analyzing-dotnet-performance/references/io-and-serialization.md
```diff
@@ -0,0 +1,160 @@
...
+    FileShare.Read, bufferSize: 4096, useAsync: true);
+
+byte[] buffer = new byte[1024];
+while (await fs.ReadAsync(buffer) != 0) { /* process */ }
+```
+
+**Impact: Up to 3x faster async reads; allocation reduced from megabytes to hundreds of bytes.**
+
+### Use Memory\<byte\> Overloads for Stream.ReadAsync/WriteAsync
+🟡 **DO** use `Memory<byte>`-based stream overloads instead of `byte[]` overloads | .NET 5+
```
stephentoub, line 38: There's an analyzer for this

skills/analyzing-dotnet-performance/references/memory-and-strings.md
```diff
@@ -0,0 +1,219 @@
...
+✅
+```csharp
+Span<char> buffer = stackalloc char[64];
+guid.TryFormat(buffer, out int written);
+```
+
+**Impact: Zero heap allocation, no GC pressure, instant alloc/dealloc.**
+
+### Use C# 10+ String Interpolation Handlers
+🟡 **DO** recompile on C# 10+ to get `DefaultInterpolatedStringHandler` | .NET 6+
```
stephentoub: What is the skill supposed to do with this? Upgrade to the latest version?

skills/analyzing-dotnet-performance/references/memory-and-strings.md
```diff
@@ -0,0 +1,219 @@
...
+✅
+```csharp
+Span<char> buffer = stackalloc char[64];
+buffer.TryWrite($"Date: {dt:R}", out int charsWritten);
+```
+
+**Impact: Zero heap allocation for formatting operations.**
+
+### Use Span-Based Comparison Instead of Substring Equality
+🟡 **DO** use `AsSpan().SequenceEqual()` instead of `Substring` for comparisons | .NET Core+
```
stephentoub: Is this different from the earlier advice to use slicing rather than substring?

src/dotnet/skills/analyzing-dotnet-performance/references/memory-and-strings.md
```diff
@@ -0,0 +1,219 @@
...
+var sb = new StringBuilder(input.Length);
+// single pass replacing all patterns
+
+return string.Create(totalLength, state, (span, s) => { /* write directly */ });
+
+var sb = new StringBuilder();
+foreach (var part in parts)
+    sb.Append(separator).Append(part);
+return sb.ToString();
+```
```
stephentoub, line 156: I don't understand this code example... is this three different alternatives? What is the first one supposed to be?

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+**Impact: Source generator gives best throughput with near-zero startup. NonBacktracking prevents O(2^N) worst case at the cost of ~128x slower best case.**
+
+### Use IsMatch When You Only Need a Boolean Result
+🟡 **DO** use `IsMatch` instead of `Match(...).Success` | .NET 7+
+
+❌
+```csharp
+bool found = Regex.Match(input, pattern).Success;
```
stephentoub, line 27: There's an analyzer for this now

skills/analyzing-dotnet-performance/references/regex-patterns.md
```diff
@@ -0,0 +1,149 @@
...
+```
+✅
+```csharp
+bool found = Regex.IsMatch(input, pattern);
+```
+
+**Impact: Avoids Match object allocation; with NonBacktracking, ~3x faster by skipping capture computation.**
+
+### Keep IsMatch as a Guard Before Replace When Most Inputs Don't Match
+🟡 **DO NOT** remove an `IsMatch` guard before `Replace` to "reduce passes" | All versions
```
stephentoub: I don't understand this guidance.

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
```diff
@@ -0,0 +1,149 @@
...
+{
+    if (!regex.IsMatch(word))
+        return null;
+    return regex.Replace(word, replacement);
+}
+```
+
+
+
+### Use Regex.Count/EnumerateMatches Instead of Matches
```
stephentoub, line 36: There's an analyzer for this now

skills/analyzing-dotnet-performance/references/regex-patterns.md
```diff
@@ -0,0 +1,149 @@
...
+foreach (ValueMatch m in Regex.EnumerateMatches(text, @"\b\w+\b"))
+{
+    ReadOnlySpan<char> word = text.Slice(m.Index, m.Length);
+}
+```
+
+**Impact: Eliminates string allocations when working with spans — particularly valuable in high-throughput parsing pipelines.**
+
+### Budget Compiled Regex Instances at Startup
+🟡 **CONSIDER** the cumulative startup cost when using many `RegexOptions.Compiled` instances | .NET Core+
```
stephentoub: This doesn't seem useful or necessary given the advice to use the source generator

src/dotnet/skills/analyzing-dotnet-performance/references/structural-patterns.md
```diff
@@ -0,0 +1,75 @@
+# Structural Patterns
+
+Patterns detected by the **absence** of a keyword or interface. These require codebase-wide counting scans, not single-file matching.
+
+### Seal Classes for Devirtualization
```
stephentoub, line 5: This was already discussed earlier

skills/analyzing-dotnet-performance/references/structural-patterns.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+**Impact: Virtual calls up to 500x faster; type checks ~25x faster. Severity scales with count.**
+
+**Scale-based severity:**
+- 1-10 unsealed leaf classes → ℹ️ Info
+- 11-50 unsealed leaf classes → 🟡 Moderate
+- 50+ unsealed leaf classes → 🟡 Moderate (elevated priority)
+
+### Implement IEquatable\<T\> on Structs to Avoid Boxing
```
stephentoub: It's not clear to me what implementing the interface has to do with boxing

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed this pattern (correctness, not perf).

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed — compiler already catches this.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed — redundant with the sync-over-async pattern.

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: fixed title to reference substring/string slicing instead of arrays.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed the try/finally from the ArrayPool example.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: qualified the impact statement — it now says "up to" and notes results vary by context.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed — safety/reliability, not perf.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed — conflicts with the source generator guidance.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: Addressed by Copilot CLI: removed — safety/correctness, not perf.

skills/analyzing-dotnet-performance/references/io-and-serialization.md
artl93: Addressed by Copilot CLI: removed — not about perf.

skills/analyzing-dotnet-performance/references/io-and-serialization.md
artl93: Addressed by Copilot CLI: removed — JsonSerializer's fast path covers this.

skills/analyzing-dotnet-performance/references/async-patterns.md
artl93: Addressed by Copilot CLI: removed — not primarily a perf concern.

skills/analyzing-dotnet-performance/references/async-patterns.md
artl93: Addressed by Copilot CLI: removed — convenience, not perf.

skills/analyzing-dotnet-performance/references/async-patterns.md
artl93: Addressed by Copilot CLI: removed — correctness, not perf.

src/dotnet/skills/analyzing-dotnet-performance/references/collections-and-linq.md
artl93: Addressed by Copilot CLI: qualified to ultra-hot-path-only with benchmarked evidence. Also added a Common Pitfalls entry warning against broad suggestion.

skills/analyzing-dotnet-performance/references/collections-and-linq.md
artl93: Addressed by Copilot CLI: moved to the Common Pitfalls table in SKILL.md as a guardrail — the skill should not suggest replacing readable LINQ with loops outside hot paths.

skills/analyzing-dotnet-performance/references/collections-and-linq.md
artl93: Addressed by Copilot CLI: added mention of `params ReadOnlySpan<T>` as the modern (.NET 9+) alternative.

skills/analyzing-dotnet-performance/references/memory-and-strings.md
artl93: Addressed by Copilot CLI: removed — the skill can't upgrade framework versions.

skills/analyzing-dotnet-performance/references/memory-and-strings.md
artl93: Addressed by Copilot CLI: merged into the existing Span/AsSpan slicing guidance — same advice.

src/dotnet/skills/analyzing-dotnet-performance/references/memory-and-strings.md
artl93: Addressed by Copilot CLI: split the confusing code block into three distinct anti-pattern / fix pairs — chained Replace, chained ToLower+Trim+Substring, and LINQ on char sequences.

skills/analyzing-dotnet-performance/references/regex-patterns.md
artl93: Addressed by Copilot CLI: removed — unnecessary given the source generator guidance.

src/dotnet/skills/analyzing-dotnet-performance/references/structural-patterns.md
artl93: Addressed by Copilot CLI: removed the redundant stub from critical-patterns.md. The full seal-classes entry with detection recipes remains in structural-patterns.md.

skills/analyzing-dotnet-performance/references/regex-patterns.md
artl93: Addressed by Copilot CLI: added detailed rationale. `Regex.Replace` always allocates a new string even when nothing matches. `IsMatch` is a fast-path guard that skips that allocation when most inputs don't match — the win comes from avoiding the unconditional string allocation on the no-match path.

skills/analyzing-dotnet-performance/references/structural-patterns.md
artl93: Addressed by Copilot CLI: added detailed explanation. Without `IEquatable<T>`, `EqualityComparer<T>.Default.Equals` for structs resolves to the `object.Equals(object)` overload, which boxes the struct argument on every comparison. Implementing the interface provides a typed `Equals(T)` overload that the comparer can call without boxing.

skills/analyzing-dotnet-performance/references/regex-patterns.md
stephentoub: > Regex.Replace always allocates a new string even when nothing matches

No, it doesn't.

skills/analyzing-dotnet-performance/references/structural-patterns.md
stephentoub: What API is this referring to?

skills/analyzing-dotnet-performance/references/regex-patterns.md
artl93: You're right — Regex.Replace does not always allocate. Removed this pattern. (Copilot CLI)

skills/analyzing-dotnet-performance/references/structural-patterns.md
artl93: Fair point — removed the IEquatable pattern entirely. (Copilot CLI)

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis. I get the default nature here as well. I submit this should cost very little.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

skills/analyzing-dotnet-performance/references/critical-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

src/dotnet/skills/analyzing-dotnet-performance/references/io-and-serialization.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
artl93: IMHO: having an analyzer for this infers that it must be run in build or in the IDE. I suggest that having this here is still useful under static analysis.

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
stephentoub: Any changes AI makes should really run build, anyway. dotnet format can also run analyzers. Using AI for things that could alternately be done 100% deterministically, both detection and fixes, and with zero token cost, is using the wrong tool, IMO. On top of which more instructions can decrease the effectiveness of the rest. But it's up to you.

src/dotnet/tests/analyzing-dotnet-performance/fixtures/critical-antipatterns.cs
```diff
@@ -0,0 +1,41 @@
...
+public class TextProcessor
+{
+    private static readonly HttpClient _client = new HttpClient();
+
+    /// <summary>
+    /// Downloads content and processes it line-by-line.
+    /// </summary>
+    public async Task<string> ProcessUrlAsync(string url)
+    {
+        var content = await _client.GetStringAsync(url).Result; // sync-over-async
```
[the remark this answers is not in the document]
artl93: this is a counter-example for the skill

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
artl93: Resolving - static analysis IMO is still useful without a build.

src/dotnet/agents/optimizing-dotnet-performance.agent.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Output Format
+
+Keep reports concise and actionable. Avoid verbose prose.
+
+1. **Summary Assessment**: 1-2 sentences on the issue or opportunity
+2. **Root Cause**: Why the code is slow (one paragraph max)
+3. **Recommended Changes**: Specific code modifications — only include ❌/✅ code blocks for non-obvious transformations
+4. **Expected Impact**: Realistic estimates (e.g., "2-3x faster in this scenario")
+5. **Trade-offs**: One-line per trade-off, only when relevant
```
danmoseley: mention to evaluate maintanability effects specifically maybe? very often that's a tradeoff for a perf improvement. 

also whether it's worth it -- we don't need a command line parser to be vectorized/multithreaded - maybe that's #4 though.

eval-results.md
```diff
@@ -0,0 +1,30 @@
+# Eval Results — analyzing-dotnet-performance
```
danmoseley: is this intended to be committed or just pasted in the PR comments?

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,277 @@
...
+```
+✅
+```csharp
+Span<byte> buf = stackalloc byte[1024];
+for (int i = 0; i < 10_000; i++) { Process(buf); }
+```
+**Impact: StackOverflowException — unrecoverable, no catch possible.**
+
+### Avoid Boxing Value Types
+🔴 **AVOID** | .NET Core+
```
danmoseley: I think?
```suggestion
🔴 **AVOID** | .NET 6+
```

src/dotnet/skills/analyzing-dotnet-performance/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+| Severity | Criteria | Action |
+|----------|----------|--------|
+| 🔴 **Critical** | Deadlocks, crashes, security vulnerabilities, >10x regression | Must fix |
+| 🟡 **Moderate** | 2-10x improvement opportunity, best practice for hot paths | Should fix on hot paths |
+| ℹ️ **Info** | Pattern applies but code may not be on a hot path | Consider if profiling shows impact |
+
+**Prioritization rules:**
+1. If the user identified hot-path code, elevate all findings in that code to their maximum severity
+2. If hot-path context is unknown, report 🔴 Critical findings unconditionally; report 🟡 Moderate findings with a note: _"Impactful if this code is on a hot path"_
+3. Never suggest micro-optimizations on code that is clearly not performance-sensitive (startup, configuration, one-time initialization)
```
danmoseley: AI says -- this is contradicted because regex construction at startup is mentioned in critical-patterns.md as a critical finding and it's in eval/scenario-2

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,277 @@
...
+
+```bash
+# .IndexOf(string) without StringComparison (culture-aware, 2-3x slower)
+grep -rn --include='*.cs' -E '\.IndexOf\("[^"]+"\)' --exclude-dir=bin --exclude-dir=obj . | wc -l
+
+# .Substring( calls (allocates new string — consider AsSpan)
+grep -rn --include='*.cs' '\.Substring(' --exclude-dir=bin --exclude-dir=obj . | wc -l
+
+# .StartsWith/.EndsWith/.Contains without StringComparison (culture-aware, 2-3x slower)
+grep -rn --include='*.cs' -E '\.(StartsWith|EndsWith|Contains)\("[^"]+"\)' --exclude-dir=bin --exclude-dir=obj . | wc -l
```
danmoseley: AI points out that Contains will match many common API that don't have comparison overloads. This is true, not sure it matters, but you could break that into its own grep command with that note.

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,277 @@
...
+✅
+```csharp
+private static readonly SearchValues<char> s_hex = SearchValues.Create("ABCDEF");
+int pos = text.AsSpan().IndexOfAny(s_hex);
+```
+**Impact: 2-10x faster for chars; 10-30x faster for multi-string (.NET 9+).**
+
+## Detection
+
+Scan recipes for critical anti-patterns. Run these and report exact counts.
```
danmoseley: AI says that exact counts, gotten through grep and code tracing, are likely to be beyond AI capability and it will make up stuff. 

I think what it means is that it can't say stuff like "allocates 7 objects" reliably. But what you mean is simply "say how many things you found".
```suggestion
Scan recipes for critical anti-patterns. Run these and report exact counts of issues found in each case.
```

src/dotnet/skills/analyzing-dotnet-performance/SKILL.md
danmoseley: I guess the issue here is assuming startup is not performance sensitive (!) .. I'd remove that.

Also for regex specifically, the guidance (checked with Stephen) is ALWAYS transform any static pattern to `[GeneratedRegex]`. This is basically always good or neutral. Startup and cold paths included. This should be clear.

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
```diff
@@ -0,0 +1,277 @@
...
+```csharp
+private static readonly Regex s_re =
+    new(@"\w+@\w+\.\w+", RegexOptions.Compiled);
+```
+✅
+```csharp
+[GeneratedRegex(@"\w+@\w+\.\w+")]
+private static partial Regex EmailRegex();
+```
+**Impact: Near-zero startup; required for AOT/trimming scenarios.**
```
danmoseley: Maybe emphasize here (per my other comment) it should ALWAYS be recommended for static patterns. AOT is yet another reason.

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
```diff
@@ -0,0 +1,95 @@
...
+```csharp
+[GeneratedRegex("pattern")]
+private static partial Regex MyRegex();
+
+var safe = new Regex(untrustedPattern, RegexOptions.NonBacktracking);
+
+var oneOff = new Regex("pattern");
+```
+
+**Impact: Source generator gives best throughput with near-zero startup. NonBacktracking prevents O(2^N) worst case at the cost of ~128x slower best case.**
```
danmoseley: correct decision tree v alidated with Stephen

1) if predictable execution time is important use NonBacktracking (flag if pattern is not supported by NonBacktracking -- user needs to make call here) with (if pattern is static) generated, else
2) if pattern is dynamic, and either AOT scenario OR construction time is more important than execution time (eg., executed once on small text, maybe very simple pattern) use Interpreted else
3) if pattern is dynamic use Compiled.
4) else use Generated (90%+ case)

Now how to encode this for the AI I don't know. The only vital concern here is to use NonBacktracking for predictable execution time if the scneario requires that (relevance for potential DOS). Maybe instruction to AI looks like simply "if pattern is static change to GeneratedRegex, do not remove NonBacktracking if presenbt" that's simple and pretty safe.

danmoseley: edited above

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
```diff
@@ -0,0 +1,95 @@
+# Regex Patterns
+
+### Choose the Right Regex Engine Mode
+🟡 **DO** select the regex engine based on your scenario's trade-offs | .NET 7+
```
danmoseley:
```suggestion
🟡 **DO** use [GeneratedRegex] for all static regex patterns, but never remove NonBacktracking if present  | .NET 7+
```
I think encodes the right behavior.

src/dotnet/skills/analyzing-dotnet-performance/references/regex-patterns.md
danmoseley: attempted to address  below.

src/dotnet/skills/analyzing-dotnet-performance/references/critical-patterns.md
danmoseley: (probably worth doing actually)

src/dotnet/skills/analyzing-dotnet-performance/references/structural-patterns.md
```diff
@@ -0,0 +1,38 @@
...
+🟡 **DO** seal all leaf classes (those not subclassed) | .NET Core 3.0+
+
+Sealing lets the JIT devirtualize/inline virtual calls and use pointer comparison for type checks. Every non-abstract, non-static class that is not subclassed should be sealed.
+
+**Detection:** This is an absence pattern — scan for classes that are NOT sealed.
+
+```bash
+# Count unsealed (non-abstract, non-static) classes
+grep -rn --include='*.cs' -E '^\s*(public |internal |private )?(partial )?class ' --exclude-dir=bin --exclude-dir=obj . | grep -v 'sealed' | grep -v 'abstract' | grep -v 'static' | wc -l
```
danmoseley: misses "protected class" and "file class"

danmoseley:
```suggestion
grep -rn --include='*.cs' -E '^\s*((public|internal|private|protected|file)\s+)?(partial\s+)?class ' --exclude-dir=bin --exclude-dir=obj . | grep -v 'sealed' | grep -v 'abstract' | grep -v 'static' | wc -l
```

src/dotnet/skills/analyzing-dotnet-performance/SKILL.md
danmoseley: several times in this PR startup is given as an example of a not perf sensitive path. This is not right.

src/dotnet/skills/analyzing-dotnet-performance/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+**For files under 500 lines, read the entire file first** — you'll spot most patterns faster than running individual grep recipes. Use grep to confirm counts and catch patterns you might miss visually.
+
+For each relevant pattern category, run the detection recipes below. Report exact counts, not estimates.
+
+**Core scan recipes** (run these when reference files aren't available):
+```
+# Strings & memory
+grep -n '\.IndexOf(\"' FILE                    # Missing StringComparison
+grep -n '\.Substring(' FILE                    # Substring allocations
+grep -n '\.StartsWith\|\.EndsWith\|\.Contains' FILE  # Missing StringComparison
```
danmoseley:
```suggestion
grep -En '\.(StartsWith|EndsWith|Contains)\("[^"]*"\)' FILE # Missing StringComparison
```

danmoseley: to match above. although, do we want to include quotes here, what about variables of type string passed in? maybe remove quotes from here and the other instance of this pattern
```suggestion
grep -En '\.(StartsWith|EndsWith|Contains)\s*\(' FILE # Missing StringComparison
```
this also handles any nested parens but is an improvement (here and above)

PR #26, opened 2026-02-15
simd modifying existing code skill

jeffschwMSFT:
while doing some broad .NET experience analysis, the model suggested it might not think of identifying common SIMD patterns and correctly modify the code without a skill.  here are some common simd patterns and ways to optimize.


src/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+- Loops that count occurrences of a byte value across a buffer
+- Validation loops that check every byte against a condition (e.g., "is this valid UTF-8?", "does this contain a null byte?")
+
+For each candidate found, record:
+1. The file path and method name
+2. The pattern category (A, B, C, or D)
+3. The current scalar approach
+4. The input data type and typical buffer size if determinable from context
+
+### Step 2: Assess and prioritize candidates
```
danmoseley: should this encourage using existing optimized API's more up front? eg `SearchValues<T>`, `MemoryExtensions.XX` are always better if available than hand rolled vectorized code. But maybe this is made clear elsewhere

src/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,81 @@
...
+- No existing scalar code to optimize (greenfield)
+- Buffers consistently < 64 bytes
+- Regex with back-tracking/capture groups/lookahead
+- String processing on small collections (< 20 items) where HashSet/sort dominates
+- `ReadOnlySpan<char>` UTF-16 that can't convert to byte processing
+- Code that doesn't process large byte/char buffers in loops
+
+## Pattern Categories
+
+**A — Byte/string search loops:** `IndexOf`, `Contains`, manual `for`/`while` scanning byte-by-byte, `Regex.IsMatch` on ASCII data
```
tannergooding: I would expect that we want to keep most paths using `IndexOf` as `IndexOf` and similar, as those are already optimized with SIMD

I would expect that manual scans are then better to replace with APIs like `IndexOf` in most cases, rather than trying to re-simdify them

src/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,81 @@
...
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+using System.Runtime.Intrinsics.Arm;
+```
+
+### Dispatch order (always preserve scalar fallback)
+1. `Avx2.IsSupported` → `Vector256<byte>`
+2. `Sse42.IsSupported` → `Vector128<byte>`
+3. `AdvSimd.IsSupported` → `Vector128<byte>` (Arm)
```
tannergooding: We typically do not want to use platform specific intrinsics like `Avx2`, `Sse42`, or `AdvSimd` and would prefer to use the xplat intrinsics instead.

In particular, we want to check `Vector256<T>.IsHardwareAccelerated` prior to using `Vector256` and `Vector128<T>.IsHardwareAccelerated` prior to using `Vector128`.

src/dotnet/skills/simd-vector-math/SKILL.md
```diff
@@ -0,0 +1,105 @@
+---
+name: simd-vector-math
+description: Optimizes scalar float/double vector math in .NET 8+ with System.Runtime.Intrinsics SIMD (cosine similarity, dot product, L2 distance, normalization). Only transforms existing code — never generates greenfield.
+---
+
+# SIMD Vector Math Optimization
+
+> **Act fast:** Scan for candidates immediately, assess quickly, implement the transformation, build, and run tests. Do not over-explore the codebase — focus on the target methods.
```
tannergooding: Similar consideration here. I expect in most cases we don't want to SIMDify the code at all and rather want to replace this with a call to an existing vectorized API on Span (or if the layering is correct TensorPrimitives).

There are then a subset of other APIs we may want to simdify, but we want to use cross platform intrinsics whenever possible and not platform specific intrinsics.

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
+---
+name: simd-pattern-matching
+description: Optimizes scalar byte/string pattern matching in .NET 8+ with cross-platform Vector128/Vector256 SIMD intrinsics. Transforms hot-path scalar code into vectorized implementations — never generates greenfield.
+---
+
+# SIMD Pattern Matching Optimization
+
+> **STOP — Not everything is a SIMD opportunity.** String processing on small collections (`ToLower`, `Trim`, `Sort`, dedup on < 20 items), operations covered by framework APIs (`Span<T>.IndexOf`, `SearchValues<T>`, `System.Text.Ascii`), and code without large byte/char buffer loops are NOT candidates for manual SIMD. Using `Vector256` for ASCII lowercasing when `System.Text.Ascii.ToLower()` exists is harmful — it adds complexity for zero benefit. If the code does not contain a scalar loop over a byte/char buffer ≥ 64 bytes, report "no optimization opportunity" and stop immediately.
```
tannergooding, line 8: Should this be more explicit on what counts as "framework APIs"?

This only lists `IndexOf`, but I'd expect almost anything supported by `Span<T>` (which is namely not on `Span<T>` itself, but exposed via extension methods on `MemoryExtensions`) to qualify as something we don't want to vectorize. Instead, we expect that `Span<T>` API has already vectorized its implementation if applicable.

--- 

In general this kind-of seems like it could be summarized as something like:
> Replace manual loops that can be implemented by existing built-in APIs taking Span or Array with a call to that built-in API instead (unless some explicit marker saying not to do so exists, which ideally we have for any such deviations explaining why it is special).

We then only consider vectorization of what remains.

tannergooding: -- It's not entirely clear to me what part of the skill here should be stop vs early exit vs just part of the requested work. But it seems like this isn't strictly "SIMD" but rather "loop optimization opportunities" where most end up being calls to Span APIs/extensions and the few remaining go deeper and may get SIMD optimization

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+> **Early exit:** If the code is simply reimplementing a framework API (`Span<T>.IndexOf`, `MemoryExtensions.IndexOfAny`, `SearchValues<T>`, etc.), replace with the API call and stop. Those are already SIMD-optimized internally. This skill is for cases that need manual vectorization.
+
+Scan an existing .NET 8+ codebase for scalar pattern matching code on hot paths that would benefit from manual SIMD vectorization using cross-platform `Vector128`/`Vector256` intrinsics. Focus on byte-level operations over large buffers where no existing framework API covers the operation. If no SIMD-eligible candidates exist, report that and stop.
```
tannergooding, line 12: Should this be limited to just `Vector128` and then something else expands that to include Vector256 and Vector512 if deemed worthwhile?

The actual vectorization for all platforms is namely `Vector128` and then we tend to opportunistically add Vector256/512 paths if we expect input sizes will be large enough to warrant the extra code.

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Decision Gate (mandatory — do this FIRST, before writing any code)
+
+1. Does the code contain a scalar loop over a byte or char buffer? **If NO → stop, report "no optimization opportunity"**
```
tannergooding, line 16: Why would this be limited to just byte/char? `Vector128<T>` supports `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `float`, and `double`. With `char` being indirectly supported via reinterpretation as `ushort`. In a generic context, this is checked for via `Vector128<T>.IsSupported` so we know that its an accelerable type.

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+2. Are the buffers ≥ 64 bytes on the hot path? **If NO → stop**
```
tannergooding, line 17: Why just `>= 64` bytes?

There are many workloads that benefit even for `16 bytes` (single SIMD operation), particularly for smaller `T` like `byte`, `sbyte`, `short`, and `ushort`.

That is, there isn't really a penalty to using V128 SIMD when the accesses are memory based regardless. Any overhead typically comes from naively handling the fallback for "trailing elements" or small sizes. There's a fairly simple pattern that allows avoiding that overhead and we're using it in some code paths, but not others.

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+    // scalar fallback (never delete this path)
+}
+else if (Vector256.IsHardwareAccelerated && buffer.Length >= Vector256<byte>.Count)
+{
+    // Vector256 code path
+}
+else
+{
+    // Vector128 code path
+}
```
tannergooding, line 83: So while this pattern is used quite a bit in corelib, its not really the "optimal" pattern and leads to additional checks and branches which hurt the "core" path for small inputs and results in extra work for larger inputs.

We use a much more optimal pattern in `System.Numerics.Tensors` which ensures that we have:
* 0 branches on hardware without acceleration
* 2 total branches (1 predicted, 1 jump table) for small inputs that can't be accelerated
* 3 branches (1 predicted, 1 loop check, 1 jump table) for inputs that can be accelerated
* Better trimming so that only one code path needs to live based on what the target machine supports

It's a little different to setup, but an example, with commentary, can be seen here: https://source.dot.net/#System.Numerics.Tensors/System/Numerics/Tensors/netcore/Common/TensorPrimitives.IUnaryOperator.cs,fdab74764af40a1e

Its notably well setup to be easy to follow, works with the current unsafe patterns we're using but can be trivially changed over to a "safe" approach that uses spans and slicing, can reduce code duplication (in corelib) using `ISimdVector<TSelf, T>` (avoiding repetition between V128/256/512 paths), and makes things much more pay for play so more paths can be accelerated.

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+### SIMD operations reference
+
+Use cross-platform `Vector128`/`Vector256` operations:
+- **Create/broadcast:** `Vector128.Create(value)`, `Vector256.Create(value)`
+- **Load/store:** `Vector128.LoadUnsafe(ref src, offset)`, `Vector256.StoreUnsafe(vec, ref dst, offset)`
+- **Comparison:** `Vector128.Equals(a, b)`, `Vector128.GreaterThan(a, b)`, `Vector128.LessThan(a, b)`
+- **Bitwise:** operators `&`, `|`, `^`, `~`
+- **Mask extraction:** `vec.ExtractMostSignificantBits()` → `uint` bitmask
+- **Population count:** `BitOperations.PopCount(mask)` for counting matches
+- **Shuffle:** `Vector128.Shuffle(vec, indices)` for nibble-lookup tables
```
tannergooding, line 95: Does there need to be an explicit callout for arithmetic operations?

Can we simplify this to just indicate that we prefer using the APIs on these types without listing them all?

We notably also have a number of other helpers so that things like `ExtractMostSignificantBits` is ideally needed/used less, such as `All`/`Any`/`None`, `IndexOf`, `LastIndexOf`, `Count`, etc

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Vectorized range comparison
+Broadcast range bounds → subtract lower bound (wraps for out-of-range bytes in unsigned arithmetic) → compare against range width. For validation: check all lanes pass. For counting: `ExtractMostSignificantBits` → `PopCount`.
```
tannergooding, line 98: This is an example where just using `Count` or `CountWhereAllBitsSet` is better than `ExtractMSB + PopCount`

plugins/dotnet/skills/simd-pattern-matching/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Nibble-lookup for character classes
+Build two 16-entry lookup tables (high/low nibble). `Vector128.Shuffle` with nibble index → AND results → `ExtractMostSignificantBits` + `PopCount` for counting.
+
+### Memory access pattern
+Head (scalar for pre-vector bytes) → Body (`Vector256.LoadUnsafe` / `Vector128.LoadUnsafe`) → Tail (overlapping last-vector for idempotent operations, or scalar remainder). Always use `ref MemoryMarshal.GetReference(span)` and `LoadUnsafe(ref T, nuint elementOffset)`.
```
tannergooding, line 104: Backtracking for idempotent operations is great, but we can also do the same thing for non-idempotent scenarios.

See for example here: https://source.dot.net/#System.Numerics.Tensors/System/Numerics/Tensors/netcore/Common/TensorPrimitives.IAggregationOperator.cs,00a250d202e0c1bb,references

Where the Vectrorized128 path is doing an aggregation operation and so can't include already processed elements twice, so we just get a mask based on what the remainder was so that we can appropriately adjust those to some identity value for the operation.

In other non-aggregation cases you can just merge it with the last stored vector so that the final store preserves those how they should've been, rather than potentially changing them.

plugins/dotnet/skills/simd-vector-math/SKILL.md
```diff
@@ -0,0 +1,122 @@
+---
+name: simd-vector-math
+description: Optimizes scalar float/double vector math in .NET 8+ with cross-platform Vector128/Vector256 SIMD intrinsics. Transforms hot-path scalar code into vectorized implementations — never generates greenfield.
```
tannergooding, line 3: Not sure I understand the difference between pattern matching and vector-math here.

Seems like these are both just "optimize loop code" scenarios where if we don't have an existing built-in, we want to vectorize it. All the general vectorization steps should then be essentially the same.

PR #39, opened 2026-02-19
Nullable reference migration skill

### LINKED ISSUE #38: skill: adding nullable annotations
this doesn't seem to have a PR yet so I'll make one.

danmoseley:
Generated by looking at our docs, blogs, other people's blogs, PR's that added NRT in many of our repos and discussions on them, relevant issues in some cases, NRT guidance in our repos... 

aspnet reviewed by @halter73 , ef core reviewed by @AndriySvyryd 

* Validated skill against a random repo that doesn't have NRT and is small: https://github.com/image-charts/c-sharp/pull/1 
* Validate scanner script before and after the above, and on jbogard/MediatR (which is post NRT) checking all flagged issues are correct
* Scanner script tests run by hand, not wired up into yml as I see there's a parallel PR that changes all that around.
* Validated both skill and scanner script before/after on System.Text.RegularExpressions in corefx to compare results. (Main takeaway is that we now do a much better result of migration as back then, we were working around compiler analysis issues.) The conversion succeeds and eliminates warnings.

* Tested with the skill-validator on a meaingfully sized repo, after failing with a toy repo, here's the results:
```
The skill passed validation this time. Key findings:

  ✅ Passed: +11.0% improvement, statistically significant [+8.8%, +13.3%]

  The real repo scenario (Decider, 35 files, 300 warnings) showed the most benefit (+13.3%):

   - The skill-enhanced agent was more systematic — tracked warning counts progressively (300→188→44→8→0)
   - Used the = null! pattern the skill teaches
   - Ran all tests to verify no breakage
   - Added WarningsAsErrors to prevent regressions
   - Quality rubric scored higher (+0.13)

  The tradeoff is more tokens and time (~50% more of each), but the quality and methodology improvements outweigh that per
  the validator's weighted scoring.

  As you predicted — the skill shows more value with a real, larger repo (300 warnings) vs the synthetic 20-warning
  project.

```

* In DapperLib/Dapper, took HEAD and reverted all NRT annotations. Then ran the skill, and compared its result with what it reverted (which was high quality by Marc Gravell). It found some oversights in its work, and updated the skill further with those learnings.

Re: size. This is a relatively large skill. It's 298 lines (under the 500 lines limit recommended) but ~9K tokens (above the 2.5K size skills bench recommends). My reasoning: this is a skill that is not likely to be used every day, and is likely to be used on relatively large codebases, with many subtleties to deal with to do a good job (which the skill attempts to teach). We've learned this in annotating our own product -- it created huge PR's with long review processes touching many files. That is quite different to eg a skill for writing pinvokes or vectorizing an algorithm, which might be brought in during the course of general coding; instead one would be likely to start a new AI context specficially for this work. More importantly, skills bench shows an 11% improvement over no-skill, which suggests to me the tokens are paying for themselves. 

Fixes #38

### DIFF
--- a/tests/dotnet/migrate-nullable-references/eval.yaml
+++ b/tests/dotnet/migrate-nullable-references/eval.yaml
@@ -3,7 +3,7 @@ scenarios:
     prompt: |
       Enable nullable reference types in this C# project and resolve all warnings.
       The project is at TaskManager/TaskManager.csproj. Make sure the public API
-      surface accurately reflects nullability. Build with: dotnet build TaskManager/TaskManager.csproj
+      surface accurately reflects nullability.
     setup:
       files:
         - path: "TaskManager/TaskManager.csproj"
@@ -239,7 +239,7 @@ scenarios:
       This project has nullable reference types enabled project-wide, but each file
       still has #nullable disable at the top. Remove #nullable disable from
       UserProfile.cs ONLY and fix all resulting nullable warnings in that file.
-      Do NOT modify any other files. Build with: dotnet build FileByFile/FileByFile.csproj
+      Do NOT modify any other files.
     setup:
       files:
         - path: "FileByFile/FileByFile.csproj"
@@ -333,9 +333,9 @@ scenarios:
         path: "FileByFile/NotificationService.cs"
         value: "#nullable disable"
     rubric:
-      - "Did the agent remove #nullable disable from UserProfile.cs and ONLY that file? OrderProcessor.cs and NotificationService.cs must still have #nullable disable at the top."
+      - "Did the agent follow the skill's file-by-file migration strategy, migrating UserProfile.cs completely (removing #nullable disable, annotating correctly) before moving on to other files?"
       - "Did the agent correctly annotate Bio as string? (nullable) since it is never set in the constructor and can legitimately be null, while keeping Name and Email as non-nullable string since they are set via the constructor?"
-      - "Does the project build with zero errors after the agent's changes? The agent should not need to modify any file other than UserProfile.cs."
+      - "Does the project build with zero errors after the agent's changes?"
     timeout: 300
 
   - name: "Enable NRT in ASP.NET Core Web API with EF Core"
@@ -344,7 +344,6 @@ scenarios:
       warnings fixed. It's a straightforward task. The project is at BookStore/BookStore.csproj.
       It uses Entity Framework Core with an in-memory database and has a Web API controller.
       Make sure annotations are correct for both EF Core entities and API models.
-      Build with: dotnet build BookStore/BookStore.csproj
     setup:
       files:
         - path: "BookStore/BookStore.csproj"


src/dotnet/skills/migrate-nullable-references/references/helper-extensions.md
```diff
@@ -0,0 +1,26 @@
...
+using System.Linq;
+
+internal static class NullableExtensions
+{
+    /// <summary>
+    /// Filters null values from a sequence and narrows the type from T? to T.
+    /// Unlike .Where(x => x != null), this gives the compiler correct type information.
+    /// </summary>
+    public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> source) where T : class
+        => source.Where(x => x is not null)!;
```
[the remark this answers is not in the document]
danmoseley: I am not sure this is right (this came from roslyn itself ... https://github.com/dotnet/roslyn/blob/af0287e7e9bb9a50fc6754ef266ba8f68118dc97/src/Dependencies/Collections/Extensions/IEnumerableExtensions.cs#L331)

src/dotnet/skills/nullable-reference-migration/SKILL.md
```diff
@@ -0,0 +1,286 @@
...
+   - **Yes, with a code path the compiler can't see** → add `!` with a comment explaining why.
+   - **Yes, by adding a guard** → add a null check (`if`, `??`, `is not null`).
+   - **No** → the type should be nullable (go back to step 1 — the answer is "Yes").
+
+Guidance:
+
+- Prefer explicit null checks (`if`, `is not null`, `??`) over the null-forgiving operator (`!`).
+- Use the null-forgiving operator only when you can prove the value is never null but the compiler cannot, and add a comment explaining why.
+- When a method legitimately returns null, change the return type to `T?` — do not hide nulls behind a non-nullable signature.
+- `Debug.Assert(x != null)` acts as a null-state hint to the compiler just like an `if` check. Use it at the top of a method or block to inform the flow analyzer about invariants and eliminate subsequent `!` operators in that scope. Note: `Debug.Assert` informs the compiler but is stripped from Release builds — it does not protect against null at runtime. For public API boundaries, prefer an explicit null check or `ArgumentNullException`.
```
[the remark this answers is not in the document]
danmoseley: @copilot I dont understand this. Debug.Assert is definitely a hint to the compiler. Maybe not if you're targeting a really old .NET Core, or .NET Framework but then we don't support the NRT compiler.

plugins/dotnet/skills/migrate-nullable-references/SKILL.md
```diff
@@ -0,0 +1,297 @@
...
+
+## ASP.NET Core Considerations
+
+If the project uses ASP.NET Core, see [references/aspnet-core.md](references/aspnet-core.md) — enabling NRTs can change MVC model validation and JSON serialization behavior.
+
+## Helper Extension Methods
+
+See [references/helper-extensions.md](references/helper-extensions.md) for `WhereNotNull` and other helper methods to add during migration.
+
+## More Info
```
AaronRobinsonMSFT, line 280: Unclear if this matters are much anymore, but the en-us locale should generally be removed from URLs. That way it defers to the language of the user. In this case, I'm not sure what "best practice" is here.

Also, @adityamandaleeka pointed out general guidance for URLs at https://github.com/dotnet/skills/pull/16#issuecomment-3893465008. I'm not sure we need to follow that for our official skills, but we should definitely confirm all URLs are to domains we own.

plugins/dotnet/skills/migrate-nullable-references/scripts/Get-NullableReadiness.ps1
```diff
@@ -0,0 +1,487 @@
+<#
+.SYNOPSIS
+    Scans a C# project or solution for nullable reference type (NRT) readiness.
```
eerhardt, line 3: Why ps1 and not a single-file .cs file?

plugins/dotnet/skills/migrate-nullable-references/references/aspnet-core.md
```diff
@@ -0,0 +1,16 @@
+# ASP.NET Core Considerations
```
eerhardt, line 1: (For my knowledge about skills) For these references, is it possible to link to the actual documentation instead of duplicating it here? Its unfortunate to duplicate it and then if these rules ever change these need to be updated.

plugins/dotnet/skills/migrate-nullable-references/SKILL.md
danmoseley: removed en-us. they are all to our domains.

src/dotnet/skills/migrate-nullable-references/references/helper-extensions.md
```diff
@@ -0,0 +1,26 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+
+internal static class NullableExtensions
+{
+    /// <summary>
+    /// Filters null values from a sequence and narrows the type from T? to T.
+    /// Unlike .Where(x => x != null), this gives the compiler correct type information.
+    /// </summary>
+    public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> source) where T : class
```
eerhardt: Are these generally helpful? Should they be real APIs?

src/dotnet/skills/migrate-nullable-references/SKILL.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Step 3: Fix dereference warnings
+
+> **Prioritization:** Work through files in dependency order — start with core models and shared utilities that other code depends on, then move to higher-level consumers. Within each file, fix public and protected members first (these define the contract), then internal and private members. This order minimizes cascading warnings: fixing a core type's annotations often resolves warnings in its consumers automatically.
+
+Build the project and work through dereference warnings. These are the most common:
+
+| Warning | Meaning | Typical fix |
+|---------|---------|-------------|
+| CS8602 | Dereference of a possibly null reference | Add a null check, use `?.`, or use a pattern like `if (x is not null)` |
```
eerhardt: Doesn't this contradict line 35 above?

> **Zero runtime behavior changes.** NRT migration is strictly a metadata and annotation exercise. The generated IL must not change — no new branches, no new null checks, no changed control flow, no added or removed method calls.

plugins/dotnet/skills/migrate-nullable-references/references/aspnet-core.md
lewing: one option I've used successfully was including a link to the external doc, it makes it trivial for an agent to know the source of truth and periodically review and update it if needed and it doesn't cost a lot in tokens.

plugins/dotnet/skills/migrate-nullable-references/references/aspnet-core.md
```diff
@@ -0,0 +1,16 @@
...
+- **Enable `JsonSerializerOptions.RespectNullableAnnotations = true` (.NET 9+)**: For .NET 9+ projects, always enable `RespectNullableAnnotations` (along with `RespectRequiredConstructorParameters`) to align runtime serialization behavior with your NRT annotations. Without this, `System.Text.Json` silently assigns `null` to non-nullable properties, undermining compile-time null safety. When enabled, the serializer throws `JsonException` when a non-nullable property receives an explicit `null` during deserialization, or emits `null` for a non-nullable property during serialization. Be aware this enforcement has hard limitations rooted in how NRTs are represented in IL. It does **not** cover:
```
halter73, line 7: It might be worth calling out that enabling NRT can change [the behavior of minimal APIs as well](https://github.com/dotnet/aspnetcore/blob/8cf6a9599a26b6c419a2cb1a30e8ed72ba941359/src/Http/Http.Extensions/src/RequestDelegateFactory.cs#L2405-L2413). Parameters that might have previously been interpreted as optional might no longer be treated as optional without explicitly marking it nullable.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/parameter-binding?view=aspnetcore-10.0#optional-parameters

plugins/dotnet/skills/migrate-nullable-references/references/aspnet-core.md
danmoseley: I do have links to the original docs at the bottom of SKILL.md. I think we're all discovering what is best pattern here: we want to keep tokens low and probably want docs source of truth, but OTOH the existence of the skill to some extent represents a hypothesis that pulling info together is helpful? I guess we will find out over time

plugins/dotnet/skills/migrate-nullable-references/scripts/Get-NullableReadiness.ps1
danmoseley: shrug - if a standard emerges, I'll convert. both are Microsoft tech :)

plugins/dotnet/skills/migrate-nullable-references/references/aspnet-core.md
danmoseley: thanks, added https://github.com/dotnet/skills/pull/39/changes/a33ee49cbf2e1275d37058c7ca22a253ffeb4f01

src/dotnet/skills/migrate-nullable-references/SKILL.md
danmoseley: good point, clarified now.

src/dotnet/skills/migrate-nullable-references/SKILL.md
```diff
@@ -0,0 +1,299 @@
+---
+name: migrate-nullable-references
+description: Enable nullable reference types in a C# project and systematically resolve all warnings. USE FOR: adopting NRTs in existing codebases, file-by-file or project-wide migration, fixing CS8602/CS8618/CS86xx warnings, annotating APIs for nullability, cleaning up null-forgiving operators, upgrading dependencies with new nullable annotations. DO NOT USE FOR: projects already fully migrated with zero warnings (unless auditing suppressions), suppressing warnings without fixing them, C# 7.3 or earlier projects. INVOKES: Scan-NullableReadiness.ps1 scanner script.
```
lewing: The colons in `USE FOR:`, `DO NOT USE FOR:`, and `INVOKES:` cause YAML parse failures in Copilot CLI ("Nested mappings are not allowed in compact mappings"). Use a block scalar instead:

```suggestion
description: >
  Enable nullable reference types in a C# project and systematically resolve all warnings.
  USE FOR: adopting NRTs in existing codebases, file-by-file or project-wide migration,
  fixing CS8602/CS8618/CS86xx warnings, annotating APIs for nullability, cleaning up
  null-forgiving operators, upgrading dependencies with new nullable annotations.
  DO NOT USE FOR: projects already fully migrated with zero warnings (unless auditing
  suppressions), suppressing warnings without fixing them, C# 7.3 or earlier projects.
  INVOKES: Scan-NullableReadiness.ps1 scanner script.
```

src/dotnet/skills/migrate-nullable-references/references/helper-extensions.md
danmoseley: @agocke thoughts? I don't see these used outside Roslyn, but they're used heavily there. I'm inclined to remove unless you think this is something important and general we should include here

tests/dotnet/migrate-nullable-references/eval.yaml
```diff
@@ -0,0 +1,588 @@
...
+        path: "FileByFile/UserProfile.cs"
+        value: "string?"
+      - type: "file_contains"
+        path: "FileByFile/OrderProcessor.cs"
+        value: "#nullable disable"
+      - type: "file_contains"
+        path: "FileByFile/NotificationService.cs"
+        value: "#nullable disable"
+    rubric:
+      - "Did the agent remove #nullable disable from UserProfile.cs and ONLY that file? OrderProcessor.cs and NotificationService.cs must still have #nullable disable at the top."
```
adityamandaleeka: This + the "Do NOT modify any other files" instruction in the prompt makes it seem like this is testing general instruction compliance. What's the intent here?

tests/dotnet/migrate-nullable-references/eval.yaml
```diff
@@ -0,0 +1,588 @@
...
+      - "Does the project build with zero errors after the agent's changes? The agent should not need to modify any file other than UserProfile.cs."
+    timeout: 300
+
+  - name: "Enable NRT in ASP.NET Core Web API with EF Core"
+    prompt: |
+      This small project (6 files) needs `<Nullable>enable</Nullable>` added and all resulting
+      warnings fixed. It's a straightforward task. The project is at BookStore/BookStore.csproj.
+      It uses Entity Framework Core with an in-memory database and has a Web API controller.
+      Make sure annotations are correct for both EF Core entities and API models.
+      Build with: dotnet build BookStore/BookStore.csproj
```
adityamandaleeka: I noticed you have these "build with" lines in multiple scenarios. Curious, did you find it necessary?

PR #54, opened 2026-02-20
Add .NET P/Invoke skill

AaronRobinsonMSFT:
This pull request adds comprehensive reference documentation and new evaluation scenarios to improve .NET P/Invoke usability and correctness. The main changes include the addition of diagnostic and type mapping guides for native interop, as well as new test scenarios to ensure correct generation of P/Invoke declarations for both .NET 8 and .NET Framework targets.

### Documentation Enhancements

* Added a new `diagnostics.md` reference guide covering common P/Invoke pitfalls, failure modes, debugging strategies, and key resources. This guide helps developers avoid and troubleshoot interop bugs, with practical advice for signature matching, memory management, and error handling.
* Added a `type-mapping.md` reference documenting precise mappings between C/Win32 types and .NET types, highlighting dangerous types, blittable types, and struct layout considerations. This is crucial for preventing subtle interop bugs due to type mismatches.

### Testing Improvements

* Introduced new evaluation scenarios in `eval.yaml` to test the generation of P/Invoke declarations from C headers for both .NET 8 (using `LibraryImport`) and .NET Framework (using `DllImport`). The scenarios assert correct type mapping (e.g., using `nuint` for `size_t`) and method signature style, ensuring robust code generation across platforms.

### DIFF
--- a/src/dotnet/skills/dotnet-pinvoke/SKILL.md
+++ b/src/dotnet/skills/dotnet-pinvoke/SKILL.md
@@ -403,5 +403,5 @@ For binding Objective-C libraries (macOS/iOS), use [Objective Sharpie](https://l
 
 ## Reference Files
 
-- **[references/type-mapping.md](references/type-mapping.md)** — Complete native-to-.NET type mapping table, struct layout patterns, blittable type rules
-- **[references/diagnostics.md](references/diagnostics.md)** — Common pitfalls, failure modes and recovery, debugging approach, external resources
+- **[references/type-mapping.md](references/type-mapping.md)** — Complete native-to-.NET type mapping table, struct layout patterns, blittable type rules. **Load when** encountering types not covered in Step 2 above, or when working with struct layout or blittable type questions.
+- **[references/diagnostics.md](references/diagnostics.md)** — Common pitfalls, failure modes and recovery, debugging approach, external resources. **Load when** debugging an existing P/Invoke failure or reviewing interop code for correctness issues.


src/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -0,0 +1,405 @@
...
+| `LPSTR` / `char*` | `string` | Must specify encoding (ANSI or UTF-8). Always requires marshalling cost for `in` parameters |
+
+**For the complete type mapping table, struct layout, and blittable type rules**, see [references/type-mapping.md](references/type-mapping.md).
+
+### Step 3: Write the Declaration
+
+Given a C header:
+
+```c
+int32_t process_records(const Record* records, size_t count, uint32_t* out_processed);
```
danmoseley, line 58: should docs be consulted too? they may have subtleties not expressed in the header (at least windows docs may).

AaronRobinsonMSFT: I don't think docs should ever be consulted for P/Invokes. Official Win32 API docs are notoriously lacking in many details about P/Invoke signatures, hence [CsWin32](https://github.com/microsoft/CsWin32) and [win32metadata](https://github.com/microsoft/win32metadata). Header files really should be the manner in which native P/Invoke signatures are written out.

danmoseley: that sounds like potentially a good guideline (guardrail) to add as a bullet in the skill. "Where documentation and header diverge, trust the header." or something. It's not obvious to me that that AI would do that. I'ts b een expensively trained on vast amounts of docs, and happily looks them up?

AaronRobinsonMSFT: Agree. Will add.

src/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+### Objective Sharpie (Objective-C APIs)
+
+For binding Objective-C libraries (macOS/iOS), use [Objective Sharpie](https://learn.microsoft.com/previous-versions/xamarin/cross-platform/macios/binding/objective-sharpie) to generate initial P/Invoke and binding definitions from Objective-C headers.
+
+---
+
+## Validation
+
+### Review checklist
+
```
danmoseley, line 380: are there any patterns or concerns that if observed the AI should be sure to notify the user about? I mean a situation where it could follow all the rules and guidance here correctly but still infer there is a security or stability concern that it can't fix as part of this process? (I don't have an example but if this can be imagined, then we should instruct the AI to alert the user perhaps.

src/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -0,0 +1,407 @@
...
+   <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
+   <EnableAotAnalyzer>true</EnableAotAnalyzer>
+   ```
+2. **Verify struct sizes match** — for every struct crossing the boundary, assert `Marshal.SizeOf<T>()` equals the native `sizeof`
+3. **Round-trip test** — call the native function with known inputs and verify expected outputs
+4. **Test with non-ASCII strings** — pass strings containing characters outside the ASCII range to confirm encoding is correct
+
+## Reference Files
+
+- **[references/type-mapping.md](references/type-mapping.md)** — Complete native-to-.NET type mapping table, struct layout patterns, blittable type rules
```
danmoseley: in what circumstances would we not want it to read this file? if it should always read it then presumably it should be inline (optional load is the reason for the references pattern)

(I mentioned this earlier but not sure of your answer)

AaronRobinsonMSFT: Do you know the best way to encode that decision?

src/dotnet/skills/dotnet-pinvoke/SKILL.md
AaronRobinsonMSFT: Thanks for that, it is an interesting angle to consider. I've called out things that are notorious for interop (that is, string encoding, lifetime, delegates) in the document. Other areas, I feel, are things we'll learn as we go I hope.

src/dotnet/skills/dotnet-pinvoke/SKILL.md
AaronRobinsonMSFT: My latest commit has added something to the references below.

PR #63, opened 2026-02-20
Add microbenchmarking skill

caaavik-msft:
I've been iterating over this for a couple days and just wanted to put up what I have so far and get some early feedback. I plan on checking in my automated evaluation code which uses Copilot SDK inside a docker container as the test environment. I'll also look into getting it set up properly with the right project structure as well, but I have just put it inside the dotnet plugin for now until we see where it might be better placed.

What I have found with multiple evaluations over the last week is that AI is very prone to doing the wrong thing or following bad assumptions about BenchmarkDotNet, and it is very hard to have it be useful to create benchmarks. With this I was able to get it to a point where I could reliably have it create benchmarks and not get stuck in loops and waste time running long benchmarks that provide no value.

A lot of the skill's focus goes into recommending ways for the AI to effectively use BenchmarkDotNet such that it will validate incrementally using dry runs and then run benchmarks in small batches rather than sending off a run that will take longer than 30 minutes.

I also found that multiple models seem to default to using BDN 1.14.0 which is over 2 years old, and then they have issues with .NET 10 compatibility using it that they get stuck on, so I had to guide it to just trusting whatever `dotnet add package BenchmarkDotNet` returns. It also was very difficult to convince the AI that it might be operating on old information, and so I had to really reinforce the value of the updated information for it to even try and load some of the reference files.

I have been using the following prompt for automated evaluation:

> You are creating a brand new BenchmarkDotNet benchmark project from scratch.
> Your working directory ({workDir}) is empty — create all project files there.
> There are no existing benchmarks to find or reuse.
> 
> The dotnet/runtime source tree is available read-only at: {runtimeSrcRoot}
> Use it to understand the API surface of the library you are benchmarking.
> 
> Task:
> Create a BenchmarkDotNet benchmark project for {targetDescription}, whose source code is in the dotnet/runtime repository.
> These benchmarks are intended to be part of a permanent coverage suite that is run periodically to detect performance regressions.
> Target ~10 benchmark test cases (counting all parameter combinations across all benchmark methods).
> Run the benchmarks comparing .NET 9 and .NET 10 performance side-by-side and provide the results.
>
> Permissions:
> - You may read/search files in the dotnet/runtime source directory.
> - You may create and edit files in your working directory.
> - You may run shell commands.

With this skill loaded, it is able to handle this in a single shot without any hiccups on multiple models in under 15 minutes (including time spent running the benchmark) and produce good quality benchmarks with valid results and analysis. Without skills, I have not gotten it to work once as it runs into multiple issues with setting it up correctly and finding the right configs/arguments to use, but also it has a very slow iteration loop as it always runs the full benchmark suite each time rather than using configs and args to iterate faster.


src/dotnet/skills/microbenchmarking/SKILL.md
```diff
@@ -0,0 +1,138 @@
+---
+name: microbenchmarking
+description: >
+  Activate this skill when BenchmarkDotNet is involved in the task — creating,
```
danmoseley:
```suggestion
  Activate this skill when BenchmarkDotNet (BDN) is involved in the task — creating,
```

tests/dotnet/microbenchmarking/eval.yaml
```diff
@@ -0,0 +1,29 @@
...
+        path: "TextAnalyzer/*.cs"
+      - type: "file_contains"
+        path: "**/*.cs"
+        value: "Benchmark"
+    rubric:
+      - "The benchmarks follow best practices: methods return their results to prevent dead code elimination, input data is stored in fields/properties rather than const or literal values, and initialization/data setup is in a setup method not inside the benchmark methods"
+      - "The agent did not hardcode a specific BenchmarkDotNet version that may be outdated, instead letting NuGet resolve the latest stable version"
+      - "The benchmarks were actually run and completed, producing results that compare .NET 9 vs .NET 10"
+      - "The agent correctly identified that AverageWordLength and WordLengthHistogram got significantly faster on .NET 10 (expected ~2-3x faster), while UniqueWordCount showed no meaningful change"
+      - "The agent completed the task efficiently within the time constraints rather than spending excessive time on unnecessary iterations or troubleshooting"
```
danmoseley, line 26: we're not validating anything about the content of the benchmark program - should we?

tests/dotnet/microbenchmarking/eval.yaml
```diff
@@ -0,0 +1,29 @@
...
+      asked to benchmark AverageWordLength, WordLengthHistogram, and UniqueWordCount
+      to check how the upgrade will impact their performance. The source code is in
+      the TextAnalyzer subdirectory (TextAnalyzer.cs and TextAnalyzer.csproj). We don't
+      need anything too comprehensive, don't spend more than 10 minutes on this.
+      Don't worry about cleaning up any files or projects you create along the way.
+      If the .NET 9 runtime is missing, install it using the official dotnet-install
+      script from https://dot.net/v1/dotnet-install.sh (or dotnet-install.ps1 on Windows).
+    setup:
+      copy_test_files: true
+    assertions:
```
adityamandaleeka, line 15: Seems like these assertions aren't about what the agent does, but just test setup? They probably shouldn't be here.

src/dotnet/skills/microbenchmarking/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+  running, configuring, or reviewing BDN benchmarks. Also activate when
+  microbenchmarking .NET code would be useful and BenchmarkDotNet is the likely
+  tool. Consider activating when answering a .NET performance question requires
+  measurement and BenchmarkDotNet may be needed.
+  Covers microbenchmark design, BDN configuration and project setup, how to run
+  BDN microbenchmarks efficiently and effectively, and using BDN for side-by-side
+  performance comparisons.
+  Do NOT use for standalone profiling (dotnet-trace, PerfView), production
```
adityamandaleeka: "standalone profiling" might be a confusing term.

plugins/dotnet/skills/microbenchmarking/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+
+BDN programs use either **`BenchmarkSwitcher`** (provides interactive benchmark selection for humans, parses CLI arguments) or **`BenchmarkRunner`** (runs specified benchmarks directly). Both support CLI flags like `--filter` and `--runtimes`, but only when `args` is passed through — without it, CLI flags are silently ignored. When using `BenchmarkSwitcher`, always pass `--filter` to avoid hanging on an interactive prompt.
+
+BDN behavior is customized through **attributes**, **config objects**, and **CLI flags**.
+
+See [references/project-setup-and-running.md](references/project-setup-and-running.md) for entry point setup, config object patterns, and CLI flags. If you need to collect data beyond wall-clock time — such as memory allocations, hardware counters, or profiling traces — see [references/diagnosers-and-exporters.md](references/diagnosers-and-exporters.md).
+
+## Running benchmarks
+
+BenchmarkDotNet console output is extremely verbose — hundreds of lines per case showing internal calibration, warmup, and measurement details. Redirect all output to a file to avoid consuming context on verbose iteration output:
```
adityamandaleeka, line 85: Have you found this to be useful? Coding agents already use techniques like this for context management, and are getting better about it over time.

plugins/dotnet/skills/microbenchmarking/references/bdn-internals-and-tuning.md
```diff
@@ -0,0 +1,154 @@
+# BenchmarkDotNet Execution and Configuration
```
adityamandaleeka, line 1: There's a LOT of info in these files. It seems like the runtime repo is having success with EgorBot. I wonder if a lot of this stuff can be removed without losing effectiveness.

cc @EgorBo for thoughts.

tests/dotnet/microbenchmarking/eval.yaml
caaavik-msft: It is validated in the first rubric item. However, AI is already very good at writing benchmarks reliably from my experiments, so I think the value here is primarily about effectively using BDN to solve tasks rather than writing them.

plugins/dotnet/skills/microbenchmarking/references/bdn-internals-and-tuning.md
caaavik-msft: I think the main reason I have this here is because without it, the AI doesn't really have a good way to understand how long the benchmarks are that it is about to run, it's very eager to just go create 20 benchmarks and run with default settings which takes a long time. So the goal here was for it to develop an intuition to be able to estimate how long it will run for which might matter more in an interactive workflow with a developer, but it matters less if it is just running in a pipeline and has no time constraints. I do think though maybe a lot can be trimmed, at least initially.

plugins/dotnet/skills/microbenchmarking/SKILL.md
caaavik-msft: I have found that it does not do this when running BDN by default, because it doesn't know what the output format looks like to filter it effectively. But for things like `dotnet build` or `dotnet test`, the training data knows better what strings to look for to filter it. But I should try do some more evals without this and see how it compares context-usage wise.

caaavik-msft: I have done some side-by-side testing of the skill with and without this, and without telling it this, it gets confused by what it is seeing in stdout and keeps thinking it is running even after it is finished. So, I think this is still valuable, although maybe there are other ways to handle this without it saving to disk.

adityamandaleeka: Are you running on Windows? I've noticed that Copilot CLI seems to have an issue with understanding when a spawned process has exited on Windows.

tests/dotnet/microbenchmarking/eval.yaml
caaavik-msft: I've updated it to be a better assertion that tests that there is a project with BenchmarkDotNet and there is a markdown report somewhere

src/dotnet/skills/microbenchmarking/SKILL.md
caaavik-msft: I've modified it to `Do NOT use for profiling/tracing .NET code (dotnet-trace, PerfView), ...` If that sounds better.

plugins/dotnet/skills/microbenchmarking/SKILL.md
caaavik-msft: Yes I have been running on Windows. I think what also made it confused is that the output was being truncated because of how much output was being written, that it thought it hadn't finished since it didn't have the results showing it finished, so it kept assuming it was still running.

plugins/dotnet/skills/microbenchmarking/references/bdn-internals-and-tuning.md
caaavik-msft: I have trimmed down all the files a bit more

PR #72, opened 2026-02-20
Improve dotnet-pinvoke skill: stop signals, routing, inline anti-patterns

lewing:
## Builds on PR #54

This PR stacks on #54 (Add .NET P/Invoke skill by @AaronRobinsonMSFT) — it should be reviewed/merged after #54 lands.

## Changes

Improvements based on multi-model evaluation against [skill-builder patterns](https://github.com/lewing/agent-plugins/tree/main/plugins/skill-trainer/skills/skill-builder).

### What changed
- **Structured description keywords** — Reformatted to \USE FOR:\ / \DO NOT USE FOR:\ for better routing
- **When to Use section** — 6 concrete trigger scenarios  
- **Stop Signals section** — Prevents agents from marching through all 8 steps on simple tasks
- **Inline anti-patterns** — 4 \❌ NEVER\ callouts near Steps 2, 4, and 5

### Why
Without stop signals, an agent asked to 'write a P/Invoke for CreateFile' would go through all 8 steps including migration and tooling — now it knows to stop at Step 3 for single-function tasks.

### Evidence
Multi-model eval (3 models, 2 families) scored **4.8/5 average**:

| Model | Score |
|-------|-------|
| Claude Sonnet 4 | 4.8/5 |
| GPT-5.1 | 5.0/5 |
| Claude Haiku 4.5 | 4.6/5 |

- [Assessment gist](https://gist.github.com/lewing/859ab56b0c37601804c03a5c601cfd8d)
- [Training log gist](https://gist.github.com/lewing/e272e71dca0f4d1036a02a65796b5e1c)

### DIFF
--- a/plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
+++ b/plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
@@ -1,6 +1,14 @@
 ---
 name: dotnet-pinvoke
-description: Correctly call native (C/C++) libraries from .NET using P/Invoke and LibraryImport. Covers function signatures, string marshalling, memory lifetime, SafeHandle, and cross-platform patterns. USE FOR: writing new P/Invoke or LibraryImport declarations, reviewing or debugging existing native interop code, wrapping a C or C++ library for use in .NET, diagnosing crashes, memory leaks, or corruption at the managed/native boundary. DO NOT USE FOR: COM interop, C++/CLI mixed-mode assemblies, or pure managed code with no native dependencies.
+description: >
+  Correctly call native (C/C++) libraries from .NET using P/Invoke and LibraryImport.
+  Covers function signatures, string marshalling, memory lifetime, SafeHandle, and
+  cross-platform patterns.
+  USE FOR: writing new P/Invoke or LibraryImport declarations, reviewing or debugging
+  existing native interop code, wrapping a C or C++ library for use in .NET, diagnosing
+  crashes, memory leaks, or corruption at the managed/native boundary.
+  DO NOT USE FOR: COM interop, C++/CLI mixed-mode assemblies, or pure managed code with
+  no native dependencies.
 ---
 
 # .NET P/Invoke


plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -9,6 +9,23 @@ Calling native code from .NET is powerful but unforgiving. Incorrect signatures,
 
 This skill covers both `DllImport` (available since .NET Framework 1.0) and `LibraryImport` (source-generated, .NET 7+). When targeting .NET Framework, always use `DllImport`. When targeting .NET 7+, prefer `LibraryImport` for new code. When native AOT is a requirement, `LibraryImport` is the only option.
 
+## When to Use This Skill
```
AaronRobinsonMSFT, line 20: Ugh. I have no idea why this was removed. I think when I asked copilot to make the skill less than 500 lines it removed this. This should have been there. Apologizes @danmoseley I didn't realize you were calling out this wasn't present at all. That is on me :(

plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -50,6 +67,10 @@ The most dangerous mappings — these cause the majority of bugs:
 
 **For the complete type mapping table, struct layout, and blittable type rules**, see [references/type-mapping.md](references/type-mapping.md).
 
+> ❌ **NEVER** use `int` or `long` for C `long` — it's 32-bit on Windows, 64-bit on Unix. Always use `CLong`.
+> ❌ **NEVER** use `ulong` for `size_t` — causes stack corruption on 32-bit. Use `nuint` or `UIntPtr`.
+> ❌ **NEVER** use `bool` without `MarshalAs` — the default marshal size is wrong.
```
AaronRobinsonMSFT, line 80: All of these are captured in the type-mapping.md file. I don't think they should be here.

plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -109,6 +130,8 @@ internal static partial int ProcessRecords(
 4. **Specify encoding explicitly.** Never rely on `CharSet.Auto`.
 5. **Never introduce `StringBuilder` for output buffers.**
 
+> ❌ **NEVER** rely on `CharSet.Auto` or omit string encoding — there is no safe default.
```
AaronRobinsonMSFT, line 141: This should have been captured in the type-mapping.md file. I'm really confused on this right now.

plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -137,6 +160,8 @@ internal static partial int SetName(string name);
 
 When memory crosses the boundary, exactly one side must own it — and both sides must agree.
 
+> ❌ **NEVER** free with a mismatched allocator — `Marshal.FreeHGlobal` on `malloc`'d memory is heap corruption.
```
AaronRobinsonMSFT, line 171: Ugh. This was another item that was in the original proposal. I'm very confused on what happend here.

plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
lewing: it is in the assessment https://gist.github.com/lewing/859ab56b0c37601804c03a5c601cfd8d
image
and the pr description.  It's a pattern that works but isn't a hard rule.

AaronRobinsonMSFT: I get that it is in the assessment, but what is the source of that guidance from the assessment? How do we know this is good guidance? I've been following the guidelines on anthropic and I've not seen this called out.

lewing: https://github.com/dotnet/runtime/pull/124734

lewing: > I get that it is in the assessment, but what is the source of that guidance from the assessment? How do we know this is good guidance? I've been following the guidelines on anthropic and I've not seen this called out.

Did you read the links I shared previously? https://github.com/lewing/agent-plugins/blob/main/plugins/skill-trainer/skills/skill-trainer-knowledge/references/skill-builder-knowledge.md#burying-critical-rules-as-numbered-workflow-steps I've found it to work multiple times in testing some of the weaker models.  It's a style suggestion not a requirement.  Do as you please.

https://github.com/dotnet/runtime/pull/124734

plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -50,6 +75,10 @@ The most dangerous mappings — these cause the majority of bugs:
 
 **For the complete type mapping table, struct layout, and blittable type rules**, see [references/type-mapping.md](references/type-mapping.md).
 
+> ❌ **NEVER** use `int` or `long` for C `long` — it's 32-bit on Windows, 64-bit on Unix. Always use `CLong`.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: This feedback is correct in the sense the it is implying `long` is always 32-bit. Which of course isn't the case and copilot is correct. The `CLong` type though will always do the right thing for whenever C `long` is used. I'm fine adding clarity here or ignoring this feedback.

plugins/dotnet/skills/dotnet-pinvoke/SKILL.md
```diff
@@ -50,6 +75,10 @@ The most dangerous mappings — these cause the majority of bugs:
...
+> ❌ **NEVER** use `ulong` for `size_t` — causes stack corruption on 32-bit. Use `nuint` or `UIntPtr`.
+> ❌ **NEVER** use `bool` without `MarshalAs` — the default marshal size is wrong.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: > while it’s C99 _Bool/bool that requires [MarshalAs(UnmanagedType.U1)]

Minor pedantic push back. The C standard says nothing about the size of `_Bool` or `bool`. It is always 1 byte as far as I've found, but it doesn't need to be. I think the suggest feedback by copilot is fine, but the point of the **NEVER** was to imply the nuance the suggested comment is expressing.

PR #79, opened 2026-02-23
Add refactoring-to-async skill

mrsharm:
## Summary
Adds the **refactoring-to-async** skill that teaches bottom-up async conversion patterns for .NET codebases.

## Skill Validation Results — refactoring-to-async

| Skill | Test | Baseline | With Skill | Δ | Verdict |
|-------|------|----------|------------|---|---------|
| refactoring-to-async | Refactor synchronous service to async | 3.3/5 | 5.0/5 | +1.7 | ✅ |
| refactoring-to-async | Async refactoring should not apply to CPU-bound code | 1.0/5 | 1.0/5 | 0.0 | ✅ |

**Overall improvement: +18.8%** (3 runs, not statistically significant)

## What the Skill Teaches
- **Bottom-up async conversion order** - start from leaf methods, propagate upward
- CancellationToken propagation through the entire call chain
- ConfigureAwait(false) usage in library code vs ASP.NET Core apps
- ValueTask selection criteria
- Identifying sync-over-async anti-patterns

## Why This Skill Passes
The base model frequently struggles with the correct order of async conversion - suggesting changes to calling methods first instead of leaf methods. This skill teaches the precise procedural sequence.

## Files
- src/dotnet/skills/refactoring-to-async/SKILL.md
- src/dotnet/tests/refactoring-to-async/eval.yaml + fixture files

### DIFF
--- a/src/dotnet/skills/refactoring-to-async/SKILL.md
+++ b/src/dotnet/skills/refactoring-to-async/SKILL.md
@@ -1,0 +1,186 @@
+---
+name: refactoring-to-async
+description: Convert synchronous .NET code to async/await, including proper Task propagation, cancellation support, and avoiding common async anti-patterns. Use when converting blocking I/O calls to async, fixing thread pool starvation, or modernizing sync-over-async code.
+---
+
+# Refactoring to Async
+
+## When to Use
+
+- Converting synchronous I/O-bound code to async/await
+- Fixing thread pool starvation caused by blocking calls
+- Modernizing legacy `.Result` / `.Wait()` / `.GetAwaiter().GetResult()` patterns
+- Adding `CancellationToken` support to async call chains
+
+## When Not to Use
+
+- The code is CPU-bound (async won't help; consider `Parallel.For` or `Task.Run`)
+- The synchronous code has no I/O operations
+- The user wants to parallelize work, not make it async
+
+## Inputs
+
+| Input | Required | Description |
+|-------|----------|-------------|
+| Code to refactor | Yes | The synchronous methods to convert |
+| Scope | No | Single method, class, or full call chain |
+
+## Workflow
+
+### Step 1: Identify blocking I/O calls
+
+Search for synchronous I/O patterns in the codebase:
+
+```bash
+grep -rn "\.Result\b\|\.Wait()\|\.GetAwaiter()\.GetResult()\|ReadToEnd()\|\.Read()\|\.Write(" --include="*.cs" .
+```
+
+Common blocking patterns to convert:
+
+| Synchronous | Async Replacement |
+|---|---|
+| `stream.Read(buffer)` | `await stream.ReadAsync(buffer, ct)` |
+| `stream.Write(data)` | `await stream.WriteAsync(data, ct)` |
+| `reader.ReadToEnd()` | `await reader.ReadToEndAsync(ct)` |
+| `File.ReadAllText(path)` | `await File.ReadAllTextAsync(path, ct)` |
+| `File.WriteAllBytes(...)` | `await File.WriteAllBytesAsync(..., ct)` |
+| `client.Send(request)` | `await client.SendAsync(request, ct)` |
+| `connection.Open()` | `await connection.OpenAsync(ct)` |
+| `command.ExecuteReader()` | `await command.ExecuteReaderAsync(ct)` |
+| `Thread.Sleep(ms)` | `await Task.Delay(ms, ct)` |
+| `task.Result` | `await task` |
+| `task.Wait()` | `await task` |
+
+### Step 2: Convert bottom-up
+
+Start from the lowest-level I/O calls and work upward through the call chain. This avoids sync-over-async wrappers.
+
+**Before:**
+
+```csharp
+public string GetUserData(int userId)
+{
+    var response = _httpClient.Send(new HttpRequestMessage(HttpMethod.Get, $"/users/{userId}"));
+    var body = new StreamReader(response.Content.ReadAsStream()).ReadToEnd();
+    return body;
+}
+```
+
+**After:**
+
+```csharp
+public async Task<string> GetUserDataAsync(int userId, CancellationToken ct = default)
+{
+    var response = await _httpClient.GetAsync($"/users/{userId}", ct);
+    response.EnsureSuccessStatusCode();
+    return await response.Content.ReadAsStringAsync(ct);
+}
+```
+
+### Step 3: Propagate async through the call chain
+
+Every caller of an async method must also become async. Follow the chain upward:
+
+```csharp
+// Layer 1: Data access (already converted)
+public async Task<User> GetUserAsync(int id, CancellationToken ct) { ... }
+
+// Layer 2: Business logic (convert next)
+public async Task<UserDto> GetUserProfileAsync(int id, CancellationToken ct)
+{
+    var user = await GetUserAsync(id, ct);
+    return MapToDto(user);  // sync mapping is fine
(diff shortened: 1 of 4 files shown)


src/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,186 @@
+---
+name: refactoring-to-async
+description: Convert synchronous .NET code to async/await, including proper Task propagation, cancellation support, and avoiding common async anti-patterns. Use when converting blocking I/O calls to async, fixing thread pool starvation, or modernizing sync-over-async code.
+---
+
+# Refactoring to Async
+
+## When to Use
```
danmoseley, line 8: move any part of use/not use into decription similar to
https://github.com/dotnet/runtime/pull/125005/changes#diff-ded9450821c1df27638def6250c00784c7f795e3a9c56ad13d09a34853a0d09bR3-R4
if it can avoid unnecessaryt reloading.

Possibly not all can go there. For example: user wants to parallelize work is something that can be known before loading the skill, and avoid load. Whereas possibly "code is CPU bound" might need loading the skill? but in general I think ?all of this should move up there.

src/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+This should return zero results in the refactored code paths.
+
+## Anti-Patterns to Avoid
+
+| Anti-Pattern | Problem | Correct Approach |
+|---|---|---|
+| `task.Result` or `task.Wait()` | Blocks thread, risks deadlock | `await task` |
+| `async void` methods | Exceptions crash the process | `async Task` (except event handlers) |
+| `Task.Run` wrapping async I/O | Wastes a thread pool thread | Call async method directly |
+| Missing `ConfigureAwait(false)` in libraries | Can deadlock in UI/ASP.NET sync contexts | Add `ConfigureAwait(false)` in library code |
```
danmoseley, line 167: this is missing from all the examples. maybe indicate those are app code? 

It might be worth mentioning as instructions, not just in anti-patterns -- if code is library add `.ConfigureAwait(false)` to every await. It should be used consistently or not at all.

src/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+- [ ] No `async void` methods (except UI event handlers)
+
+## Common Pitfalls
+
+| Pitfall | Solution |
+|---------|----------|
+| Deadlock after conversion | Ensure `await` is used everywhere; no `.Result` mixed with `await` |
+| Performance worse after conversion | Async adds overhead for CPU-bound work; only use for I/O |
+| Forgetting to update tests | Test methods must return `Task` and use `await` |
+| Breaking interface consumers | Consider keeping sync wrappers temporarily during staged migration |
```
danmoseley, line 185: this may be a big high level/conceptual for this skill?

src/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+| `reader.ReadToEnd()` | `await reader.ReadToEndAsync(ct)` |
+| `File.ReadAllText(path)` | `await File.ReadAllTextAsync(path, ct)` |
+| `File.WriteAllBytes(...)` | `await File.WriteAllBytesAsync(..., ct)` |
+| `client.Send(request)` | `await client.SendAsync(request, ct)` |
+| `connection.Open()` | `await connection.OpenAsync(ct)` |
+| `command.ExecuteReader()` | `await command.ExecuteReaderAsync(ct)` |
+| `Thread.Sleep(ms)` | `await Task.Delay(ms, ct)` |
+| `task.Result` | `await task` |
+| `task.Wait()` | `await task` |
+
```
danmoseley, line 53:
```suggestion
| `reader.Read()` (DbDataReader) | `await reader.ReadAsync()` |
| `command.ExecuteNonQuery()` (DbCommand) | `await command.ExecuteNonQueryAsync()` |

```
maybe

src/dotnet/tests/refactoring-to-async/eval.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - "Changed method signatures to return Task or Task<T> with Async suffix"
+      - "Added CancellationToken parameter propagation throughout the call chain"
+      - "Updated interfaces to match the async method signatures"
+      - "Did not introduce async void methods (except event handlers)"
+      - "Verified the code compiles after conversion with dotnet build"
+    expect_tools: ["bash"]
+    timeout: 120
+
+  - name: "Async refactoring should not apply to CPU-bound code"
+    prompt: "I have a method that does heavy matrix multiplication using nested for loops. Can you make it faster?"
```
danmoseley, line 25: is this really something the user would write? seems like a gimme as written. more likely the user would say "make DoIt() async so it's faster" and the AI would have to figure out that  it's CPU  bound?

src/dotnet/tests/refactoring-to-async/eval.yaml
```diff
@@ -0,0 +1,33 @@
...
+    assertions:
+      - type: "output_not_matches"
+        pattern: "(async Task|await.*ReadAsync|await.*WriteAsync)"
+    rubric:
+      - "Did NOT suggest converting CPU-bound computation to async/await"
+      - "Suggested parallelism approaches (Parallel.For, Task.Run, SIMD, or algorithmic optimization)"
+      - "Correctly identified that async is for I/O-bound work, not CPU-bound work"
+    timeout: 60
```
danmoseley, line 33: do you expect `ConfigureAwait(false)` in this UserService case? either way, should be a test for the opposite case (eg winforms) and verify in each case it's present or not as expected

src/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+Common errors after async refactoring:
+
+| Error | Fix |
+|---|---|
+| `CS4032`: `await` in non-async method | Add `async` to the method signature and return `Task` or `Task<T>` |
+| `CS0029`: Cannot convert `Task<T>` to `T` | Add `await` before the call |
+| `CS0127`: Method returns `Task` but body returns value | Change return type to `Task<T>` |
+| `CS1998`: Async method lacks `await` | Remove `async` if no awaits are needed, or the method is genuinely sync |
```
danmoseley, line 148:
```suggestion
| Error | Fix |
|---|---|
 Error | Fix |
   |---|---|
   | `CS4032`: `await` in non-async method | Add `async` to the method signature and return `Task` or `Task<T>` |
   | `CS0029`: Cannot convert `Task<T>` to `T` | Add `await` before the call |
   | `CS0127`: Method returns `Task` but body returns value | Change return type to `Task<T>` |
   | `CS1983`: Return type of async method must be void, Task, Task<T>, ValueTask, or ValueTask<T> | Wrap the return type: `async string Foo()` → `async Task<string> Foo()` |
   | `CS1998`: Async method lacks `await` | Remove `async` if no awaits are needed, or the method is genuinely sync |
   | `CS0535`: Does not implement interface member | Update the interface to match the new async signatures (or vice versa) |
   | `CS7036`: No argument given for required parameter `cancellationToken` | Add `ct` argument at call sites after adding `CancellationToken` to signatures |
   | `CS1503`: Argument type mismatch (`Task<T>` passed where `T` expected) | Add `await` at the call site |
```
if you want to add more?

PR #90, opened 2026-02-23
Add optimizing-ef-core-queries skill (+8.7% eval, near-miss)

mrsharm:
## Summary
Adds the **optimizing-ef-core-queries** skill for fixing EF Core performance issues.

## Eval Results
| Metric | Score |
|--------|-------|
| **Overall Improvement** | **+8.7%** |
| Threshold | 10% |
| **Result** | **Near-miss** (1.3% below threshold) |

## What the Skill Teaches
- N+1 query detection and fixes (Include, AsSplitQuery, Select projection)
- AsNoTracking for read-only queries
- Compiled queries for hot paths
- Common query traps (ToList before Where, Count vs Any, client-side evaluation)
- Raw SQL / FromSqlInterpolated for complex queries

## Files
- src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
- src/dotnet/tests/optimizing-ef-core-queries/eval.yaml

### DIFF
--- a/src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
+++ b/src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
@@ -93,7 +93,7 @@ var orderSummaries = await db.Orders
 | Scenario | Use |
 |----------|-----|
 | 1 level of Include | Single query (default) |
-| Multiple Includes (cartesian risk) | `AsSplitQuery()` |
+| Multiple Includes (Cartesian risk) | `AsSplitQuery()` |
 | Include with large child collections | `AsSplitQuery()` |
 | Need transaction consistency | Single query |
 


src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Step 1: Enable query logging to see the actual SQL
+
+```csharp
+// In Program.cs or DbContext configuration:
+optionsBuilder
+    .UseSqlServer(connectionString)
+    .LogTo(Console.WriteLine, LogLevel.Information)
+    .EnableSensitiveDataLogging()  // shows parameter values (dev only!)
+    .EnableDetailedErrors();
```
AndriySvyryd, line 38: EnableSensitiveDataLogging() and EnableDetailedErrors() generally aren't useful for finding perf issues

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+    "LogLevel": {
+      "Microsoft.EntityFrameworkCore.Database.Command": "Information"
+    }
+  }
+}
+```
+
+### Step 2: Fix N+1 query patterns
+
+**The #1 EF Core performance killer.** Happens when loading related entities in a loop.
```
AndriySvyryd, line 55: This is a bit overstated, the example would only affect apps that use lazy-loading. It would be more helpful to look for lazy-loading in general, as it can cause perf problems outside of obvious loops.

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+**Use `AsNoTrackingWithIdentityResolution()` when the query returns duplicate entities to avoid duplicated objects in memory.**
+
+### Step 4: Use compiled queries for hot paths
+
+```csharp
+// Define once as static
+private static readonly Func<AppDbContext, int, Task<Order?>> GetOrderById =
+    EF.CompileAsyncQuery((AppDbContext db, int id) =>
```
AndriySvyryd, line 124: This will only make a measurable difference for complex queries

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+### Step 5: Avoid common query traps
+
+| Trap | Problem | Fix |
+|------|---------|-----|
+| `ToList()` before `Where()` | Loads entire table into memory | Filter first: `.Where().ToList()` |
+| `Count()` to check existence | Scans all rows | Use `.Any()` instead |
+| `.Select()` after `.Include()` | Include is ignored with projection | Remove Include, use Select only |
+| `string.Contains()` in Where | May not translate, falls to client eval | Use `EF.Functions.Like()` for SQL LIKE |
```
AndriySvyryd, line 140: This is not a perf issue

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+- [ ] No client-side evaluation warnings in logs
+- [ ] Include/split strategy matches data shape
+
+## Common Pitfalls
+
+| Pitfall | Solution |
+|---------|----------|
+| Lazy loading silently creating N+1 | Remove `Microsoft.EntityFrameworkCore.Proxies` or disable lazy loading |
+| Global query filters forgotten in perf analysis | Check `HasQueryFilter` in model config; use `IgnoreQueryFilters()` if needed |
+| `DbContext` kept alive too long | DbContext should be scoped (per-request); don't cache it |
```
AndriySvyryd, line 175: Should also be combined with DbContext pooling

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
roji: Yes. I would generally discourage lazy loading in general - not only can it cause N+1, but it also requires sync-only I/O, which is bad for scalability and other scenarios. 

Though I do agree that lazy loading and N+1 **does** occur, it's likely one of the worst perf problems.

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
roji: Agreed - I would generally discourage using compiled queries unless there's a very confirmed, measured impact, and only as a sort of last-resort optimization. I'd leave this out of a generic "optimize EF" skill, even if it's in our docs.

src/dotnet/skills/optimizing-ef-core-queries/SKILL.md
roji: Agreed - in general there should be a seperate section for DbContext pooling, that's likely to be a much more impactful optimization than compiled queries, plus it (usually) doesn't have any impact on the code beyond just turning it on.

PR #103, opened 2026-02-24
Add dump-collect skill for modern .NET

steveisok:
## Summary

Add a Copilot CLI skill for configuring and collecting crash dumps from modern .NET applications (CoreCLR and NativeAOT) across Linux, macOS, Windows, and containers (Docker/Kubernetes).

**Scope:** Enable + collect only — no dump analysis.

### Skill files (`src/dotnet/skills/dump-collect/`)

| File | Purpose |
|------|---------|
| `SKILL.md` | Orchestrating skill (~750 tokens) with runtime detection logic, stop signals, and platform routing |
| `references/coreclr-dumps.md` | Env var configuration, `dotnet-dump collect`, `createdump` usage with decision table |
| `references/nativeaot-dumps.md` | `createdump` co-location, .NET 11+ `DOTNET_DbgCreateDumpToolPath`, OS-level fallbacks |
| `references/container-dumps.md` | Docker and K8s setup with non-root, Alpine, SELinux/AppArmor guidance |

### Eval tests (`src/dotnet/tests/dump-collect/eval.yaml`)

5 scenarios validated with `skill-validator`:

| Scenario | Baseline | With Skill | Δ |
|----------|----------|------------|---|
| CoreCLR Linux env var config | 3/5 | 5/5 | +2 |
| NativeAOT .NET 11 in Kubernetes | 1/5 | 5/5 | +4 |
| macOS NativeAOT post-crash recovery | 4/5 | 5/5 | +1 |
| CoreCLR Alpine Docker as non-root | 4/5 | 5/5 | +1 |
| Decline dump analysis (out-of-scope) | 2/5 | 5/5 | +3 |

## Key design decisions

- **Stop signals** prevent the agent from drifting into dump analysis — the biggest source of baseline quality issues
- **Decision tables** at the top of each reference enable quick tool selection without reading full content
- **Runtime detection** via `strings` / `nm` / `/proc/<pid>/maps` distinguishes CoreCLR from NativeAOT
- **NativeAOT dump type restriction**: only full dumps (`DOTNET_DbgMiniDumpType=4`) are supported for NativeAOT and CoreCLR single-file
- **`createdump` flags**: verified against [dotnet/runtime source](https://github.com/dotnet/runtime/blob/main/src/coreclr/debug/createdump/main.cpp) — `--withheap` (not `--heap`), `--normal` (not `--mini`)

## Validation

- 12 rounds of multi-model validation (Sonnet 4.5 + GPT-5.1-Codex)
- Final scores: 19/20 from both models (accuracy 5, completeness 5, clarity 4, actionability 5)
- All 5 eval scenarios pass with 5/5 vs 1-4/5 baseline

### DIFF
--- a/src/dotnet/skills/dump-collect/SKILL.md
+++ b/src/dotnet/skills/dump-collect/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: dump-collect
-description: "Configure and collect crash dumps for modern .NET applications. USE FOR: enabling automatic crash dumps for CoreCLR or NativeAOT, capturing dumps from running .NET processes, setting up dump collection in Docker or Kubernetes, using dotnet-dump collect or createdump. DO NOT USE FOR: analyzing or debugging dumps, post-mortem investigation with lldb/windbg/dotnet-dump analyze, profiling or tracing."
+description: "Configure and collect crash dumps for modern .NET applications. USE FOR: enabling automatic crash dumps for CoreCLR or NativeAOT, capturing dumps from running .NET processes, setting up dump collection in Docker or Kubernetes, using dotnet-dump collect or createdump. DO NOT USE FOR: analyzing or debugging dumps, post-mortem investigation with lldb/windbg/dotnet-dump analyze, profiling or tracing, or for .NET Framework processes."
 ---
 
 # .NET Crash Dump Collection
@@ -92,3 +92,4 @@ Follow the instructions in the loaded reference to configure or collect dumps. A
 1. **Confirm the dump output directory exists** and has write permissions before enabling collection.
 2. **Report the dump file path** back to the user after collection succeeds.
 3. **Verify configuration took effect** — for env vars, echo them; for OS settings, read them back.
+4. **Remind the user to disable automatic dumps if they were enabled temporarily** — remove or unset `DOTNET_DbgEnableMiniDump` and related env vars to avoid accumulating dump files.
--- a/src/dotnet/skills/dump-collect/references/nativeaot-dumps.md
+++ b/src/dotnet/skills/dump-collect/references/nativeaot-dumps.md
@@ -30,6 +30,8 @@ echo '/tmp/dumps/core.%e.%p.%t' | sudo tee /proc/sys/kernel/core_pattern
 mkdir -p /tmp/dumps
 ```
 
+> **Note:** `ulimit` only applies to processes started in the current shell session. For an already-running process, use on-demand collection with `gcore` (see On-Demand Dump Collection below).
+
 **Format specifiers for `core_pattern`:**
 | Spec | Meaning |
 |------|---------|
@@ -178,7 +180,7 @@ Expand-Archive "$env:TEMP\Procdump.zip" -DestinationPath "$env:TEMP\Procdump" -F
 
 **Using Task Manager:**
 1. Open Task Manager → Details tab
-2. Right-click the process → "Create dump file"
+2. Right-click the process → "Create memory dump file"
 3. Note the output path shown in the dialog
 
 ## Alternative: Using createdump with DOTNET_DbgEnableMiniDump


src/dotnet/skills/dump-collect/references/coreclr-dumps.md
```diff
@@ -0,0 +1,130 @@
...
+|-------|------|------|----------|
+| 1 | Mini | Small | Stack traces only, minimal disk usage |
+| 2 | Heap | Large | Need to inspect managed heap objects |
+| 3 | Triage | Small | Quick triage, limited data |
+| 4 | Full | Largest | Full process memory, maximum diagnostic value |
+
+### Important Notes
+
+- The dump output directory must exist before the crash — CoreCLR will not create it.
+- Format specifiers (`%p`, `%e`, `%h`, `%t`) require .NET 7+. On .NET 6, use a literal path.
```
danmoseley, line 52: can we remove everything about .NET 6 and earlier? it's very not supported

src/dotnet/tests/dump-collect/eval.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - "Uses DOTNET_DbgCreateDumpToolPath for .NET 11 NativeAOT"
+      - "Adds SYS_PTRACE capability to the container securityContext"
+      - "Mounts a volume for dump output"
+      - "Sets DOTNET_DbgMiniDumpType=4 since NativeAOT only supports full dumps"
+      - "Does not attempt to analyze dumps"
+    timeout: 60
+
+  - name: "Recover crash dump from macOS NativeAOT without createdump"
+    prompt: |
+      My NativeAOT app just crashed on macOS. I don't have createdump
```
danmoseley, line 49: how does eval work for this kind of thing. is it actually running on a mac, or it's just asking the skill what it would do if it was

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
steveisok: At first I thought the same thing. Does it cost us much to be helpful with older versions that people may still run in one form or another?

src/dotnet/tests/dump-collect/eval.yaml
steveisok: I asked your question through the CLI and it matched what I thought and observed locally:

The eval does not run on a Mac or any target platform. It's a pairwise LLM-as-judge evaluation that works like this:

 1. The validator sends the prompt to the agent twice — once as a bare baseline (no skill loaded) and once with the skill's SKILL.md + references injected into the agent's context
 2. The agent runs in a sandboxed temp directory with tool access (bash, file read/write), but permissions are locked to that sandbox — so it can't actually run dotnet, configure core dumps, etc. It mostly just produces text guidance
 3. A judge LLM then scores both outputs against the rubric items (1-5 per criterion) and does a pairwise comparison to determine which response is better
 4. The output_contains/output_matches assertions are simple string/regex checks on the agent's text output — verifying it mentions the right env vars, paths, tools, etc.

So for the macOS scenario, it's testing that the skill causes the agent to recommend checking /cores/ and ~/Library/Logs/DiagnosticReports — not that it actually runs those commands on a Mac. It's evaluating the quality of the guidance, not executing it.

danmoseley: That's what I assumed, but I can't help but wonder that even a reasonable skill/LLM would go off the rails when eg it can't find anything in those directories. eg., the rubric says "does not attempt to analyze" .. well it clearly won't inevitably as there's no dump.

So I'm wondering whether it should explicitly say "tell me what you would do if you were on a mac" and "OK, what if you found this" rather than pretending it's on a mac with a dump when neither are true.

steveisok: I added more to the eval. Doesn't hurt.

src/dotnet/skills/dump-collect/SKILL.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Step 1 — Identify the Scenario
+
+Ask or determine:
+
+1. **Goal**: Enable automatic crash dumps, or capture a dump from a running process right now?
+2. **Platform**: Linux, macOS, or Windows? Running in a container (Docker/Kubernetes)?
+3. **Runtime**: CoreCLR or NativeAOT?
+
+### Detecting CoreCLR vs NativeAOT
```
jeffschwMSFT, line 28: none of these detection methods work on Windows.  does the LLM trip over that?

steveisok: That is a good point! I think we are missing dumpbin from this list.

steveisok: Fixed - good callout

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
danmoseley: Eh, no particular opinion but it reduces tokens

src/dotnet/skills/dump-collect/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+dumpbin /clrheader <binary.exe> | Select-String "CLI Header" -Quiet
+
+# NativeAOT — no CLI header, has Redhawk symbols
+dumpbin /symbols <binary.exe> | Select-String "Rhp" -Quiet
+```
+
+**From a running process (Linux):**
+```bash
+# CoreCLR — loads libcoreclr
+grep -q "libcoreclr" /proc/<pid>/maps && echo "CoreCLR" || echo "Likely NativeAOT"
```
noahfalk: This is going to treat all CoreCLR single-file apps as if they were NativeAOT.
/proc/{pid}/exe should be the entrypoint binary for a running process so you could take that and use the same binary file triage as before.

src/dotnet/skills/dump-collect/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+# CoreCLR — loads libcoreclr.dylib
+vmmap <pid> 2>/dev/null | grep -q "libcoreclr" && echo "CoreCLR" || echo "Likely NativeAOT"
+```
+
+**From a running process (Windows PowerShell):**
+```powershell
+# CoreCLR — loads coreclr.dll
+(Get-Process -Id <pid>).Modules.ModuleName -contains "coreclr.dll"
+```
+
```
noahfalk, line 73: Perhaps we need a safety guard here that says if the app is neither CoreCLR nor NativeAOT then stop - the skill doesn't apply.

src/dotnet/skills/dump-collect/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+## Step 1 — Identify the Scenario
+
+Ask or determine:
+
+1. **Goal**: Enable automatic crash dumps, or capture a dump from a running process right now?
+2. **Platform**: Linux, macOS, or Windows? Running in a container (Docker/Kubernetes)?
+3. **Runtime**: CoreCLR or NativeAOT?
+
+### Detecting CoreCLR vs NativeAOT
```
noahfalk, line 28: This CoreCLR vs. NativeAOT seems like something we should consider refactoring into its own isolated skill in the future. Lots of scenarios may need to distinguish this, not just dump-collection.

src/dotnet/tests/dump-collect/eval.yaml
```diff
@@ -0,0 +1,152 @@
+scenarios:
```
noahfalk, line 1: The skill devoted a lot of text to distinguishing NativeAOT from CoreCLR but all the prompts pre-determined which one was being used. Perhaps some test prompts that don't do this to confirm the classification works?

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
```diff
@@ -0,0 +1,130 @@
...
+$env:DOTNET_EnableCrashReport = "1"
+```
+
+### Dump Type Reference
+
+| Value | Type | Size | Use When |
+|-------|------|------|----------|
+| 1 | Mini | Small | Stack traces only, minimal disk usage |
+| 2 | Heap | Large | Need to inspect managed heap objects |
+| 3 | Triage | Small | Quick triage, limited data |
```
noahfalk: A better description might be "The same as mini, but redacts known file paths"

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
```diff
@@ -0,0 +1,130 @@
...
+- The legacy `COMPlus_` prefix (e.g., `COMPlus_DbgEnableMiniDump`) still works but `DOTNET_` is preferred for .NET 6+.
+- Single-file published apps only support full dumps (`DOTNET_DbgMiniDumpType=4`), same as NativeAOT.
+
+## On-Demand Dump Collection
+
+### Using dotnet-dump (Recommended)
+
+```bash
+# Install (one-time)
+dotnet tool install -g dotnet-dump
```
noahfalk, line 62: This only works on machines with the SDK installed. For more limited environments we also have the web download links.

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
```diff
@@ -0,0 +1,130 @@
...
+# Collect a dump from a running process
+dotnet-dump collect -p <pid>
+
+# Specify dump type and output path
+dotnet-dump collect -p <pid> --type Full --output /tmp/dumps/myapp.dmp
+```
+
+**Supported `--type` values:** `Full`, `Heap`, `Mini`
+
+### Using createdump
```
noahfalk: I think we should leave this out. createdump is intended to be an implementation detail, not a customer facing part of the DX.

src/dotnet/skills/dump-collect/references/nativeaot-dumps.md
```diff
@@ -0,0 +1,269 @@
...
+| Automatic dump on crash (fallback) | OS-level core dumps (`ulimit`, `core_pattern`, WER) | All |
+| On-demand from running process | `gcore` (Linux), `lldb` (macOS), `procdump` (Windows) | Per-platform |
+
+NativeAOT applications are native executables, but they still support the `DOTNET_DbgEnableMiniDump` environment variables — as long as `createdump` is available.
+
+## Using createdump with NativeAOT (Preferred — All Platforms)
+
+If `createdump` is placed **next to the application binary**, the same environment variables used for CoreCLR work for NativeAOT:
+
+> **Note:** NativeAOT only supports full dumps. `DOTNET_DbgMiniDumpType` must be set to `4` (Full). The same applies to CoreCLR single-file apps.
```
noahfalk: Given this in the NativeAOT doc I don't think we need to include notes about CoreCLR single-file behavior.

src/dotnet/skills/dump-collect/references/nativeaot-dumps.md
```diff
@@ -0,0 +1,269 @@
...
+```bash
+export DOTNET_DbgCreateDumpToolPath=/opt/tools/createdump
+export DOTNET_DbgEnableMiniDump=1
+export DOTNET_DbgMiniDumpType=4
+export DOTNET_DbgMiniDumpName="/tmp/dumps/%e_%p_%t.dmp"
+```
+
+This is especially useful in containers where you can install `createdump` once in a shared location.
+
+## OS-Level Fallback (When createdump Is Not Available)
```
noahfalk: Is there a reason why the OS tools are the fallback rather than the first choice? Copying around createdump is a nuissance and its biggest value prop (smaller dumps) doesn't work on NativeAOT.

steveisok: Because ;-)

src/dotnet/skills/dump-collect/SKILL.md
steveisok: I didn't address this. I'd say let this go and see how we need to evolve.

src/dotnet/skills/dump-collect/references/container-dumps.md
```diff
@@ -0,0 +1,397 @@
...
+- **SELinux**: The container may need `--security-opt label=disable` or an appropriate SELinux policy allowing ptrace
+- **AppArmor**: Use `--security-opt apparmor=unconfined` for debugging, or create a custom profile that allows ptrace
+
+⚠️ Disabling security modules is acceptable for debugging but should not be used in production.
+
+## Troubleshooting
+
+| Problem | Cause | Fix |
+|---------|-------|-----|
+| `dotnet-dump collect` fails with "permission denied" | Missing `SYS_PTRACE` capability | Add `SYS_PTRACE` to securityContext / cap_add |
```
danmoseley, line 396: warning here to not leave on in production? (I know that is suggested on 387). similar for `--privileged` below?

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
```diff
@@ -0,0 +1,100 @@
...
+
+**On Windows (PowerShell):**
+```powershell
+$env:DOTNET_DbgEnableMiniDump = "1"
+$env:DOTNET_DbgMiniDumpType = "4"
+$env:DOTNET_DbgMiniDumpName = "C:\dumps\%e_%p_%t.dmp"
+$env:DOTNET_EnableCrashReport = "1"
+```
+
+### Dump Type Reference
```
danmoseley, line 40: seems like we say to always use 4 though

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
danmoseley: I've seen references to old .NEt in other ones so this is fine.

src/dotnet/skills/dump-collect/references/coreclr-dumps.md
```diff
@@ -0,0 +1,100 @@
...
+# Check env vars are set
+env | grep DOTNET_Dbg
+env | grep DOTNET_EnableCrashReport
+
+# Ensure dump directory exists and is writable
+ls -la /tmp/dumps/
+
+# After a crash, check for dump files
+ls -la /tmp/dumps/*.dmp
+```
```
danmoseley, line 100: should it clean up, ie., remove the env vars?

src/dotnet/skills/dump-collect/references/nativeaot-dumps.md
```diff
@@ -0,0 +1,269 @@
...
+gcore -o /tmp/dumps/myapp <pid>
+```
+
+## macOS
+
+### Automatic Crash Dumps
+
+```bash
+# Enable core dumps for the current shell session
+ulimit -c unlimited
```
danmoseley, line 92: what if the process is already running? i probably missed mention of that but I guess it's do ulimit and reboot/restart process.

src/dotnet/skills/dump-collect/references/nativeaot-dumps.md
```diff
@@ -0,0 +1,269 @@
...
+# Capture a full dump
+& "$env:TEMP\Procdump\procdump.exe" -ma <pid> C:\dumps\myapp.dmp
+
+# Capture on crash (waits for an unhandled exception)
+& "$env:TEMP\Procdump\procdump.exe" -ma -e -w <processname> C:\dumps\myapp.dmp
+```
+
+**Using Task Manager:**
+1. Open Task Manager → Details tab
+2. Right-click the process → "Create dump file"
```
danmoseley:
```suggestion
2. Right-click the process → "Create memory dump file"
```

src/dotnet/skills/dump-collect/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Stop Signals
+
+🚨 **Read before starting any workflow.**
+
+- **Stop after dumps are enabled or collected.** Do not open, analyze, or triage dump files.
+- **If the user already has a dump file**, this skill does not cover analysis. Let them know analysis is out of scope.
+- **Do not install analysis tools** (dotnet-dump analyze, windbg). Only install collection tools (dotnet-dump collect). Using `lldb` for on-demand dump capture on macOS is allowed — it ships with Xcode command-line tools and is not being used for analysis.
+- **Do not trace root cause** of crashes. Report the dump file location and move on.
+- **Do not modify application code.** Configuration is environment-only (env vars, OS settings, container specs).
```
danmoseley, line 18: Same comment about whether we should tell it how to clean up once it has the dump

src/dotnet/skills/dump-collect/SKILL.md
```diff
@@ -0,0 +1,94 @@
+---
+name: dump-collect
+description: "Configure and collect crash dumps for modern .NET applications. USE FOR: enabling automatic crash dumps for CoreCLR or NativeAOT, capturing dumps from running .NET processes, setting up dump collection in Docker or Kubernetes, using dotnet-dump collect or createdump. DO NOT USE FOR: analyzing or debugging dumps, post-mortem investigation with lldb/windbg/dotnet-dump analyze, profiling or tracing."
```
danmoseley: what 
```suggestion
description: "Configure and collect crash dumps for modern .NET applications. USE FOR: enabling automatic crash dumps for CoreCLR or NativeAOT, capturing dumps from running .NET processes, setting up dump collection in Docker or Kubernetes, using dotnet-dump collect or createdump. DO NOT USE FOR: analyzing or debugging dumps, post-mortem investigation with lldb/windbg/dotnet-dump analyze, profiling or tracing, or for .NET Framework processes."
```

src/dotnet/skills/dump-collect/SKILL.md
noahfalk: Sorry if it wasn't clear - I wasn't proposing you do anything about this now. It was a "hey this might have broader applicability so lets keep an eye out"

PR #122, opened 2026-02-25
Add skill for AOT-compatibility

agocke:
Also adds a real-world project conversion between the before and after tests.

### DIFF
--- a/plugins/dotnet/skills/dotnet-aot-compat/SKILL.md
+++ b/plugins/dotnet/skills/dotnet-aot-compat/SKILL.md
@@ -7,6 +7,7 @@ description: >
   IL3050), adding DynamicallyAccessedMembers annotations, enabling IsAotCompatible.
   DO NOT USE FOR: publishing native AOT binaries, optimizing binary size, replacing
   reflection-heavy libraries with alternatives.
+  INVOKES: no tools — pure knowledge skill.
 ---
 
 # dotnet-aot-compat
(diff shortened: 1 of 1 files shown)


plugins/dotnet/skills/dotnet-aot-compat/SKILL.md
```diff
@@ -106,7 +129,7 @@ void Process([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMe
 }
 ```
 
-When you annotate a parameter, **all callers** must now pass properly annotated types. This cascades outward — follow each caller and annotate or refactor as needed.
+When you annotate a parameter, **all callers** must now pass properly annotated types. This cascades outward — follow each caller and annotate or refactor as needed. **The caller's annotation must include at least the same member types as the callee's.** If the callee requires `PublicConstructors | NonPublicConstructors`, the caller must specify the same or a superset — using only `NonPublicConstructors` will produce IL2091.
```
agocke, line 132: This feels like strangely-specific direction...

plugins/dotnet/skills/dotnet-aot-compat/SKILL.md
```diff
@@ -215,63 +216,46 @@ Build all target frameworks to ensure:
...
+- If the user asked a scoped question (e.g., "fix warnings in this file"), don't expand to the entire project.
+
 ## Polyfills for Older TFMs
 
-`DynamicallyAccessedMembersAttribute` shipped in .NET 5. For projects targeting netstandard2.0 or net472, you need a polyfill. The trimmer recognizes the attribute by name, so a local copy works:
+For multi-targeting projects that include netstandard2.0 or net472, you need polyfills for `DynamicallyAccessedMembersAttribute` and related types. See [references/polyfills.md](references/polyfills.md).
+
+## Common Gotchas
+
+1. **External types without AOT-safe serialization**: When a type comes from a dependency you can't modify (e.g., `ResponseError` from `Azure.Core`) and it lacks a source-generated serializer, `Options.GetConverter<T>()` is reflection-based and will produce IL warnings. First check if the type implements `IJsonModel<T>` (common in Azure SDK) — if so, bypass `JsonSerializer` entirely:
```
agocke, line 235: The note about `ResponseError` looks wrong -- I think it's overfitting to the specific test case.

plugins/dotnet/skills/dotnet-aot-compat/SKILL.md
steveisok: It seems without some kind of bound, the models may get lost. Is it too weird?

plugins/dotnet/skills/dotnet-aot-compat/SKILL.md
steveisok: ResponseError was the last area the models got hung up on and couldn't finish the job. I don't know if taking this away will throw the models back into not being able to finish.

PR #124, opened 2026-02-26
Add dotnet-trace-collect skill for .NET performance diagnostics

brianrob:
Add a new skill that guides developers through capturing diagnostic artifacts for production .NET performance issues. Covers tool selection, data collection commands, and analysis recommendations across Windows, Linux, containers, and Kubernetes for both modern .NET and .NET Framework.

Includes eval.yaml with 12 test scenarios covering high CPU, memory leaks, excessive GC, hangs, slow requests, native call stacks, Windows containers, long-running repros, and environment-specific tool trade-offs.

## Skill Validation Results

| Skill | Scenario | Baseline | With Skill | Δ | Skills Loaded | Verdict |
|-------|----------|----------|------------|---|---------------|--------|
| dotnet-trace-collect | High CPU in Kubernetes on Linux (.NET 8) | 3.3/5 | 5.0/5 | +1.7 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | .NET Framework on Windows without admin privileges | 2.7/5 | 5.0/5 | +2.3 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | .NET 10 on Linux with root access and native call stacks | 1.0/5 ⏰ timeout | 4.0/5 | +3.0 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Memory leak on Linux (.NET 8) | 2.3/5 | 3.0/5 | +0.7 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Slow requests on Windows with PerfView | 3.7/5 | 5.0/5 | +1.3 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Excessive GC on Linux (.NET 8) | 3.3/5 | 5.0/5 | +1.7 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Hang or deadlock diagnosis on Linux | 2.0/5 | 4.3/5 | +2.3 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Windows container high CPU with PerfView | 1.7/5 | 5.0/5 | +3.3 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Long-running intermittent issue with PerfView triggers | 2.0/5 | 4.7/5 | +2.7 | ✅ dotnet-trace-collect; tools: skill, report_intent, view | ✅ |
| dotnet-trace-collect | Linux pre-.NET 10 needing native call stacks | 2.0/5 | 5.0/5 | +3.0 | ✅ dotnet-trace-collect; tools: skill, report_intent, view | ✅ |
| dotnet-trace-collect | Windows modern .NET with admin high CPU | 3.0/5 | 5.0/5 | +2.0 | ✅ dotnet-trace-collect; tools: skill, report_intent, view | ✅ |
| dotnet-trace-collect | Memory leak on .NET Framework Windows | 3.3/5 | 5.0/5 | +1.7 | ✅ dotnet-trace-collect; tools: skill, report_intent, view | ✅ |
| dotnet-trace-collect | Kubernetes with console access prefers console tools | 4.0/5 | 5.0/5 | +1.0 | ✅ dotnet-trace-collect; tools: skill | ✅ |
| dotnet-trace-collect | Container installation without .NET SDK | 3.0/5 | 5.0/5 | +2.0 | ✅ dotnet-trace-collect; tools: skill | ✅ |


src/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+| OS | Yes | Windows or Linux |
+| Deployment | Yes | Non-container, container, or Kubernetes |
+| Admin privileges | Recommended | Whether the developer has admin/root access on the target machine |
+| Repro characteristics | Recommended | Whether the issue is easy to reproduce or requires a long time to manifest |
+
+## Workflow
+
+### Step 1: Understand the environment
+
+Ask the developer to clarify:
```
noahfalk:
```suggestion
Determine or ask the developer to clarify:
```

I think we usually try to have the AI figure stuff out on its own and only ask the user as a last resort.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### Quick decision matrix (first-pass triage)
+
+| Environment | Preferred tool | Fallback / Notes |
+|-------------|----------------|------------------|
+| Windows + modern .NET + admin | PerfView | If admin is unavailable, use `dotnet-trace` |
+| Windows + .NET Framework + admin | PerfView | Without admin, there is no trace fallback; use dumps for hangs/memory leaks only |
+| Linux + .NET 10+ + root | `dotnet-trace collect-linux` | Use `dotnet-trace` if root or kernel prerequisites are not met |
+| Linux + pre-.NET 10 | `dotnet-trace` | Add `perfcollect` when native stacks are needed (requires root) |
+| Linux container/Kubernetes | `dotnet-monitor` sidecar | Inside-container fallback: `dotnet-trace` / `dotnet-dump`; use `collect-linux` only on .NET 10+ with root |
```
noahfalk: What if we only recommend dotnet-monitor if it is already available and the agent has no console access to the container or host? In cases where the agent could get a remote shell dotnet-trace is potentially more functional (collect-linux) and avoids setting up auth which I've seen be thorny in some scenarios.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+#### Linux (non-container, pre-.NET 10)
+
+1. **`dotnet-trace`** (preferred) — managed trace collection; no admin required.
+2. **`perfcollect`** — when **native call stacks** are needed (requires admin/root).
+
+#### Linux Container / Kubernetes
+
+1. **`dotnet-monitor`** (preferred for ease of deployment) — designed for containers; runs as a sidecar. No tools needed in the app container.
+2. **`dotnet-trace collect-linux`** (.NET 10+ with root) — produces richer traces than `dotnet-monitor` including native call stacks and kernel events, but must be run inside the container with root privileges.
+3. **`dotnet-counters`**, **`dotnet-trace`**, **`dotnet-dump`** — inside the container if the tools are installed in the image.
```
noahfalk: For dumps Steve created some good guidance in another skill. Perhaps we scope this one to traces only? The skill is called dotnet-trace-collect so it seems odd to be including non-trace diagnostic artifacts.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### Live monitoring (any environment, modern .NET)
+
+- **`dotnet-counters`** — lightweight live metrics; useful as a first step to narrow down the symptom before collecting a trace or dump.
+
+#### Memory dumps (any environment, modern .NET)
+
+- **`dotnet-dump`** — collect and perform basic analysis of memory dumps.
+
+#### Memory leaks
```
noahfalk, line 125: Should we split this into multiple skills?

One skill would focus on mapping symptoms to preferred diagnostic artifacts + analysis approach. Other dedicated skills would discuss how to best collect diagnostic artifact X in environment Y.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
...
+
+### Common Commands
+
+```bash
+# Production-safe default: keep auth enabled (default) and bind to loopback
+dotnet-monitor collect --urls http://127.0.0.1:52323 --metricUrls http://127.0.0.1:52325
+
+# Dev-only shortcut in isolated environments (avoid in production)
+dotnet-monitor collect --urls http://127.0.0.1:52323 --no-auth
+```
```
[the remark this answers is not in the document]
brianrob: Per documentation it is `--metricUrls <urls>`

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+- **Without admin privileges**: Two process dumps (`dotnet-dump`) can give a sense of what's growing on the heap, but may not be enough to identify the root cause. If dumps aren't sufficient, reproduce the issue in an environment where admin privileges are available to collect richer data (traces).
+- **Modern .NET on Linux (pre-.NET 10)**: Recommend two `dotnet-dump` captures (for heap diff) plus `dotnet-trace` while memory is growing (for allocation tracking). No trigger needed — capture during the growth period. Both together give the best picture.
+- **Modern .NET 10+ on Linux with admin**: Recommend two `dotnet-dump` captures (for heap diff) plus `dotnet-trace collect-linux` while memory is growing (richer data including native stacks). No trigger needed.
+- **.NET Framework**: Recommend two dumps plus a PerfView trace while memory is growing to see what is being allocated. No trigger is needed — just capture the trace during the growth period. Do not wait for an `OutOfMemoryException`.
+
+#### Excessive GC
```
noahfalk, line 133: Do customers often frame their symptoms as 'excessive GC'? I would have guessed the frame would be excessive CPU usage or latency and then a trace would demonstrate those symptoms were caused by excessive GC.

src/dotnet/tests/dotnet-trace-collect/eval.yaml
```diff
@@ -0,0 +1,216 @@
...
+    prompt: >-
+      My Linux .NET 8 app's memory keeps growing over hours. What exact data
+      should I collect to identify a memory leak?
+    assertions:
+      - type: "output_matches"
+        pattern: "(dotnet-dump|dotnet dump)"
+      - type: "output_matches"
+        pattern: "(dotnet-trace|dotnet trace)"
+      - type: "output_matches"
+        pattern: "(two|2|multiple|pair)"
```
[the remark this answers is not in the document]
brianrob: This should just be removed.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+Slow requests require a **thread time trace** to see where threads are spending time — waiting on locks, I/O, external calls, etc. Use larger buffers since thread time traces generate more data.
+
+- **Windows (PerfView)**: Use `PerfView /ThreadTime collect /BufferSizeMB:1024 /CircularMB:2048`. The `/ThreadTime` argument adds thread-level wait and block detail. Do not include a stop trigger by default — let the user design one based on their specific scenario.
+- **Linux (dotnet-trace)**: `dotnet-trace` captures thread time data by default — no special arguments needed. Use `dotnet-trace collect -p <PID>`.
+- **Linux .NET 10+ with root**: Use `dotnet-trace collect-linux --profile thread-time` for richer data with native stacks.
+- **Containers**: `dotnet-monitor` can capture traces via its REST API (`/trace?pid=<PID>&durationSeconds=30`).
+
+#### Hangs
+
+1. **Collect a process dump** using `dotnet-dump collect`, `procdump`, Task Manager (Windows), or `createdump` / `gcore` (Linux). Dump collection does not require admin privileges.
```
noahfalk: I know folks often start with a dump, but perhaps we want to start with a trace instead? I'd expect a trace would do a better job distinguishing slow progress/livelock/deadlock/starvation up-front. If the hang is a deadlock then the analysis could follow up with a dump if one is needed to chase locks in the dependency cycle.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+6. **Output location**: Where the collected file will be saved and how to copy it off the target for analysis.
+7. **Artifact handoff checklist**: Include runtime version, OS/kernel, container image tag or build SHA, PID/process name, UTC collection start/end timestamps, exact command used, and final artifact path when handing traces/dumps to someone else for analysis.
+
+### Step 4: Recommend analysis approach
+
+After data is collected, recommend the appropriate tool for analysis. Do **not** perform the analysis — just point the developer to the right tool and documentation.
+
+| Collected Data | Analysis Tool | Notes |
+|----------------|---------------|-------|
+| `.nettrace` file | PerfView (Windows), `dotnet-trace report` (cross-platform), Speedscope (web) | PerfView gives the richest view on Windows |
```
noahfalk: dotnet-trace report is very limited. I think we are better off pointing at speedscope, copying the file to windows, or building analysis skills that programmatically interogate the file using TraceEvent.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+
+| Pitfall | Solution |
+|---------|----------|
+| Using `dotnet-trace` on .NET Framework | `dotnet-trace` only works with modern .NET (.NET Core 3.0+). Use PerfView for .NET Framework. |
+| PerfView without admin privileges | PerfView requires admin for ETW tracing. Fall back to `dotnet-trace` if admin is not available. |
+| `perfcollect` in container without `SYS_ADMIN` | Containers drop `SYS_ADMIN` by default. Run with `--privileged` or add `SYS_ADMIN` capability, or fall back to `dotnet-trace`. |
+| Huge trace files from long repros | On Windows, use PerfView `/StopOn` triggers that fire on the symptom you want to capture (e.g., `/StopOnPerfCounter`, `/StopOnGCEvent`, `/StopOnException`) with `/CircularMB` and `/BufferSizeMB`. Do not use `/StopOn` for the recovery — the circular buffer preserves data leading up to the stop trigger. |
+| Diagnostic port not accessible in container | Mount `/tmp` as a shared volume between the app container and `dotnet-monitor` sidecar for the diagnostic Unix domain socket. |
+| Forgetting to install tools in container image | Add `dotnet tool install` to your Dockerfile, or use `dotnet-monitor` as a sidecar to avoid modifying the app image. |
```
noahfalk, line 211: dotnet tool install requires an SDK dependency which often folks wouldn't want as part of the container setup. There are ways to get around it but I think the easiest option is using an aka.ms download link with curl or wget.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
...
+| OS | Windows, Linux, macOS |
+| Runtime | Modern .NET (.NET Core 3.0+) |
+| .NET Framework | ❌ Not supported |
+| Admin required | No |
+| Container | Works inside container; use `--diagnostic-port` for sidecar |
+
+### Installation
+
+```bash
+dotnet tool install -g dotnet-trace
```
noahfalk: Only works if the SDK is installed. I think showing the aka.ms download links as a no SDK option would be useful.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
...
+
+```bash
+# List running .NET processes
+dotnet-trace ps
+
+# Collect a trace with the default profile (CPU sampling)
+dotnet-trace collect -p <PID>
+
+# Collect with a specific profile
+dotnet-trace collect -p <PID> --profile cpu-sampling
```
noahfalk: In the latest tool version cpu-sampling only works on collect-linux. You might want 'dotnet-sampled-thread-time'
https://github.com/dotnet/diagnostics/blob/main/src/Tools/dotnet-trace/CommandLine/Commands/ListProfilesCommandHandler.cs#L62

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
...
+
+### Trade-offs
+
+- ✅ Cross-platform, no admin needed, lightweight
+- ❌ No native (unmanaged) call stacks — only managed frames
+- ❌ Less system-level detail than ETW (PerfView) or perf (perfcollect)
+
+---
+
+## dotnet-trace collect-linux (.NET 10+)
```
noahfalk: If this section is dotnet-trace collect-linux then it seems like the section above should be titled dotnet-trace collect for parity.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
+# Tool Compatibility Reference
```
noahfalk: Rather than making this one file for all the tools I'd suggest breaking it into one file per tool. This avoids burning a bunch of context when the AI already knows which tool it is interested in.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
+---
+name: dotnet-trace-collect
```
mdh1418, line 2: Would this name be misconstrued as "just" `dotnet-trace collect`? If we keep this as a guide for general tracing scenarios, wondering if dotnet-tracing-tools would be appropriate, or if we expand to all `dotnet-diagnostic-tools`. But if we split into one skill per tool, then nevermind this comment.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+- Benchmarking during development (e.g. BenchmarkDotNet setup)
+- Analyzing collected trace or dump files (this skill recommends tools for analysis, but does not perform it)
+
+## Inputs
+
+| Input | Required | Description |
+|-------|----------|-------------|
+| Symptom | Yes | What the developer is observing (high CPU, memory growth, slow requests, hangs, excessive GC, etc.) |
+| Runtime | Yes | .NET Framework or modern .NET (and version, especially whether .NET 10+) |
+| OS | Yes | Windows or Linux |
```
mdh1418, line 30: Would there be value in including mobile scenarios in this PR?

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
...
+**Purpose**: Collects diagnostic traces using `perf_events`, a Linux OS technology. Provides native call stacks, kernel events, and machine-wide tracing that standard `dotnet-trace collect` cannot.
+
+| Attribute | Value |
+|-----------|-------|
+| OS | Linux only (kernel >= 6.4 with `CONFIG_USER_EVENTS=y`) |
+| Runtime | .NET 10+ only |
+| .NET Framework | ❌ Not supported |
+| Admin required | Yes (root) |
+| Container | Works inside container (requires root) |
+| glibc | >= 2.35 (not supported on Alpine 3.22, CentOS Stream 9, or RHEL 9-based distros) |
```
mdh1418: We have a few constraints/known issues that aren't documented right now. (e.g. can't trace cross-namespace processes, requiring symbols to be on disk during tracing)

Are skills allowed to point agents to fetch learn.microsoft docs? I imagine we would be updating the docs rather than updating the docs and skills every time we fix or discover a constraint.

Also, will agents always update the tools to the latest versions?

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
```diff
@@ -0,0 +1,691 @@
...
+
+# Output in Speedscope format for web-based viewing
+dotnet-trace collect -p <PID> --format speedscope
+```
+
+### Output Formats
+
+| Format | Extension | Analysis Tool |
+|--------|-----------|---------------|
+| `nettrace` (default) | `.nettrace` | PerfView, Visual Studio, `dotnet-trace report` |
```
mdh1418: Should we include TraceEvent?
In an ideal world, how could agents analyze traces themselves rather than prompting users to view them visually in PerfView/Visual Studio as a first step. (this would help in case an "empty" trace is collected by accident, yet the agent isn't aware that it doesn't contain the desired events).

Is there room for a more powerful `dotnet-trace report` where agents can quickly probe .nettraces for data, rather than needing every agent to generate their own TraceEvent-based analysis app?

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
mdh1418: Will agents using this skill automatically seek the help prompt? Similar to https://github.com/dotnet/skills/pull/124/changes#r2860158281, if agents are taking this set as the only commands available, we'd need to maintain this doc as often as the tool updates (as we should be doing with the official docs).

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: Surprisingly, yes.  There are many cases where customers come to us because they notice that the GC counters (perf counters or event counters) are higher than usual, or downright excessive.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: Updated in next commit.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: Updated in next commit.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: Updated in next commit. (Delegates to dump-collect skill).

src/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: I like this idea.  Updated in the next commit.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: Do we have aka.ms links for dotnet-monitor?  I couldn't find any.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
brianrob: Fixed in the next commit.

@mdh1418, I don't expect it to always pull the latest help.  I suspect it might do this if a command that it's looking for doesn't exist or fails for some reason related to the set of arguments.

In terms of keeping this up-to-date, I think that as long as we're adding new scenarios and old scenarios continue to work, an update is not required unless we feel that the new functionality should be referenced from this skill, or replace a pre-existing capability.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
brianrob: Seems reasonable.  Fixed in next commit.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
brianrob: Great idea.  Looks like we do this in dump-collect too.  This makes a pretty good improvement in context usage.  Fixed in next commit.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: I was trying to mimic the naming dump-collect.  That said, I haven't seen this problem in any of my testing, so I'm not super concerned.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: Thanks for recommending this.  After experimenting with this, we're going to delegate that to a separate skill since it really bloats the skill.md in a material way (about 2X)

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
brianrob: This is a good point.  I don't think agents will update tools unless we tell them to.  My thought is to document these limitations here and then when we fix them, we'll update the skill.  Fixed in next commit.

src/dotnet/skills/dotnet-trace-collect/references/tool-compatibility.md
brianrob: Fixed in next commit.

src/dotnet/skills/dotnet-trace-collect/SKILL.md
brianrob: I've removed it from the recommended list.

plugins/dotnet/skills/dotnet-trace-collect/SKILL.md
noahfalk: I'm not aware it has one. It only has dotnet-tool install or docker image (docker pull mcr.microsoft.com/dotnet/monitor)

brianrob: Thanks for confirming, @noahfalk.

PR #133, opened 2026-02-26
Adding scaffold-generate-aspnet/SKILL.md

sayedihashimi:
This should be picked up when a prompt requests to generate content for ASP.NET Core that uses a database. This skill will ensure that best practices are followed when using ASP.NET Core + EF. For example without this skill it's common for copilot to initialize the database in Program.cs instead of using migrations.

### DIFF
--- a/src/dotnet/skills/scaffold-generate-aspnet/SKILL.md
+++ b/src/dotnet/skills/scaffold-generate-aspnet/SKILL.md
@@ -179,3 +179,8 @@ Skip this step for non-API scaffolders.
 | `dotnet ef` not found | Try `dotnet tool restore` first; only install globally as fallback |
 | Added unnecessary packages | Only add packages explicitly required for the scaffolded functionality |
 | Generated code uses different naming conventions | Inspect existing project files to match naming patterns before generating |
+
+## References
+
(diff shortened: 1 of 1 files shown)


src/dotnet/skills/scaffold-generate-aspnet/SKILL.md
```diff
@@ -0,0 +1,181 @@
+---
+name: scaffold-generate-aspnet
+description: Generate or scaffold ASP.NET Core code — Razor Pages, Blazor components, MVC controllers, views, and Minimal API endpoints — entirely through Copilot without CLI tools. Use when (1) adding CRUD pages, views, or API endpoints backed by Entity Framework (EF Core) and a database, (2) generating code to create, read, update, and delete data using a DbContext, (3) scaffolding UI components that match the project's existing CSS framework and coding patterns, or (4) creating data-driven forms, tables, and navigation for a model class. Do not use for non-ASP.NET projects or when CLI-based scaffolding with `dotnet scaffold` is preferred.
+---
+
+# Generate or Scaffold ASP.NET Core Code with Copilot
```
timheuer: As a general skill, specifying Copilot might not be great here. 'entirely through Copilot' seems confusing.

sayedihashimi: I have fixed this.

PR #141, opened 2026-02-27
[Infra] Add devops agentic workflow

JanKrivanek:
### Motivation

Adding a daily run of agentic workflow that collects and maintain operational status of the repo in a form of pinned issue

### Detailed description

https://github.com/dotnet/skills/blob/f78a07be72227cfb063c362606afe6a29ba5ffe1/eng/agentic-workflows/README.md

### Sample output

In the [first comment](https://github.com/dotnet/skills/pull/141#issuecomment-3973646193)


src/dotnet-msbuild/agentic-workflows/README.md
ViktorHofer, line 1: This README describes the shipping component. Which should not point to non-shipping stuff here.

.github/workflows/devops-health-investigate.yml
ViktorHofer, line 1: Why the layered approach? Why not directly using the gh agentic workflow?

JanKrivanek: The tokens rotation is impossible to get right with gh-aw ... but - let's maybe have it currently without that rotation, so taht we can simplify 👍

PR #147, opened 2026-02-27
Add implementing-server-sent-events skill

mrsharm:
## New Skill: implementing-server-sent-events

Adds a skill for implementing Server-Sent Events (SSE) endpoints in ASP.NET Core 8 minimal APIs.

### Key Gotchas Covered
- **No built-in MapSSE() or MapServerSentEvents()** — must use manual response writing
- Must set Content-Type: text/event-stream, Cache-Control: no-cache, Connection: keep-alive
- SSE format requires double newline (`\n\n`) to terminate events
- **Must call FlushAsync()** after each event write (otherwise client receives nothing)
- Must use `HttpContext.RequestAborted` for clean disconnect handling
- Must set `X-Accel-Buffering: no` for reverse proxy compatibility
- Support `Last-Event-ID` header for client reconnection
- Must NOT return Results.Ok() after streaming events

### Eval Results (3-run validation)
- **Overall: +13.6% improvement — PASSED**
- BL=4.0 SK=4.7
- Pairwise: skill wins consistently

### DIFF
--- a/src/dotnet/skills/implementing-server-sent-events/SKILL.md
+++ b/src/dotnet/skills/implementing-server-sent-events/SKILL.md
@@ -1,0 +1,207 @@
+---
+name: implementing-server-sent-events
+description: Implement Server-Sent Events (SSE) endpoints in ASP.NET Core. Use when building real-time streaming from server to client without WebSockets.
+---
+
+# Implementing Server-Sent Events (SSE) in ASP.NET Core
+
+## When to Use
+- Server-to-client real-time push (notifications, live updates, streaming progress)
+- When you DON'T need bidirectional communication (use WebSockets for that)
+- SSE is simpler than WebSockets and works over standard HTTP
+- Automatic reconnection built into EventSource browser API
+
+## When Not to Use
+- Bidirectional communication needed → use WebSockets
+- Binary data streaming → use WebSockets or gRPC streaming
+- Need more than 6 concurrent connections per domain in HTTP/1.1 → use HTTP/2
+
+## Inputs
+
+| Input | Required | Description |
+|-------|----------|-------------|
+| Event data source | Yes | The data to stream (IAsyncEnumerable, Channel, timer, etc.) |
+| Event types | No | Named event types for `event:` field |
+| Client reconnection | No | Whether to support Last-Event-ID reconnection |
+
+## Workflow
+
+### Step 1: CRITICAL — There Is No Built-In MapSSE() or MapServerSentEvents()
+
+ASP.NET Core has NO built-in SSE endpoint helper. You must manually write the SSE protocol.
+
+```csharp
+// COMMON MISTAKE: Trying to use a non-existent API
+// app.MapSSE("/events", ...);                    // DOES NOT EXIST
+// app.MapServerSentEvents("/events", ...);       // DOES NOT EXIST
+// app.UseServerSentEvents();                     // DOES NOT EXIST
+
+// CORRECT: Use a standard minimal API endpoint with manual SSE protocol
+app.MapGet("/events", async (HttpContext context, CancellationToken ct) =>
+{
+    // CRITICAL: Set these three headers for SSE
+    context.Response.Headers.ContentType = "text/event-stream";
+    context.Response.Headers.CacheControl = "no-cache";
+    context.Response.Headers["Connection"] = "keep-alive";
+
+    // CRITICAL: Disable response buffering for reverse proxies (nginx, etc.)
+    context.Response.Headers["X-Accel-Buffering"] = "no";
+
+    await context.Response.Body.FlushAsync(ct);
+
+    // Stream events...
+});
+```
+
+### Step 2: CRITICAL — SSE Protocol Format
+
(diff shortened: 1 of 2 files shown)


src/dotnet/skills/implementing-server-sent-events/SKILL.md
```diff
@@ -0,0 +1,207 @@
+---
+name: implementing-server-sent-events
+description: Implement Server-Sent Events (SSE) endpoints in ASP.NET Core. Use when building real-time streaming from server to client without WebSockets.
+---
+
+# Implementing Server-Sent Events (SSE) in ASP.NET Core
+
+## When to Use
```
danmoseley, line 8: move when to use and when not to use entirely or largely into the description, which is the thing the ai reads when deciding whether or not to load. probably shouldn't have this section here.

src/dotnet/skills/implementing-server-sent-events/SKILL.md
```diff
@@ -0,0 +1,207 @@
...
+### Step 2: CRITICAL — SSE Protocol Format
+
+The SSE format has strict rules. Each field ends with `\n`, and each event ends with `\n\n` (double newline).
+
+```csharp
+// CRITICAL: The SSE format is NOT just "send text"
+// Each event MUST end with TWO newlines (\n\n)
+
+// Simple data event:
+await context.Response.WriteAsync($"data: {message}\n\n", ct);
```
danmoseley, line 65: should `{message}` be sanitized for newlines here and below. AI says - risk of injection attack

src/dotnet/skills/implementing-server-sent-events/SKILL.md
danmoseley, line 8: SSE endpoints hold long-lived HTTP connections. Without connection limits, an attacker (or misbehaving client) can exhaust server resources by opening many concurrent connections. Consider adding a note about:
- Enforcing per-client or total connection limits (e.g., via middleware or a `SemaphoreSlim`)
- Requiring authentication before establishing the SSE connection
- Setting an idle timeout to drop stale connections

This is especially relevant for a skill doc since agents will copy the pattern as-is without thinking about production hardening.

src/dotnet/skills/implementing-server-sent-events/SKILL.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Common Mistakes
+
+1. **Using a non-existent MapSSE() or MapServerSentEvents() method**: ASP.NET Core has no built-in SSE helper. You must manually set headers and write the SSE protocol format.
+2. **Forgetting double newline**: Events MUST end with `\n\n`. A single `\n` means the event is not complete and the client won't process it.
+3. **Not flushing**: Without `FlushAsync()` after each event, the response is buffered and the client receives nothing until disconnect.
+4. **Not handling RequestAborted**: The loop runs forever if you don't check `context.RequestAborted`. `OperationCanceledException` on disconnect is normal.
+5. **Returning a result after streaming**: Don't `return Results.Ok()` after writing SSE events — the response body is already being written.
+6. **Missing Content-Type header**: Must be exactly `text/event-stream`, not `application/json` or `text/plain`.
+7. **Missing X-Accel-Buffering: no**: Reverse proxies (nginx) buffer responses by default, breaking SSE.
```
danmoseley, line 207: Worth adding a note about CORS configuration. SSE endpoints are frequently consumed cross-origin via the browser `EventSource` API, which does not support custom headers. Without `app.UseCors(...)` or a CORS policy on the endpoint, cross-origin clients will silently fail. This is a common deployment surprise.

src/dotnet/skills/implementing-server-sent-events/SKILL.md
danmoseley: (says AI)

src/dotnet/skills/implementing-server-sent-events/SKILL.md
danmoseley: (says AI)

src/dotnet/skills/implementing-server-sent-events/SKILL.md
```diff
@@ -0,0 +1,207 @@
...
+|-------|----------|-------------|
+| Event data source | Yes | The data to stream (IAsyncEnumerable, Channel, timer, etc.) |
+| Event types | No | Named event types for `event:` field |
+| Client reconnection | No | Whether to support Last-Event-ID reconnection |
+
+## Workflow
+
+### Step 1: CRITICAL — There Is No Built-In MapSSE() or MapServerSentEvents()
+
+ASP.NET Core has NO built-in SSE endpoint helper. You must manually write the SSE protocol.
```
BrennanConroy, line 31: Wrong. We have `TypedResults.ServerSentEvents`
https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#support-for-server-sent-events-sse

Most of this Skill should probably be rewritten to use `TypedResults.ServerSentEvents`

PR #176, opened 2026-03-03
Add --run-in-docker to skill-validator to run Copilot CLI in a docker container

caaavik-msft:
# Summary

This PR adds an optional Docker execution mode to `skill-validator` so agent runs, judges, and setup commands can execute in an isolated container instead of directly on the host machine.

## Motivation

The main use case for this is for local development, but it might also be useful for running in CI if we want to build on top of it. I was building some skills and found that when using some weaker models, they made destructive changes to my host system to accomplish the task (e.g. reinstalling .NET). With this, agents and judges run inside a container with only access to the files they need bound to the host machine. This does not add any additional security measures for network isolation.

## Implementation

This makes use of the `--headless` mode for running copilot as described here: https://github.com/github/copilot-sdk/blob/main/docs/guides/setup/backend-services.md.

It requires a `GITHUB_TOKEN` be present to pass into the container so that it can use that to authenticate to the Copilot API. I have an example in the `README` which explains that you can get this token with `gh auth token`. For people with multiple gh accounts (e.g. personal and enterprise), you can also do `gh auth token --user <name>`.

A Dockerfile is included in the repo to use as the base image:

```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build

ARG COPILOT_SDK_VERSION
RUN dotnet new console -o /tmp/dl \
    && dotnet add /tmp/dl package GitHub.Copilot.SDK --version $COPILOT_SDK_VERSION \
    && dotnet build /tmp/dl -c Release \
    && cp /tmp/dl/bin/Release/net10.0/runtimes/*/native/copilot /usr/local/bin/copilot \
    && chmod +x /usr/local/bin/copilot \
    && rm -rf /tmp/dl

RUN copilot --version
```

This ensures that we use the exact same Copilot CLI binary that is shipped with the SDK. The SDK version is resolved programmatically inside the SkillValidator so it is kept in sync. It places the copilot binary at `/usr/local/bin/copilot` inside the container.

To handle path mapping/translation, when running in docker mode, all temp/work directories are placed inside a single directory in the TMP folder, and that entire directory is mounted into the container with read-write. This makes it easy to map paths to and from the host and container equivalent when needed. Skill directories are also mounted into the container with read-only access, and only the directories that are being evaluated will be mounted.

The container uses a randomised port `-p 0:4321` which is resolved later using `docker port`. The container is always cleaned up after finishing, including on `ProcessExit` and `CancelKeyPress` events.

## Future Extensibility

I have a proof of concept working locally which I chose not to push for now to keep this PR simple which runs all agents inside their own containers rather than having a single container that is used to run all agents and judges. This would help reduce any risks of agents modifying the environment and impacting other evaluations if that sounds desirable, but it does mean that each agent would use a separate `CopilotClient` rather a single shared `CopilotClient`.


eng/skill-validator/src/Services/DockerCopilotServer.cs
```diff
@@ -0,0 +1,380 @@
...
+        // Check skill dir mounts
+        foreach (var (hostSkillDir, containerMount) in _skillMounts)
+        {
+            var relativePath = Path.GetRelativePath(hostSkillDir, fullPath);
+            if (!Path.IsPathRooted(relativePath) &&
+                !relativePath.StartsWith(".." + Path.DirectorySeparatorChar, StringComparison.Ordinal) &&
+                relativePath != "..")
+            {
+                return Path.Combine(containerMount, relativePath).Replace("\\", "/");
+            }
```
JanKrivanek: NIT: This repeats and should be extracted to a helper function.

Also the separators normalization part might be needed regardles of path being rooted

eng/skill-validator/src/Services/DockerCopilotServer.cs
```diff
@@ -0,0 +1,380 @@
...
+
+        var hostDir = GetHostDir();
+        Directory.CreateDirectory(hostDir);
+
+        var runArgs = new List<string>
+        {
+            "run",
+            "--name", containerName,
+            "-p", $"0:{InternalPort}", // Map internal port to a random host port
+            "-e", "GITHUB_TOKEN",
```
JanKrivanek, line 250: This feels like anything within the container (e.g. misbehaving skill) can now use/flush the token.
cc: @ViktorHofer

caaavik-msft: I just checked to see how the Copilot SDK uses `GitHubToken` and it looks like it just passes it in via an environment variable anyways https://github.com/github/copilot-sdk/blob/4e1499dd23709022c720eaaa5457d00bf0cb3977/dotnet/src/Client.cs#L1003-L1006. So I think it's not going to be practical to hide that. I think Copilot CLI would be the only practical place to fix this by having it hide the env var from any child subprocesses it creates. Also with full shell access, a misbehaving skill could always just run `gh auth token` to get a new token.

eng/skill-validator/src/Services/DockerCopilotServer.cs
caaavik-msft: Made this change, although the non-rooted path returns an exception. If the relative path is rooted, it means that they are on different drives. This code just means that only paths that are mapped inside the container will return without exception.

eng/skill-validator/src/Services/DockerCopilotServer.cs
caaavik-msft: There is a sample here actually showing a way to solve this properly, but it would make things too complicated: https://github.com/github/copilot-sdk/tree/main/test/scenarios/bundling/container-proxy

ViktorHofer: > I think Copilot CLI would be the only practical place to fix this by having it hide the env var from any child subprocesses it creates.

Damn, we should definitely fix this in the Copilot SDK. Would you mind filing an issue?

Overall, as long as we don't use this in CI where we have less control over the PAT, this is fine.

caaavik-msft: @ViktorHofer I don't think this is an issue with the SDK, it's a limitation of the CLI which the SDK wraps as using GITHUB_TOKEN means that you are wanting to pass it in via an environment variable, that's the intended behaviour of the feature. Copilot SDK documents 6 different ways to auth: https://github.com/github/copilot-sdk/blob/main/docs/auth/index.md#authentication-priority

I had a look at what the Copilot SDK itself uses for it's tests and it uses the COPILOT_HMAC_KEY approach instead, maybe we should prefer that instead https://github.com/github/copilot-sdk/blob/11dde6e4de841bc8437b24f49e37591858dea914/.github/workflows/dotnet-sdk-tests.yml#L83

PR #182, opened 2026-03-04
Add migrate-dotnet8-to-dotnet9 skill

danmoseley:
_Recreated from #154 (original PR was orphaned when the fork relationship with dotnet/skills was re-established)._

# Add .NET 8 → .NET 9 Migration Skill

Adds a new skill to guide agents through migrating .NET 8 projects to .NET 9, systematically resolving all breaking changes. Modeled after the migrate-dotnet9-to-dotnet10 skill structure.

## Structure

- **SKILL.md** — 6-step migration workflow (TFM update → build → source fixes → behavioral fixes → infrastructure → verify)
- **10 reference documents** covering all .NET 9 breaking change categories:
  - `csharp-compiler.md` — C# 13 compiler changes (InlineArray on records, iterator safe context, collection expressions)
  - `core-libraries.md` — params span overloads, SYSLIB0054–0057, BigInteger limits, String.Trim removal
  - `sdk-msbuild.md` — Terminal Logger, VS version requirements, out-of-support TFM warnings
  - `aspnet-core.md` — ValidateOnBuild/ValidateScopes, ForwardedHeaders, middleware changes
  - `efcore.md` — Pending migrations exception, explicit transactions, Cosmos DB discriminator/ rename, sync I/O
  - `cryptography.md` — X509CertificateLoader migration (SYSLIB0057)
  - `serialization-networking.md` — BinaryFormatter removal, JsonDocument null, HttpClientFactory SocketsHttpHandler
  - `winforms-wpf.md` — WFO1000 security analyzers, PictureBox HttpRequestException, StatusStrip
  - `containers-interop.md` — zlib removal from images, CET support, Monitor tag changes
  - `deployment-runtime.md` — Env var precedence reversal, floating-point saturation, empty env vars
- **eval.yaml** — 12 scenarios with setup files, assertions, and rubrics

## Key design decisions

- **BinaryFormatter**: Skill instructs agent to stop and ask the user which replacement serializer to use (major architectural decision) rather than picking one
- **Reference delegation**: SKILL.md acts as a concise index pointing to reference docs for detailed fix guidance, matching the 9→10 pattern
- **`expect_activation: false`**: This is a pure knowledge/reference skill with no MCP tools. Activation via `SkillInvokedEvent` is unreliable for this skill type — the skill content influences agent output through `SkillDirectories` injection but doesn't always trigger explicit invocation events.

## Eval Results (claude-opus-4.6, 3 runs, judged by claude-opus-4.6)

| Scenario | Baseline | With Skill | Δ | Verdict |
|----------|----------|------------|---|---------|
| Empty env vars, ZIP encoding, keyed DI services | 2.3/5 | 3.0/5 | **+0.7** | ✅ |
| C# 13 compiler (InlineArray, iterator, collection expr) | 3.0/5 | 3.0/5 | 0.0 | ❌ |
| ASP.NET Core (DI validation, forwarded headers, HttpClientFactory) | 2.7/5 | 2.0/5 | -0.7 | ❌ |
| EF Core (migrations, transactions, Cosmos discriminator) | 3.0/5 | 3.0/5 | 0.0 | ✅ |
| EF Core Cosmos DB (id format, HasIndex, sync I/O) | 3.7/5 | 4.3/5 | **+0.6** | ❌ |
| JsonDocument null, BinaryFormatter, UserAgent nullable | 5.0/5 | 5.0/5 | 0.0 | ✅ |
| CI pipeline (Terminal Logger, VS version, net7.0 warnings) | 2.3/5 | 2.0/5 | -0.3 | ✅ |
| WinForms (WFO1000, PictureBox, StatusStrip) | 3.3/5 | 3.3/5 | 0.0 | ❌ |
| Containers (zlib removal, env var precedence, Monitor) | 3.3/5 | 4.3/5 | **+1.0** | ✅ |
| EF Core Cosmos DB (discriminator, sync I/O, HasIndex) | 4.3/5 | 4.0/5 | -0.3 | ❌ |
| Library (String.Trim, keyed services, InlineArray size) | 2.7/5 | 3.0/5 | **+0.3** | ✅ |
| Container env var precedence + zlib | 4.0/5 | 4.7/5 | **+0.7** | ❌ |

**Observations:**
- Skill adds measurable value on **obscure/niche changes** (empty env vars, Cosmos DB id format, zlib removal, ZIP UTF-8 flag, keyed services no-fallback) where Opus doesn't reliably know the answer from training data
- Skill adds overhead on **well-known changes** (ASP.NET Core, HttpClientFactory) where Opus's base knowledge is already strong
- Overfitting score clean at 0.09 (well below 0.25 threshold)
- Statistical significance requires more runs (3 runs directionally positive but not significant)

### DIFF
--- a/plugins/dotnet-upgrade/skills/migrate-dotnet8-to-dotnet9/SKILL.md
+++ b/plugins/dotnet-upgrade/skills/migrate-dotnet8-to-dotnet9/SKILL.md
@@ -135,7 +135,7 @@ Behavioral changes do not cause build errors but may change runtime behavior. Re
 
 1. **Floating-point to integer conversions are now saturating** — Conversions from `float`/`double` to integer types now saturate instead of wrapping on x86/x64. See `references/deployment-runtime-dotnet8to9.md`.
 
-2. **EF Core: Pending model changes exception** — `Migrate()`/`MigrateAsync()` now throws if the model has pending changes. See `references/efcore-dotnet8to9.md`.
+2. **EF Core: Pending model changes exception** — `Migrate()`/`MigrateAsync()` now throws if the model has pending changes. **Search for `DateTime.Now`, `DateTime.UtcNow`, or `Guid.NewGuid()` in any `HasData` call — these must be replaced with fixed constants** (e.g., `new DateTime(2024, 1, 1, 0, 0, 0, DateTimeKind.Utc)`). See `references/efcore-dotnet8to9.md`.
 
 3. **EF Core: Explicit transaction exception** — `Migrate()` inside a user transaction now throws. See `references/efcore-dotnet8to9.md`.
 


plugins/dotnet/skills/migrate-dotnet8-to-dotnet9/references/efcore.md
```diff
@@ -0,0 +1,134 @@
+# Entity Framework Core 9 Breaking Changes
+
+These changes affect projects using EF Core or Microsoft.Data.Sqlite.
```
AndriySvyryd:
```suggestion
These changes affect projects using EF Core.
```

plugins/dotnet-upgrade/skills/migrate-dotnet8-to-dotnet9/SKILL.md
```diff
@@ -0,0 +1,240 @@
+---
+name: migrate-dotnet8-to-dotnet9
+description: >
```
T-Gro, line 3: A skills description is loaded into the context every single time, for any user of the plugin. It is the trigger which decides if rest of the skills .md should be loaded or not.

I would suggest to:
1. Have the description as vanilla "Migrate from net8 to net9"
2. Have the remarks and all the "DO NOT USE FOR" in the skills body. That means the context cost will be only paid if someone is doing a migration.
3. Choose between providing a link to published document about breaking changes vs. restating what the document says. If we do both, a web search enabled agent will read the web page as well as the .md here

danmoseley: Thanks @T-Gro . I did experiment with variations, including what you said (short description and do not use in the body). I ended up with it in the description because in evaluation the skill was not getting loaded in obviously relevant scenarios.

For your 2, that's a good point, but also the skill would be loaded in unecessary situations when the user IS doing a migration. That may be the right trade off of course.

For your 3, also a good point. I think we're learning here. Most skills have chosen to not refer to docs for "core functionality" but maybe they should?

I'm inclined to leave things as is, as we're learning here and it's easy to change later. We presumably don't have many users yet, and hopefully they'll give us feedback in this repo and we can adjust. By the way these things are distributed, we ship almost continuously

BTW perhaps grouping into a migration agent might help the LLM find and load the right skill, without so much tuning of description?

BrennanConroy (approved):

.github/CODEOWNERS
```diff
@@ -30,6 +30,9 @@
 /plugins/dotnet-upgrade/skills/dotnet-aot-compat/ @agocke @dotnet/appmodel
 /tests/dotnet-upgrade/dotnet-aot-compat/ @agocke @dotnet/appmodel
 
+/plugins/dotnet-upgrade/skills/migrate-dotnet8-to-dotnet9/ @danmoseley @ViktorHofer
+/tests/dotnet-upgrade/migrate-dotnet8-to-dotnet9/ @danmoseley @ViktorHofer
```
ViktorHofer: I'm not sure why my name is here but ok :D

.github/CODEOWNERS
```diff
@@ -30,6 +30,9 @@
 /plugins/dotnet-upgrade/skills/dotnet-aot-compat/ @agocke @dotnet/appmodel
 /tests/dotnet-upgrade/dotnet-aot-compat/ @agocke @dotnet/appmodel
 
+/plugins/dotnet-upgrade/skills/migrate-dotnet8-to-dotnet9/ @danmoseley
+/tests/dotnet-upgrade/migrate-dotnet8-to-dotnet9/ @danmoseley
```
ViktorHofer: We should have a second IC here or a team. I know that you picked me before but maybe we can find someone who is on a team that owns these scenarios.

PR #200, opened 2026-03-04
Add migrating-newtonsoft-to-system-text-json skill

mrsharm:
## Summary
Adds the **migrating-newtonsoft-to-system-text-json** skill for migrating from Newtonsoft.Json (Json.NET) to System.Text.Json.

> **Note:** Reopened from #89 (accidentally closed, could not reopen due to EMU restrictions).

## Skill Validation Results — migrating-newtonsoft-to-system-text-json

| Skill | Test | Baseline | With Skill | Δ | Verdict |
|-------|------|----------|------------|---|---------|
| migrating-newtonsoft-to-system-text-json | Migrate model with Newtonsoft.Json attributes to System.Text.Json | 4.0/5 | 4.3/5 | +0.3 | ✅ |

**Overall improvement: +10.8%** (3 runs, not statistically significant)

3 Iterations.

## What the Skill Teaches
- Complete attribute mapping table (JsonProperty -> JsonPropertyName, NullValueHandling -> JsonIgnoreCondition, etc.)
- Behavioral differences that break silently (default casing, strict parsing, case sensitivity, numbers-in-strings)
- Custom converter migration (Newtonsoft JsonConverter -> System.Text.Json JsonConverter with Utf8JsonReader/Writer)
- JToken/JObject -> JsonDocument/JsonElement (read-only) or JsonNode (mutable DOM)
- Polymorphic serialization with [JsonDerivedType] (.NET 7+)
- JsonExtensionData Dictionary value type change (JToken -> JsonElement)

## Files
- src/dotnet/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
- src/dotnet/tests/migrating-newtonsoft-to-system-text-json/eval.yaml


plugins/dotnet/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,249 @@
...
+    options.NumberHandling = JsonNumberHandling.AllowReadingFromString; // Newtonsoft coerces
+    options.ReadCommentHandling = JsonCommentHandling.Skip;     // Newtonsoft allows
+    options.AllowTrailingCommas = true;                         // Newtonsoft allows
+    options.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull; // Common Newtonsoft setting
+
+    // Enum string serialization (replaces StringEnumConverter)
+    options.Converters.Add(new JsonStringEnumConverter(JsonNamingPolicy.CamelCase));
+
+    // Handle circular references (replaces PreserveReferencesHandling)
+    options.ReferenceHandler = ReferenceHandler.IgnoreCycles; // or Preserve for $ref/$id
```
GrabYourPitchforks: Setting `Preserve` significantly increases the attack surface of JSON deserialization. Most call sites do not know how to properly reason about the increased attack surface and could find themselves in a vulnerable state. This is why S.T.J does not support `$ref` / `$id` in its default configuration.

It's fine if people want to set this, but they should do it only if they truly need it and have thought through the consequences of setting it. It shouldn't be promoted as a "try it and see if it solves your problem" mechanism. (What happens in practice is that somebody will set it, find that it doesn't solve their problem, then move on to some other diagnostic step _without reverting this change._)

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,249 @@
...
+        ConfigureJsonOptions(options.JsonSerializerOptions);
+    });
+
+static void ConfigureJsonOptions(JsonSerializerOptions options)
+{
+    // Match Newtonsoft.Json default behavior:
+    options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase; // Newtonsoft default
+    options.PropertyNameCaseInsensitive = true;                 // Newtonsoft default
+    options.NumberHandling = JsonNumberHandling.AllowReadingFromString; // Newtonsoft coerces
+    options.ReadCommentHandling = JsonCommentHandling.Skip;     // Newtonsoft allows
```
GrabYourPitchforks: Similarly, this is a potentially dangerous setting and should not be encouraged unless the application has thought through the consequences. We chose not to allow this by default in S.T.J because it could lead to desynced deserialization attacks.

In these attacks, the frontend and the backend disagree on the contents of the payload. This could be because they disagree on where comments start or end, potentially allowing sensitive values to be smuggled through what appears to be an ignorable comment. (For example, Newtonsoft.Json, JSON5, and S.T.J all have different concepts on what "end of line" means for a single-line comment.)

plugins/dotnet/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
GrabYourPitchforks: To defend against this attack, the application should consider what happens if the adversary controls the _edges_ in the object graph, not just the _values_ of the nodes in the graph. For example, if you're deserializing a graph that represents a family tree, what happens if a node sets itself as its own parent? What happens if a node has two children, but both point _to the same object in memory_? (This last example isn't recursive; it's an example of a graph that has a many:1 relationship between edges and nodes.)

Both examples could violate business logic that the application otherwise tries to uphold. Allowing an untrusted client to control the edges in the graph represents a broader attack surface and involves more subtle reasoning compared to an untrusted client that can only control values within the graph.

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
GrabYourPitchforks: To defend against this attack, the application should:

1. ensure that all deserializers in the system are operating in strict RFC compliance mode with the most restrictive settings (e.g., disallowing duplicate property entries in the payload); or
2. ensure that all components involved in request handling use the same deserializer library, with the same major version and patch level, to ensure consistent payload processing; or
3. ensure the frontend deserializes the payload to an object graph, then reserializes that object graph to a new payload before sending it on to the backend, so the backend sees only what the frontend believed the request to contain.

ViktorHofer (dismissed):
@mrsharm please fix the path. These files should be under `plugins/...` not `src`

plugins/dotnet/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+**Key differences in converter API:**
+- Reader is `ref Utf8JsonReader` (struct, passed by ref) — NOT a class
+- Writer is `Utf8JsonWriter` — write methods are `WriteStringValue`, `WriteNumberValue`, `WriteBooleanValue` (typed)
+- No `serializer` parameter — use `options` and call `JsonSerializer.Serialize/Deserialize` for nested objects
+- For polymorphic deserialization: use `JsonTypeInfo` and `[JsonDerivedType]` (.NET 7+) instead of custom type handling
+
+### Step 5: Replace JToken/JObject/JArray with JsonDocument/JsonElement
```
JamesNK: I think just always recommend JsonNode and friends. It's closer than JsonDocument/JsonElement.

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,249 @@
...
+| Code using Newtonsoft.Json | Yes | Models, serialization calls, custom converters |
+| .NET version | No | Determines which System.Text.Json features are available |
+
+## Workflow
+
+### Step 1: Understand the critical behavioral differences
+
+**System.Text.Json is NOT a drop-in replacement.** These behaviors differ by default:
+
+| Behavior | Newtonsoft.Json | System.Text.Json | Impact |
```
JamesNK, line 35: STJ also escapes non-ASCII characters by default. Should recommend using relaxed encoder

eiriktsarpalis: I think we should caveat any such recommendation with the trade-offs: escaped JSON strings are semantically equivalent to unescaped ones, so this should depend on how the output is being used. Will it be embedded in HTML? Is it fed to LLMs that might get confused by the escaping? etc.

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
eiriktsarpalis: I would add that case insensitive property handling (recommended a few lines above this) similarly makes systems susceptible to interoperability attacks.

While I get that such settings might be necessary to preserve a modicum of compatibility with Json.NET, at a bare minimum we should try to document security concerns in the skill. It should start off assuming the highest bar possible (using `JsonSerializerDefaults.Strict`) and then instruct the agent to interview the user about particular requirements that force the loosening of individual settings, e.g. "Should the serializer support comments? Should the serializer allow case insensitive property binding? Should it escape non-ASCII characters? Should it allow trailing commas? Should it allow duplicate JSON properties? etc.)

plugins/dotnet/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,249 @@
...
+| `[JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]` | `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]` |
+| `[JsonConverter(typeof(MyConverter))]` | `[JsonConverter(typeof(MyConverter))]` (different base class!) |
+| `[JsonConstructor]` | `[JsonConstructor]` (same name, different namespace) |
+| `[JsonExtensionData]` | `[JsonExtensionData]` + must be `Dictionary<string, JsonElement>` (NOT `JToken`) |
+
+**Regex for finding Newtonsoft attributes:**
+```bash
+# Find all files using Newtonsoft attributes
+grep -rn "using Newtonsoft.Json" --include="*.cs"
+grep -rn "\[JsonProperty\|JsonConverter\|JsonIgnore\|JsonConstructor" --include="*.cs"
```
eiriktsarpalis: I'm pretty sure newer models can figure out regexes like that on their, so I would just skip that part for fear of it going out of sync.

eiriktsarpalis (changes requested):
It is an understatement to say that migrating serializers is a nontrivial task -- it is almost guaranteed that the new serializer is going to behave differently to an extent, so the goal of such a skill is to make sure all steps are taken so that risk is minimized.

An important aspect of minimizing risk in migrations of such scope is having a testing strategy. At a bare minimum, the agent should develop a test suite as it performs the migration comparing STJ serialization/deserialization behavior against the NJ baseline. This would help provide an empirically grounded feedback loop for the agent to apply necessary tweaks to DTOs, their attribute annotations, and their custom convert.

PR-FEEDBACK.md
ViktorHofer, line 1: Remove this file

.github/CODEOWNERS
```diff
@@ -49,4 +49,7 @@
 /plugins/dotnet/skills/dotnet-aot-compat/ @agocke @dotnet/appmodel
 /tests/dotnet/dotnet-aot-compat/ @agocke @dotnet/appmodel
 
+/plugins/dotnet/skills/migrating-newtonsoft-to-system-text-json/ @mrsharm
```
mrsharm: Update these to the appropriate team.

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
JamesNK: I'll leave it up to you. But if it's text being embedded in an HTML page then asp.net core will escape HTML characters anyway.

I've never understood why STJ does this.

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+  custom converters, and common breaking changes. Use when converting a project from
+  Newtonsoft.Json (Json.NET) to the built-in System.Text.Json serializer.
+---
+
+# Migrating from Newtonsoft.Json to System.Text.Json
+
+> **Important:** Migrating serializers is a nontrivial task. System.Text.Json will almost
+> certainly behave differently from Newtonsoft.Json in subtle ways. Always validate
+> serialization output and deserialization behavior thoroughly with real-world data after
+> migrating. Automated and manual testing of all serialization paths is essential.
```
eiriktsarpalis, line 14: Consider adding an explicit instruction to the agent that it should be adding baseline serialization tests for the application models while performing the migration.

plugins/dotnet-upgrade/skills/migrating-newtonsoft-to-system-text-json/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+## Common Pitfalls
+
+| Pitfall | Solution |
+|---------|----------|
+| Forgetting `PropertyNameCaseInsensitive = true` | Deserialization silently returns default values for all properties |
+| `JsonDocument` not disposed | Memory leak — always `using var doc = JsonDocument.Parse(...)` |
+| Using `JsonElement` after `JsonDocument` is disposed | JsonElement is invalid after dispose; clone with `element.Clone()` if needed |
+| `[JsonIgnore]` from wrong namespace | Both Newtonsoft and System.Text.Json have `[JsonIgnore]` — wrong `using` = attribute ignored |
+| Custom converter reading past the current token | System.Text.Json reader is strict — must read exactly the right tokens |
+| `JsonExtensionData` type mismatch | Use `Dictionary<string, JsonElement>`, `IDictionary<string, object>`, or `JsonObject` — not `JToken` |
```
eiriktsarpalis: Consider adding links to the official STJ docs at learn.

tests/dotnet-upgrade/migrating-newtonsoft-to-system-text-json/eval.yaml
```diff
@@ -0,0 +1,48 @@
+scenarios:
+  - name: "Migrate model with Newtonsoft.Json attributes to System.Text.Json"
```
eiriktsarpalis, line 2: Shouldn't we add evals covering more scenaria? Custom converters? Naming policies? etc? Consider mining real-world snippets either from dotnet org repos or public github repos in general (e.g. via grep.app)

eiriktsarpalis (approved):

AbhitejJohn:
@mrsharm : Presuming we are still interested in taking this, it might need a rebase on the latest main. Also tagging @dotnet/skills-upgrade-reviewers to help with reviews.

PR #236, opened 2026-03-05
Add dotnet-fsi-interactive (REPL) skill 

T-Gro:
Adds a skill teaching agents to use dotnet fsi as an interactive REPL for running .NET code without creating files. 
Includes eval scenario testing System.IO.Path edge cases.

The main skill advantage (compared to just vanilla running python, creating a .NET program or running dotnet fsi script.fsx) is when iteratively analyzing state (API calls, databases, .NET compute) that takes a lot of time to setup.


tests/dotnet/dotnet-fsi-interactive/eval.yaml
```diff
@@ -0,0 +1,27 @@
...
+      (4) What does Path.GetFileNameWithoutExtension(".gitignore") return?
+      Don't guess — these have surprising answers. Run .NET code to check.
+    assertions:
+      - type: "exit_success"
+      - type: "output_contains"
+        value: ".gitignore"
+      - type: "output_matches"
+        pattern: "(Combine|Join)"
+      - type: "output_matches"
+        pattern: "(/etc/passwd|replace)"
```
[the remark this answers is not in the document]
T-Gro: Thanks for the proposal, but way too strict - would lead to false negatives.

tests/dotnet/dotnet-fsi-interactive/eval.yaml
```diff
@@ -0,0 +1,27 @@
...
+    expect_tools: ["bash"]
+    max_turns: 10
+    rubric:
+      - "The agent correctly reports that Path.GetExtension('.gitignore') returns '.gitignore' (the whole filename is treated as extension)"
+      - "The agent correctly reports that Path.Combine('/usr', '/etc/passwd') returns '/etc/passwd' — the second absolute path replaces the first"
+      - "The agent correctly reports that Path.Join('/usr', '/etc/passwd') returns '/usr/etc/passwd' — it concatenates without replacing"
+      - "The agent verifies the answers by executing .NET code at runtime"
```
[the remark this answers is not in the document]
T-Gro: I have that in a separate dimension below via "The agent verifies the answers by executing .NET code at runtime"".
Repeating the same at all rubrics would mix up the "how" with the "what" evaluation.

plugins/dotnet/skills/dotnet-fsi-interactive/SKILL.md
```diff
@@ -0,0 +1,98 @@
+---
+name: dotnet-fsi-interactive
+description: "Run and test .NET code interactively without creating files. Explore .NET APIs, verify runtime behavior, experiment with code in a persistent REPL (Read-Eval-Print Loop) session via dotnet fsi. Also for data analysis and inspecting stateful objects."
```
danmoseley, line 3: F# code specifically, right? it should say that to avoid loading in C# context

T-Gro: Any .NET code - it is ephemeral from users perspective ("without file"), for verification and analysis that benefits from keeping state around (like having to load a lot of DB or web API data to do an iterative analysis)

danmoseley: I guess I meant -- someone could have installed this plugin -- want to write some temporary code, but prefer it being in C# because they aren't familar with F#. I have nothign against F#. If I misunderstood, disregard?

T-Gro: This is not creating a file users will see in their diff, it's similar to how agents generate and execute powershell or python for temporal tasks all the time, and this is transparent to the user (because no file is ever created,  its just for analysis/research/finding stuff out).

dotnet fsi is that, but for .NET apis and stateful exploration.

PR #155, opened 2026-03-02
Add minimal-api-file-upload skill

mrsharm:
## New Skill: minimal-api-file-upload

Teaches proper file upload handling in ASP.NET Core 8 minimal APIs, covering multiple counterintuitive configuration requirements that the model consistently gets wrong.

### Key Gotchas Covered
- Dual size limits: Kestrel MaxRequestBodySize AND FormOptions.MultipartBodyLengthLimit must both be configured (configuring only one is INCORRECT)
- DisableAntiforgery() required on file upload endpoints (auto-validation rejects form uploads)
- Safe filename generation (Path.GetFileName + Guid, not user-provided IFormFile.FileName)
- Content-type validation via magic bytes (ContentType alone is client-spoofable)
- IFormFile binding with [FromForm] for mixed parameters
- RequestSizeLimit attribute for per-endpoint size overrides

### Eval Results (3-run validation)
- **Overall: +38.9%** improvement (BL=3.0, SK=5.0)
- Confidence interval: [+9.1%, +62.5%] significant
- Effect size: g=+100.0%
- Quality improvement: 3.0/5 -> 5.0/5 (+2.0)
- Model: claude-opus-4.6

PR #264, opened 2026-03-06
Add minimal-api-file-upload skill

mrsharm:
Adds the **minimal-api-file-upload** skill for handling file uploads in ASP.NET Core 8+ minimal APIs.

> **Note:** Replaces #155 (migrated from skills-old repo to new plugins/ structure).

## What the Skill Teaches

The base model consistently gets several file upload patterns wrong. This skill addresses five key gaps:

1. **Dual size limits** — Must configure BOTH `Kestrel.MaxRequestBodySize` AND `FormOptions.MultipartBodyLengthLimit`. The base model typically configures only one, causing cryptic upload failures.
2. **Antiforgery auto-validation** — In .NET 8+, `UseAntiforgery()` silently rejects all form-bound uploads with 400 Bad Request. Must call `.DisableAntiforgery()` on API endpoints (with appropriate security caveats for cookie-auth).
3. **Magic byte validation** — The base model relies solely on `ContentType` which is client-spoofable. The skill teaches JPEG/PNG file signature verification.
4. **Safe filename generation** — The base model often uses `file.FileName` directly, enabling path traversal. The skill teaches GUID-based names and also covers sanitizing user-provided filenames when they must be retained (per reviewer feedback from @mikekistler).
5. **Streaming with MultipartReader** — For very large files (>500MB), IFormFile buffering causes OOM. The skill teaches direct-to-disk streaming via `MultipartReader`.

## Eval Results (3-run local validation)

| Scenario | Baseline | With Skill | Δ | Overfit | Verdict |
|----------|----------|------------|---|---------|---------|
| Implement secure file upload | 3.7/5 | **5.0/5** | +35% | 0.08 ✅ | ✅ |
| Upload multiple files with metadata | 3.3/5 | **5.0/5** | +52% | 0.08 ✅ | ✅ |
| Stream very large file uploads | 4.7/5 | **5.0/5** | +6% | 0.08 ✅ | ✅ |


plugins/aspnetcore/skills/minimal-api-file-upload/SKILL.md
```diff
@@ -0,0 +1,222 @@
...
+
+    // JPEG: FF D8 FF
+    // PNG: 89 50 4E 47
+    var isJpeg = header[0] == 0xFF && header[1] == 0xD8 && header[2] == 0xFF;
+    var isPng = header[0] == 0x89 && header[1] == 0x50 && header[2] == 0x4E && header[3] == 0x47;
+
+    if (!isJpeg && !isPng)
+        return Results.BadRequest("File content doesn't match declared type");
+
+    // CRITICAL: Generate a safe filename — never use user-provided filename directly
```
mikekistler: I think another valid approach is to validate the filename provided before using it. Often the filename provided by the user has some significance and must be retained.

mrsharm: Thanks, @mikekistler - just addressed this. Will await the evals - please let me know if any other changes are needed.

plugins/dotnet-aspnet/skills/minimal-api-file-upload/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+    // contain path traversal characters (e.g., "../../../etc/passwd").
+    //
+    // Option A (recommended): Generate a completely safe filename from the validated content
+    var extension = isJpeg ? ".jpg" : ".png";
+    var safeFileName = $"{Guid.NewGuid()}{extension}";
+    //
+    // Option B: If retaining the original filename matters (e.g., user-facing downloads),
+    // sanitize it first: strip path components, remove dangerous characters, and validate.
+    // var sanitized = Path.GetFileName(file.FileName);               // strips directory parts
+    // sanitized = Regex.Replace(sanitized, @"[^\w.\-]", "_");       // allow only safe chars
```
BrennanConroy: Not sure we want to be in the business of providing explicit sanitization code. If we did, we'd need to make sure it's actually correct, for example, you might need to check for reserved file names https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words and potentially avoid `.` at the beginning, or ` ` at the end of the name.

plugins/dotnet-aspnet/skills/minimal-api-file-upload/SKILL.md
```diff
@@ -0,0 +1,231 @@
...
+        if (!ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out var contentDisposition))
+            continue;
+
+        if (contentDisposition.DispositionType.Equals("form-data")
+            && !string.IsNullOrEmpty(contentDisposition.FileName.Value))
+        {
+            // Sanitize the user-provided filename to prevent path traversal
+            var originalFileName = contentDisposition.FileName.Value ?? string.Empty;
+            var sanitizedFileName = Path.GetFileName(originalFileName.Trim('"'));
+            var safeFile = $"{Guid.NewGuid()}{Path.GetExtension(sanitizedFileName)}";
```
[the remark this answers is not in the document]
BrennanConroy: Fix this

plugins/dotnet-aspnet/skills/minimal-api-file-upload/SKILL.md
```diff
@@ -0,0 +1,234 @@
...
+    return Results.Ok(new { FileName = safeFileName, file.Length });
+});
+```
+
+### Step 5: CRITICAL — Streaming Large Files Without Buffering
+
+```csharp
+// CRITICAL: IFormFile uses multipart form parsing that buffers with a memory threshold
+// and spills to temp files. For very large uploads this can consume significant
+// memory and disk. Use MultipartReader for streaming without buffering.
```
BrennanConroy: nit: This talks about consuming significant memory and disk but then writes directly to disk which is still using the disk the exact same amount. Streaming allows you to process the data in chunks which means you could avoid writing to disk if you don't need to buffer the entire file or store it as you process it.

PR #265, opened 2026-03-06
Add implementing-server-sent-events skill

mrsharm:
## Summary
Adds a skill for implementing Server-Sent Events endpoints in ASP.NET Core 8 minimal APIs.

> **Note:** Replaces #147 (migrated from skills-old repo to new plugins/ structure).

### Eval Results (3-run)
- **Overall: +13.6%** improvement (BL=4.0, SK=4.7)

## Files
- plugins/dotnet/skills/implementing-server-sent-events/SKILL.md
- tests/dotnet/implementing-server-sent-events/eval.yaml

### DIFF
--- a/PR-264-description.md
+++ b/PR-264-description.md
@@ -1,0 +1,21 @@
+## Summary
+Adds the **minimal-api-file-upload** skill for handling file uploads in ASP.NET Core 8 minimal APIs.
+
+> **Note:** Replaces #134 (migrated from skills-old repo). Skill moved to `aspnetcore` plugin per repo restructuring.
+
+## What the Skill Teaches
+- `IFormFile` binding in minimal APIs (when `[FromForm]` is needed vs automatic)
+- `IFormFileCollection` for multiple file uploads
+- Security: file size limits, content type allowlists, magic byte validation
+- Safe file naming with `Guid` + validated extension (never trust client filename)
+- Streaming uploads with `MultipartReader` for large files
+- Antiforgery considerations
+- Path traversal prevention
+
+## Eval Scenarios
+- (eval.yaml needs scenarios — currently empty)
+
+## Files
+- `plugins/aspnetcore/plugin.json` — ASP.NET Core plugin
+- `plugins/aspnetcore/skills/minimal-api-file-upload/SKILL.md` — skill instructions
+- `tests/aspnetcore/minimal-api-file-upload/eval.yaml` — eval scenarios (to be added)
--- a/PR-265-description.md
+++ b/PR-265-description.md
@@ -1,0 +1,23 @@
+## Summary
+Adds the **implementing-server-sent-events** skill for building SSE endpoints in ASP.NET Core 8 minimal APIs.
+
+> **Note:** Replaces #139 (migrated from skills-old repo). Skill moved to `aspnetcore` plugin per repo restructuring.
+
+## What the Skill Teaches
+- Manual SSE protocol implementation (no built-in `MapSSE` — doesn't exist)
+- Required headers: `Content-Type: text/event-stream`, `Cache-Control: no-cache`, `Connection: keep-alive`
+- Correct SSE event framing with `\n\n` terminators
+- `Response.Body.FlushAsync()` for immediate delivery
+- Event IDs + `Last-Event-ID` header for reconnection support
+- `retry:` field for controlling client reconnection interval
+- `X-Accel-Buffering: no` for reverse proxy compatibility
+- `HttpContext.RequestAborted` / `CancellationToken` for client disconnect detection
+- `Channel<T>` for background service → endpoint communication
+
+## Eval Scenarios
+1. Implement SSE notification endpoint in ASP.NET Core 8 minimal API
+
+## Files
+- `plugins/aspnetcore/plugin.json` — ASP.NET Core plugin
+- `plugins/aspnetcore/skills/implementing-server-sent-events/SKILL.md` — skill instructions
+- `tests/aspnetcore/implementing-server-sent-events/eval.yaml` — 1 eval scenario (needs expansion)
--- a/PR-266-description.md
+++ b/PR-266-description.md
@@ -1,0 +1,25 @@
+## Summary
+Adds the **implementing-websocket-endpoints** skill for building WebSocket endpoints in ASP.NET Core 8.
+
+> **Note:** Replaces #142 (migrated from skills-old repo). Skill moved to `aspnetcore` plugin per repo restructuring.
+
+## What the Skill Teaches
+- `UseWebSockets()` middleware setup (no `MapWebSocket` — doesn't exist)
+- Manual WebSocket upgrade via `AcceptWebSocketAsync`
+- Proper receive loop with `EndOfMessage` fragment handling
+- **Critical:** Use `CloseOutputAsync` (not `CloseAsync`) when responding to client-initiated close to avoid deadlock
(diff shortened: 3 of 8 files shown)


mrsharm:
﻿## Feedback carried over from #147

### Code Review Comments

**Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 110): Same compile issue here: `context.Response.Headers.ContentType` / `.CacheControl` are not valid APIs on `IHeaderDictionary`. Update this snippet to use `context.Response.ContentType` / `Headers["Cache-Control"]` (or typed headers) so the example builds. ```suggestion     context.Response.ContentType = "text/event-stream";     context.Response.Headers["Cache-Control"] = "no-cache"; ```  ---  **Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 176): Same header API issue in this snippet: `context.Response.Headers.ContentType` / `.CacheControl` wonΓÇÖt compile. Please update consistently across all examples (Step 5/6) to avoid copy/paste failures.  ---  **Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 97): `StreamWriter.WriteLineAsync` uses the platform newline by default (often `\r\n`), and this example also appends an extra `\n`, which can produce confusing line endings for SSE. Prefer writing explicit `\n` (or set `writer.NewLine = "\n"` and use `WriteAsync`) to keep the SSE framing unambiguous cross-platform. ```suggestion await writer.WriteAsync($"data: {msg}\n\n");  // Explicit \n\n for SSE event framing ```  ---  **Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 147): `int.Parse(lastEventId)` will throw if the client sends a non-integer `Last-Event-ID` header (or an out-of-range value), turning a reconnect into a 500. Use `int.TryParse` and fall back to a safe default (or return 400) to make the reconnection example robust. ```suggestion     var startFrom = 0;     if (!string.IsNullOrEmpty(lastEventId) && int.TryParse(lastEventId, out var parsedId) && parsedId >= 0)     {         startFrom = parsedId + 1;     } ```  ---  **Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 18): This skill doc doesnΓÇÖt include an explicit ΓÇ£ValidationΓÇ¥ section/checklist. CONTRIBUTING.md asks that changes include validation steps a reviewer can follow; consider adding a short section (e.g., how to run the sample, curl an SSE stream, verify `Last-Event-ID` behavior, and test via nginx) so the skill is verifiable.  ---  **Copilot** on src/dotnet/tests/implementing-server-sent-events/eval.yaml (line 2): This eval file is under `src/dotnet/tests/...`, but repo convention (and CI) expects evals at `tests/<plugin>/<skill-name>/eval.yaml` (e.g. `tests/dotnet/implementing-server-sent-events/eval.yaml`). With the current location, the evaluation workflow that runs `--tests-dir ./tests/dotnet` will not pick up these scenarios.  ---  **Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 4): This SKILL.md is placed under `src/dotnet/skills/...`, but the repository layout expects skills under `plugins/<plugin>/skills/<skill-name>/SKILL.md` so they are packaged and discovered (e.g. `plugins/dotnet/skills/implementing-server-sent-events/SKILL.md`). As-is, the evaluation workflow that validates `./plugins/dotnet/skills` will not discover or test this skill.  ---  **Copilot** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 44): These snippets use `context.Response.Headers.ContentType` / `.CacheControl`, but `HttpResponse.Headers` is an `IHeaderDictionary` and doesn't expose those properties. This will not compile in ASP.NET Core; set `context.Response.ContentType` (or `Headers["Content-Type"]`) and `Headers["Cache-Control"]` (or use `context.Response.GetTypedHeaders()`). ```suggestion     context.Response.ContentType = "text/event-stream";     context.Response.Headers["Cache-Control"] = "no-cache"; ```  ---  **danmoseley** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 8): move when to use and when not to use entirely or largely into the description, which is the thing the ai reads when deciding whether or not to load. probably shouldn't have this section here.  ---  **danmoseley** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 65): should `{message}` be sanitized for newlines here and below. AI says - risk of injection attack  ---  **danmoseley** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 8): SSE endpoints hold long-lived HTTP connections. Without connection limits, an attacker (or misbehaving client) can exhaust server resources by opening many concurrent connections. Consider adding a note about: - Enforcing per-client or total connection limits (e.g., via middleware or a `SemaphoreSlim`) - Requiring authentication before establishing the SSE connection - Setting an idle timeout to drop stale connections  This is especially relevant for a skill doc since agents will copy the pattern as-is without thinking about production hardening.  ---  **danmoseley** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 207): Worth adding a note about CORS configuration. SSE endpoints are frequently consumed cross-origin via the browser `EventSource` API, which does not support custom headers. Without `app.UseCors(...)` or a CORS policy on the endpoint, cross-origin clients will silently fail. This is a common deployment surprise.  ---  **danmoseley** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 8): (says AI)  ---  **danmoseley** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 207): (says AI)  ---  **BrennanConroy** on src/dotnet/skills/implementing-server-sent-events/SKILL.md (line 31): Wrong. We have `TypedResults.ServerSentEvents` https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#support-for-server-sent-events-sse  Most of this Skill should probably be rewritten to use `TypedResults.ServerSentEvents`  --- 
### Discussion Comments

**mrsharm**: ## Eval Results: implementing-server-sent-events  ### 3-Run Validation: +13.6% PASS  | Metric | Baseline | With Skill | Change | |--------|----------|------------|--------| | Overall Score | 4.0/5 | 4.7/5 | +0.7 | | Quality (overall) | 4.0/5 | 4.7/5 | +40% | | Pairwise | ΓÇö | skill wins | consistent |  The skill consistently improves SSE implementations by teaching the manual protocol (no built-in helper exists), proper flush behavior, disconnection handling, and reconnection support.  **Model**: claude-opus-4.6 (baseline + skill + judge)  ---  **danmoseley**: needs CODEOWNERS entry  ---

ManishJayaswal:
@mrsharm - - the repo has undergone some restructuring to make everything more organized. Hence, we are asking all open PRs to update the branch. Sorry about this.
This skill should be under ASP plugin. Please update the PR and submit again.
@adityamandaleeka @BrennanConroy - please review

BrennanConroy (commented):
Like I said on the original PR, there is built-in SSE support in 10.0. 
https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#support-for-server-sent-events-sse

And if you don't want to use that, you can use the [SsEFormatter](https://source.dot.net/#System.Net.ServerSentEvents/System/Net/ServerSentEvents/SseFormatter.cs,5035e42873ad1073,references) type to write SSE frames from System.Net.ServerSentEvents nupkg.

PR #131, opened 2026-02-26
Add implementing-rate-limiting skill

mrsharm:
## Skill: implementing-rate-limiting

Teaches the model to implement .NET 8 built-in rate limiting correctly, covering:
- Return 429 (not default 503) via RateLimiterOptions.RejectionStatusCode = 429`n- Middleware ordering (UseRateLimiter() after UseRouting(), before MapControllers())
- Per-client partitioned rate limiting with CreatePartitionedLimiter<HttpContext>()"
- Named policies applied per-endpoint
- OnRejected callback with Retry-After header
- Algorithm selection (fixed window, sliding window, token bucket, concurrency)

## Eval Results (3-run)
**Overall: +39.4% improvement - PASSED**

| Scenario | BL | SK |
|----------|----|----|
| Add per-client rate limiting | 3 | 4 |

PR #267, opened 2026-03-06
Add implementing-rate-limiting skill

mrsharm:
## Summary
Adds a skill for implementing .NET 8 built-in rate limiting correctly.

> **Note:** Replaces #131 (migrated from skills-old repo to new plugins/ structure).

### Eval Results (3-run)
- **Overall: +39.4%** improvement

## Files
- plugins/dotnet/skills/implementing-rate-limiting/SKILL.md
- tests/dotnet/implementing-rate-limiting/eval.yaml

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -85,6 +85,10 @@
 /plugins/dotnet-template-engine/ @YuliiaKovalova @JanKrivanek
 /tests/dotnet-template-engine/ @YuliiaKovalova @JanKrivanek
 
+# aspnetcore (ASP.NET Core web development)
+/plugins/aspnetcore/ @BrennanConroy @adityamandaleeka
+/tests/aspnetcore/ @BrennanConroy @adityamandaleeka
+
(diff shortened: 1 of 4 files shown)


.github/CODEOWNERS
```diff
@@ -49,4 +49,7 @@
 /plugins/dotnet/skills/dotnet-aot-compat/ @agocke @dotnet/appmodel
 /tests/dotnet/dotnet-aot-compat/ @agocke @dotnet/appmodel
 
+/plugins/dotnet/skills/implementing-rate-limiting/ @mrsharm
```
mrsharm: Figure out who is the right owner.

mrsharm:
Hi @SamMonoRT - this was an eval specific decision to presumably test with the latest and greatest model (at the time) to ensure that the skill is adding some value against the base model. I think it's also worthwhile testing this with other models (particularly older ones) but the assumption is that they are not new enough to not have the latest .NET behavior in their training set.

PR #75, opened 2026-02-21
Add Skill: Convert to Central Package Management (convert-to-cpm)

jeffhandley:
Replaces #28 with a rebase on main, updating to the new repo layout, and adoption of the eval.yml test approach

Introduce the `convert-to-cpm` skill, which guides the migration of .NET projects to NuGet Central Package Management (CPM). It provides detailed workflow steps, auditing guidance, validation procedures, error handling, and test scenarios for various conversion complexities.

PR #269, opened 2026-03-06
Add refactoring-to-async skill

mrsharm:
Adds the refactoring-to-async skill for bottom-up async conversion patterns in .NET.

Note: Replaces #79 (migrated from skills-old repo to new plugins/ structure).

## What the Skill Teaches

- Bottom-up async conversion order — start from leaf I/O methods, propagate upward
- Correct sync-to-async API mappings (stream.Read → ReadAsync, connection.Open → OpenAsync, etc.)
- CancellationToken propagation through the entire call chain
- Identifying sync-over-async anti-patterns (.Result, .Wait(), .GetAwaiter().GetResult())
- ConfigureAwait(false) usage in library code vs ASP.NET Core apps
- Recognizing when async does not apply (CPU-bound work)Adds the refactoring-to-async skill for bottom-up async conversion patterns in .NET.


mrsharm:
﻿## Feedback carried over from #79 (refactoring-to-async)

### Code Review Comments (10 threads)

**@Copilot** — `eval.yaml` L12
> The scenario requires identifying sync-over-async patterns like `.Result`/`.Wait()` (rubric line 14), but the `output_not_matches` assertion fails the run if the agent output contains `.Result` or `.Wait()` anywhere (including when calling them out as anti-patterns). This creates false negatives and makes the rubric effectively incompatible with the assertions. Consider tightening the regex to ...

---

**@Copilot** — `SKILL.md` L35
> The `grep` pattern uses `\b` for a word-boundary, but `grep`'s default regex syntax does not treat `\b` as a word boundary (itΓÇÖs typically interpreted as a backspace). Also, `\.Read()` only matches parameterless `Read()` calls and will miss the common `Read(...)` overloads. Consider switching to `rg` (ripgrep) or `grep -P`/`grep -E` with a corrected pattern (e.g., matching `\.Read\(` and `\.W...

---

**@Copilot** — `SKILL.md` L155
> Same issue as Step 1: `grep` without `-P` won't treat `\b` as a word boundary, so this command may not reliably find `.Result` usages. Adjust the command to use a regex flavor that supports word boundaries (or use `\<`/`\>`), so the "should return zero results" guidance is accurate.
> ```suggestion
> grep -rnP '\.Result\b|\.Wait\(\)|\.GetAwaiter\(\)\.GetResult\(\)' --include="*.cs" .
> ```

---

**@danmoseley** — `SKILL.md` L8
> move any part of use/not use into decription similar to
> https://github.com/dotnet/runtime/pull/125005/changes#diff-ded9450821c1df27638def6250c00784c7f795e3a9c56ad13d09a34853a0d09bR3-R4
> if it can avoid unnecessaryt reloading.
> 
> Possibly not all can go there. For example: user wants to parallelize work is something that can be known before loading the skill, and avoid load. Whereas possibly "code ...

---

**@danmoseley** — `SKILL.md` L167
> this is missing from all the examples. maybe indicate those are app code? 
> 
> It might be worth mentioning as instructions, not just in anti-patterns -- if code is library add `.ConfigureAwait(false)` to every await. It should be used consistently or not at all.

---

**@danmoseley** — `SKILL.md` L185
> this may be a big high level/conceptual for this skill?

---

**@danmoseley** — `SKILL.md` L53
> ```suggestion
> | `reader.Read()` (DbDataReader) | `await reader.ReadAsync()` |
> | `command.ExecuteNonQuery()` (DbCommand) | `await command.ExecuteNonQueryAsync()` |
> 
> ```
> maybe

---

**@danmoseley** — `eval.yaml` L25
> is this really something the user would write? seems like a gimme as written. more likely the user would say "make DoIt() async so it's faster" and the AI would have to figure out that  it's CPU  bound?

---

**@danmoseley** — `eval.yaml` L33
> do you expect `ConfigureAwait(false)` in this UserService case? either way, should be a test for the opposite case (eg winforms) and verify in each case it's present or not as expected

---

**@danmoseley** — `SKILL.md` L148
> ```suggestion
> | Error | Fix |
> |---|---|
>  Error | Fix |
>    |---|---|
>    | `CS4032`: `await` in non-async method | Add `async` to the method signature and return `Task` or `Task<T>` |
>    | `CS0029`: Cannot convert `Task<T>` to `T` | Add `await` before the call |
>    | `CS0127`: Method returns `Task` but body returns value | Change return type to `Task<T>` |
>    | `CS1983`: Return type of async meth...

---

### Discussion Comments (5)

**@ViktorHofer**:
> Please share the results as a comment similar to https://github.com/dotnet/skills/pull/75#issuecomment-3938172836
> Also make sure that you test at least 3 runs (`--runs 3` for local validation): https://github.com/dotnet/skills/pull/82/changes

---

**@mrsharm**:
> ## Skill Validation Results ΓÇö refactoring-to-async
> 
> | Skill | Test | Baseline | With Skill | ╬ö | Verdict |
> |-------|------|----------|------------|---|---------|
> | refactoring-to-async | Refactor synchronous service to async | 3.3/5 | 5.0/5 | +1.7 | Γ£à |
> | refactoring-to-async | Async refactoring should not apply to CPU-bound code | 1.0/5 | 1.0/5 | 0.0 | Γ£à |
> 
> **Overall improvement: +18.8%...

---

**@mrsharm**:
> ## Skill Validation Results ΓÇö refactoring-to-async
> 
> | Skill | Test | Baseline | With Skill | ╬ö | Verdict |
> |-------|------|----------|------------|---|---------|
> | refactoring-to-async | Refactor synchronous service to async | 3.3/5 | 5.0/5 | +1.7 | Γ£à |
> | refactoring-to-async | Async refactoring should not apply to CPU-bound code | 1.0/5 | 1.0/5 | 0.0 | Γ£à |
> 
> **Overall improvement: +18.8%...

---

**@mrsharm**:
> > Please share the results as a comment similar to [#75 (comment)](https://github.com/dotnet/skills/pull/75#issuecomment-3938172836) Also make sure that you test at least 3 runs (`--runs 3` for local validation): https://github.com/dotnet/skills/pull/82/changes
> 
> Done. 

---

**@danmoseley**:
> add codeowners entry, move files around to match new pattern in main.

---

.github/CODEOWNERS
```diff
@@ -49,4 +49,7 @@
 /plugins/dotnet/skills/dotnet-aot-compat/ @agocke @dotnet/appmodel
 /tests/dotnet/dotnet-aot-compat/ @agocke @dotnet/appmodel
 
+/plugins/dotnet/skills/refactoring-to-async/ @mrsharm
```
mrsharm: Update this to the right owner.

ViktorHofer:
@mrsharm check the one scenario in which the skill doesn't get activated.

mrsharm:
> @mrsharm check the one scenario in which the skill doesn't get activated.

@ViktorHofer: Intentional - is a negative test and is the correct outcome. The eval suggests optimizing matrix multiplication (CPU-bound), not about converting I/O to async. The skill's description targets I/O-bound async refactoring, so the agent should recognize this isn't an async problem and not load the skill. That's exactly what happened in all 3 runs.

.github/CODEOWNERS
mrsharm: @danmoseley / @timheuer - who should be the right owner for this skill?

ViktorHofer:
@mrsharm such tests must be annotated with `expect_activation: false`: https://github.com/dotnet/skills/blob/8e50fc855891cfa7ffd1435730fdb0a0460d5e9b/tests/dotnet-diag/dump-collect/eval.yaml#L161

Otherwise we have no way to distinguish between intentionally not activated and unintentionally not activated.

mrsharm:
@ViktorHofer - thanks! Been implemented for this and the other PRs.

.github/CODEOWNERS
```diff
@@ -24,6 +24,9 @@
 /plugins/dotnet-upgrade/skills/thread-abort-migration/ @dotnet/appmodel
 /tests/dotnet-upgrade/thread-abort-migration/ @dotnet/appmodel
 
+/plugins/dotnet/skills/refactoring-to-async/ @mrsharm
+/tests/dotnet/refactoring-to-async/ @mrsharm
+
```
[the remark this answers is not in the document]
mrsharm: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/dotnet/skills/pull/269#discussion_r2913522610)

tests/dotnet/refactoring-to-async/eval.yaml
```diff
@@ -0,0 +1,259 @@
+scenarios:
+  - name: "Refactor synchronous service to async"
+    prompt: "I have a service class with several synchronous database and HTTP calls. It's causing thread pool starvation under load. Can you convert it to async/await?"
+    setup:
+      copy_test_files: true
+    assertions:
```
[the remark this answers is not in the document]
mrsharm: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/dotnet/skills/pull/269#discussion_r2913920332)

plugins/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,183 @@
+---
+name: refactoring-to-async
+description: >
+  Convert synchronous .NET code to async/await, including proper Task propagation,
+  cancellation support, and avoiding common async anti-patterns.
+  USE FOR: converting blocking I/O calls (database, HTTP, file, stream) to async,
+  fixing thread pool starvation from .Result/.Wait()/.GetAwaiter().GetResult(),
+  modernizing sync-over-async code, adding CancellationToken support.
+  DO NOT USE FOR: CPU-bound computation (use Parallel.For or Task.Run instead),
+  code with no I/O operations, parallelizing work rather than making it async.
```
jaredpar: The budget for skills descriptions is limited which is why they  are intended to be brief. This seems longer than I would expect for a skill.

tests/dotnet/refactoring-to-async/eval.yaml
```diff
@@ -0,0 +1,259 @@
...
+        pattern: "(\\basync\\s+Task\\b.*\\(|await.*ReadAsync|await.*WriteAsync)"
+    rubric:
+      - "Did NOT suggest converting CPU-bound computation to async/await"
+      - "Suggested parallelism approaches (Parallel.For, Task.Run, SIMD, or algorithmic optimization)"
+      - "Correctly identified that async is for I/O-bound work, not CPU-bound work"
+    timeout: 120
+
+  - name: "Refactor sync-over-async in ASP.NET Core controller"
+    prompt: |
+      This ASP.NET Core controller has performance issues under load. Some methods use .GetAwaiter().GetResult() and others mix sync and async patterns. Can you fix the async anti-patterns and convert it properly?
```
jaredpar: I worry that the prompt here is too strongly connected to the prompts in the skill. Basically the skill says "for ASP.NET do X" and the eval say "this is ASP.NET". It feels like the test is being fitted for the skill vs. the user experience. My intuition is that customers would more naturally write

> This controller has performance issues under load ... 

The model would then need to infer what type of code this was to decide the right action from the skill.

tests/dotnet/refactoring-to-async/UserService.cs
```diff
@@ -0,0 +1,113 @@
...
+    private readonly string _connectionString;
+
+    public UserRepository(string connectionString)
+    {
+        _connectionString = connectionString;
+    }
+
+    public User GetById(int id)
+    {
+        using var connection = new SqlConnection(_connectionString);
```
danmoseley, line 23: should the skill recommend using `IAsyncDisposable` (ie `await using`) ? and eval should check that all these `using var` get converted.

plugins/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+This should return zero results in the refactored code paths.
+
+## Anti-Patterns to Avoid
+
+| Anti-Pattern | Problem | Correct Approach |
+|---|---|---|
+| `task.Result` or `task.Wait()` | Blocks thread, risks deadlock | `await task` |
+| `async void` methods | Exceptions crash the process | `async Task` (except event handlers) |
+| `Task.Run` wrapping async I/O | Wastes a thread pool thread | Call async method directly |
+| Missing `ConfigureAwait(false)` in libraries | Can deadlock in UI/ASP.NET sync contexts | Add `.ConfigureAwait(false)` to every `await` in library code; omit in ASP.NET Core app code (no SynchronizationContext) |
```
danmoseley:
```suggestion
| Missing `ConfigureAwait(false)` in libraries | Can deadlock in Winforms/WPF/ASP.NET sync contexts | Add `.ConfigureAwait(false)` to every `await` in library code; omit in ASP.NET Core app code (no SynchronizationContext) |
```

tests/dotnet/refactoring-to-async/eval.yaml
```diff
@@ -0,0 +1,259 @@
...
+      - type: "output_matches"
+        pattern: "(async Task<ActionResult|async Task<IActionResult)"
+      - type: "output_matches"
+        pattern: "(CancellationToken|cancellation)"
+    rubric:
+      - "Identified all three sync-over-async anti-patterns: .GetAwaiter().GetResult() in GetOrder, .Wait() in CreateOrder, and .Result in the GetReport loop"
+      - "Converted controller action methods to async Task<ActionResult<T>>"
+      - "Replaced .GetAwaiter().GetResult(), .Wait(), and .Result with await"
+      - "Added CancellationToken parameter to async actions to support request cancellation"
+      - "Used Task.WhenAll or sequential awaits instead of blocking .Result in the loop"
```
danmoseley: AI says, "Task.WhenAll fires all HTTP requests concurrently, which could overwhelm a downstream service. Sequential await should be the preferred default; WhenAll should only be suggested with a concurrency caveat."

plugins/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Common Pitfalls
+
+| Pitfall | Solution |
+|---------|----------|
+| Deadlock after conversion | Ensure `await` is used everywhere; no `.Result` mixed with `await` |
+| Performance worse after conversion | Async adds overhead for CPU-bound work; only use for I/O |
+| Forgetting to update tests | Test methods must return `Task` and use `await` |
+| Breaking interface consumers | Consider keeping sync wrappers temporarily during staged migration |
+| `ValueTask` vs `Task` confusion | Use `Task` by default; `ValueTask` only for hot-path methods that frequently return synchronously |
```
danmoseley, line 234: There should be a bit more mention of `ValueTask` in the guidance higher up and its limitations over `Task`. Eg., cannot be awaited multiple times, cannot be stored.

plugins/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+### Step 7: Verify no anti-patterns remain
+
+Search for remaining issues:
+
+```bash
+grep -rnE '\.Result\b|\.Wait\(\)|\.GetAwaiter\(\)\.GetResult\(\)' --include='*.cs' .
+```
+
+This should return zero results in the refactored code paths.
+
```
danmoseley, line 207: Is `IAsyncEnumerable` / `await foreach` in scope? Probably should be. If not, it should say it's out of scope.  Similar for `IAsyncDisposable`

plugins/dotnet/skills/refactoring-to-async/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+| Scope | No | Single method, class, or full call chain |
+
+## Workflow
+
+### Step 1: Identify blocking I/O calls
+
+Search for synchronous I/O patterns in the codebase:
+
+```bash
+grep -rnE '\.Result\b|\.Wait\(\)|\.GetAwaiter\(\)\.GetResult\(\)|ReadToEnd\(\)|\.Read\(|\.Write\(' --include='*.cs' .
```
danmoseley: Suggest to provide a list of example method names, rather than write the grep for it. It's great at writing grep, and giving it the command suggests that this is a complete list. Plus if this hits something inappropriate eg a random property named "Result" it can adjust itself.

tests/dotnet/refactoring-to-async/eval.yaml
danmoseley: yes, I agree this is arguably "overfitting". Current overfitting judge doesn't catch it. @JanKrivanek I wonder whether this suggests we should tune prompt given to the overfitting judge.

plugins/dotnet/skills/refactoring-to-async/SKILL.md
danmoseley: We only validate the hard limit of 1024 chars. Beyond that, it's hard to score whether there are too many (waste) or appropriate (prevents loading the skill inappropriately). If anything evaluation may drive to more tokens (in order to force skill load when it should). I don't know whether human intuition can help much here or not.

ViktorHofer (approved):
Eval results look good, infra changes also look good. Please get an additional approval from an area expert.

AbhitejJohn:
@mrsharm This PR was approved back in March and evals pass — is this still something you'd like to land? It likely needs a rebase on main at this point. Let us know if you need help getting it across the finish line.

PR #279, opened 2026-03-07
Add agent and plugin spec conformance validation

danmoseley:
## Add agent and plugin spec conformance validation

Extends the skill-validator to validate `.agent.md` and `plugin.json` files against the public specs, complementing the SKILL.md validation added in #238.

### What's validated

**Plugins (`plugin.json`):**
- Required fields: `name`, `version`, `description`, `skills`
- Name must match directory name and follow kebab-case naming rules
- Description length limit (1024 chars)
- Skills/agents paths must exist on disk and stay within plugin root (no absolute paths or `..` traversal)

**Agents (`.agent.md`):**
- Required frontmatter with `name` and `description` fields
- Name must match `{name}.agent.md` filename convention
- Name format follows agentskills.io naming rules (lowercase, hyphens, max 64 chars)
- Description length limit (1024 chars)
- Body line limit (500 lines, matching SKILL.md convention)

**Orphaned test directories:**
- Every directory under `tests/{plugin}/` must have a corresponding `plugins/{plugin}/`
- Every directory under `tests/{plugin}/{skill}/` must have a corresponding `plugins/{plugin}/skills/{skill}/`
- Catches stale or misnamed test directories that would otherwise silently never run

### Spec references
- SKILL.md naming rules: https://agentskills.io/specification#name-field
- VS Code agent plugins: https://code.visualstudio.com/docs/copilot/customization/agent-plugins
- Claude Code plugin manifest: https://code.claude.com/docs/en/plugins-reference

### Implementation
- `AgentProfiler.cs` — validates `.agent.md` files (mirrors `SkillProfiler` patterns)
- `PluginValidator.cs` — validates `plugin.json` files with `TryGetSafeSubdirectory` for path traversal protection
- `SkillDiscovery.cs` — discovers agents and plugins from plugin directories; `FindOrphanedTestDirectories` cross-references `tests/` against `plugins/`
- `ValidateCommand.cs` — wires plugin/agent/orphan validation into the CLI pipeline (runs before evals, failures are blocking)
- Refactored `ValidateName` into `ValidateNameFormat` (format-only) + `ValidateName` (format + directory match) so agent/plugin callers get kind-specific messages ("Agent name", "Plugin name") instead of hardcoded "Skill name"
- Empty agent name and plugin name fall back to filename/directory name for readable error output

### Other changes
- Removed unused `"agents": "./agents/"` from three `plugin.json` files that declared an agents path without having an `agents/` directory

### Test coverage
35 new tests covering both profilers, including edge cases for name format rules (hyphens, length, case), description limits, body line limits, filename conventions, plugin.json parsing, path traversal rejection, name fallback behavior, message-prefix correctness, and orphaned test directory detection.

### Validated against repo
```
Found 25 skill(s)
Validated 5 plugin(s)
Validated 4 agent(s)
```
All existing plugins, agents, and test directories pass with no errors or warnings.

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -136,7 +136,12 @@ jobs:
     uses: ./.github/workflows/evaluation-run.yml
     with:
       entries: ${{ needs.discover.outputs.entries }}
-      runs: ${{ needs.discover.outputs.is_infra == 'true' && '2' || (github.ref == 'refs/heads/main' && '5' || '3') }}
+      runs: ${{ needs.discover.outputs.is_infra == 'true' && '1' || (github.ref == 'refs/heads/main' && '5' || '3') }}
+      # Infra changes evaluate all plugins in parallel — reduce per-job concurrency
+      # to avoid API rate limits and timeouts from contention.
+      parallel-skills: ${{ needs.discover.outputs.is_infra == 'true' && '2' || '5' }}
+      parallel-scenarios: ${{ needs.discover.outputs.is_infra == 'true' && '3' || '5' }}
+      parallel-runs: ${{ needs.discover.outputs.is_infra == 'true' && '3' || '5' }}
       pr-number: ${{ github.event.pull_request.number || '' }}
     secrets:
       COPILOT_GITHUB_TOKEN: ${{ secrets.COPILOT_GITHUB_TOKEN }}


danmoseley:
That seems to have fixed it. @ViktorHofer

PR #283, opened 2026-03-08
feat: skill loading tests

DeagleGross:
## Problem

It is not free to load skills. Each loaded skill is eating-up the context of LLM, and it is crucial to make skill very specific to solving some conrete problem. We should be able to validate whether skill will be loaded or not based on its frontmatter.

## Solution

I've added `--selectivity-test` option which is the new lightweight mode that probes skill activation without running the full evaluation. Each `eval.yml` should have `should_activate` and `should_not_activate` which should be close to the topic of the skill. Validation will run a lightweight agent run, and assert if skill was not loaded for all `should_activate` prompts, and should not be called for `should_not_activate`.

## Usage & Example
```
dotnet run --project eng/skill-validator/src -- \
  plugins/dotnet-msbuild/skills/build-perf-diagnostics \
  --tests-dir tests/dotnet-msbuild \
  --selectivity-test --runs 1 --verbose
```

> [build-perf-diagnostics] 🔍 Evaluating...
[build-perf-diagnostics] 📊 📊 build-perf-diagnostics: 1,560 BPE tokens [chars/4: 1,649] (detailed ✓), 12 sections, 10 code blocks
[build-perf-diagnostics] 🎯 Running selectivity test (standalone)...
[build-perf-diagnostics/selectivity] Testing should_activate: "My .NET build takes over 5 minutes, how can I speed it up?"
⠹ Evaluating 1 skill(s)...      📂 C:\Users\dmkorolev\AppData\Local\Temp\sv-6e86045eebd84007834facc17361ee54 (skilled)
[build-perf-diagnostics/selectivity] Testing should_activate: "How do I analyze a binlog to find slow targets in MSBuild?"
[build-perf-diagnostics/selectivity]   → ✅ activated: "Our CI builds are fast but local dev builds are p…"
...
[build-perf-diagnostics/selectivity]   → ✅ correctly NOT activated: "What's the difference between .NET 8 and .NET 9?"
[build-perf-diagnostics/selectivity]   → ✅ correctly NOT activated: "My unit tests are failing with a NullReferenceExc…"
[build-perf-diagnostics] 🎯 Selectivity: recall=100%, precision=100% — PASSED

> ═══ Skill Validation Results ═══

>  🎯 Selectivity: recall=100%, precision=100% ✅
    ✓ "My .NET build takes over 5 minutes, how can I speed it up?" — should activate → activated
    ✓ "How do I analyze a binlog to find slow targets in MSBuild?" — should activate → activated
    ✓ "Roslyn analyzers are making my compilation really slow, wha…" — should activate → activated
    ✓ "I want to profile my MSBuild build to understand where time…" — should activate → activated
    ✓ "Our CI builds are fast but local dev builds are painfully s…" — should activate → activated
    ✓ "How do I add a NuGet package reference to my project?" — should NOT activate → not activated
    ✓ "My unit tests are failing with a NullReferenceException" — should NOT activate → not activated
    ✓ "How do I configure Docker for my .NET application?" — should NOT activate → not activated
    ✓ "What's the difference between .NET 8 and .NET 9?" — should NOT activate → not activated
    ✓ "How do I set up Entity Framework Core migrations?" — should NOT activate → not activated

### DIFF
--- a/eng/skill-validator/src/Commands/ValidateCommand.cs
+++ b/eng/skill-validator/src/Commands/ValidateCommand.cs
@@ -30,6 +30,9 @@ public static class ValidateCommand
         var reporterOpt = new Option<string[]>("--reporter") { Description = "Reporter (console, json, junit, markdown). Can be repeated.", AllowMultipleArgumentsPerToken = true };
         var noOverfittingCheckOpt = new Option<bool>("--no-overfitting-check") { Description = "Disable LLM-based overfitting analysis (on by default)" };
         var overfittingFixOpt = new Option<bool>("--overfitting-fix") { Description = "Generate a fixed eval.yaml with improved rubric items/assertions" };
+        var selectivityTestOpt = new Option<bool>("--selectivity-test") { Description = "Run selectivity test using should_activate / should_not_activate prompts from eval.yaml" };
+        var selectivityMinRecallOpt = new Option<double>("--selectivity-min-recall") { Description = "Minimum recall (activation on should_activate prompts) to pass (0-1)", DefaultValueFactory = _ => 0.8 };
+        var selectivityMinPrecisionOpt = new Option<double>("--selectivity-min-precision") { Description = "Minimum precision (non-activation on should_not_activate prompts) to pass (0-1)", DefaultValueFactory = _ => 0.8 };
 
         var command = new RootCommand("Validate that agent skills meaningfully improve agent performance")
         {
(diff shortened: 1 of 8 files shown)


DeagleGross:
> We already have that in SkillValidator. cc @JanKrivanek

Do you mind elaborating? I could not find a fast-running option to confirm that skill is activated per number of prompts and not activated per another. I think this is very important to build a heavy-suite of such prompts to ensure skills are not loaded for majority of prompts.

I know a full-evaluation run of agent will fail if skill is not activated, but that is not the one I've built here.

PR #298, opened 2026-03-09
Add the C# LSP configuration to the 'dotnet' plugin

jasonmalinowski:


README.md
```diff
@@ -13,6 +13,7 @@ This repository contains the .NET team's curated set of core skills and custom a
 | [dotnet-diag](plugins/dotnet-diag/) | Skills for .NET performance investigations, debugging, and incident analysis. |
 | [dotnet-msbuild](plugins/dotnet-msbuild/) | Comprehensive MSBuild and .NET build skills: failure diagnosis, performance optimization, code quality, and modernization. |
 | [dotnet-upgrade](plugins/dotnet-upgrade/) | Skills for migrating and upgrading .NET projects across framework versions, language features, and compatibility targets. |
+| [csharp-tools](plugins/csharp-tools/) | Navigate and interact with your C# code using the power of Roslyn. |
```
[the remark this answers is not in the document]
jasonmalinowski: Even if we want it higher, it shouldn't be first.

plugins/csharp-tools/plugin.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "csharp-tools",
+  "description": "Navigate and interact with your C# code using the power of Roslyn.",
+  "version": "0.0.1",
```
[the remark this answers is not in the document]
jasonmalinowski: This matches the version of this plugin distributed in other places; we want to keep that in sync.

plugins/csharp-tools/plugin.json
```diff
@@ -0,0 +1,15 @@
...
+  "author": {
+    "name": "Roslyn Team"
+  },
+  "keywords": [
+    "dotnet",
+    "csharp",
+    "refactoring",
+    "code-analysis"
+  ],
```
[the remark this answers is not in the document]
jasonmalinowski: This matches the same manifest of this plugin we've distributed in other locations.

.github/CODEOWNERS
```diff
@@ -58,3 +58,6 @@
 # dotnet-maui
 /plugins/dotnet-maui/ @Redth @jfversluis
 /tests/dotnet-maui/ @Redth @jfversluis
+#
```
jcouv: Is this intentional?

jasonmalinowski: Nope, fixed.

CONTRIBUTING.md
```diff
@@ -54,6 +54,7 @@ Skills are grouped into domain-specific plugins. When proposing a new skill, pla
 | `dotnet-diag` | Performance investigations, debugging, and incident analysis |
 | `dotnet-data` | Data access and Entity Framework |
 | `dotnet-msbuild` | MSBuild and project system |
+| `csharp-tools` | The C# LSP server and related tooling |
```
AbhitejJohn: Did we already consider putting this in the `dotnet` bucket? From the description, it does seem like it would fall under there. I'm a little confused what csharp-tools means in here at the moment - For instance, what else would we include in that category?

danmoseley: yeah as this is our first LSP one I don't know how to think about it. essentially -- are there important reasons why one might want the "dotnet" bucket (aka plugin) but not this one or vice versa?

cc @ManishJayaswal

danmoseley: My 2c, separate likely makes sense as LSP has different use cases to skills, it's more of a run time tool than a plugin, and each of the two has significant cost so it would be undesirable to couple them.

Plus, there will likely be more we want in csharp-tools in future? eg., refactoring skills?

timheuer: IMO this should be in the root default `dotnet` plugin definition. Plugins are containers of agent 'stuff' skills (ref: https://docs.github.com/en/copilot/reference/cli-plugin-reference#component-path-fields) and this is appropriate to put here.

ManishJayaswal: +1 to Tim's comment. This should go into dotnet as this is going to be generally useful in almost all context.

jasonmalinowski: > Plus, there will likely be more we want in csharp-tools in future? eg., refactoring skills?

Possibly, yes, but then it's hard to understand why that skill would be in csharp-tools versus dotnet. Already we've got [other PRs trying to add refactorings to the main dotnet plugin](https://github.com/dotnet/skills/pull/269), so we need one place for those.

JoeRobich (approved):

CONTRIBUTING.md
jasonmalinowski: I've redone the PR to just put this in the dotnet plugin. @timheuer or @ManishJayaswal if you want to re-review that would be appreciated.

plugins/dotnet/plugin.json
```diff
@@ -2,5 +2,6 @@
   "name": "dotnet",
   "version": "0.1.0",
   "description": "Common everyday C#/.NET coding skills. Expected to be useful to all .NET developers.",
-  "skills": "./skills/"
+  "skills": "./skills/",
+  "lspServers": "lsp.json"
```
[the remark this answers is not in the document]
jasonmalinowski: Already changed it.

plugins/dotnet/lsp.json
```diff
@@ -0,0 +1,19 @@
+{
+  "lspServers": {
+    "csharp": {
+      "command": "dotnet",
+      "args": [
+        "dnx",
+        "roslyn-language-server",
+        "--yes",
+        "--prerelease",
```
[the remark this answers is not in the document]
jasonmalinowski: This is planned to do, we're still working on what that flow would look like.

PR #299, opened 2026-03-09
Improve evaluation results table readability and add verdict explanations

danmoseley:
Improve evaluation results table readability and add verdict explanations

## Problem

The evaluation results table shows a Verdict column based on a weighted composite improvement score, but when the verdict disagrees with the quality delta the reasoning is opaque. Users must dig into workflow logs to understand why a scenario failed. The table is also wide (9 columns) which renders poorly on narrow GitHub viewports.

Example from PR #180: a scenario showed quality 4.3 -> 4.7 (+0.4) but got a fail verdict because the composite was -24.8% due to token and time overhead.

### Compact quality column

Merged the old Baseline, With Skill, and Delta columns into a single **Quality** column:
- Shows `baseline -> **skill**` with the better score bolded and a color indicator (🟢 improved / 🔴 regressed)
- Reduces the table from 9 columns to 6

### Verdict explanation footnotes

Added footnote references in the **Verdict** column when the verdict disagrees with the quality direction. Footnote references are bidirectional hyperlinks — clicking `[1]` in the table jumps to the explanation, and clicking `[1]` in the footnote jumps back to the row. Footnotes appear below the table with:
- A readable explanation of why the verdict disagrees
- The weighted composite score
- Raw metric values so the reader can see magnitude at a glance

`BuildVerdictFootnote()` returns `null` (no footnote) when:
- The verdict is self-explanatory from the quality direction
- The verdict is neutral
- The quality delta is unknown

When quality is unchanged (delta exactly zero) but the verdict is negative, a footnote is emitted explaining which efficiency metrics caused the failure.

### After this change, the results table looks like:

> | Skill | Scenario | Quality | Skills Loaded | Overfit | Verdict |
> |-------|----------|---------|---------------|---------|---------|
> | my-skill | scenario-1 | 4.3/5 -> **4.7/5** &#x1f7e2; | ... | ... | &#x274c; <a href="#user-content-fn-ex1" id="ref-ex1">[1]</a> |
> | my-skill | scenario-2 | 3.0/5 -> **4.5/5** &#x1f7e2; | ... | ... | &#x2705; |
>
> <a href="#user-content-ref-ex1" id="fn-ex1"><strong>[1]</strong></a> Quality improved but weighted score is -24.8% due to: tokens (1200 -> 9500), time (2.1s -> 8.3s)

### Test inflation commit

The second commit inflates the `csharp-scripts` skill description to trigger token overhead during evaluation, which should exercise the footnote feature. **Revert before merge.**

## Testing

- 6 markdown table tests cover: footnotes when verdict disagrees (both directions), no footnote when verdict agrees, footnote when quality unchanged but verdict negative, overfit column presence, and dash rendering
- Test `ImprovementScore` values are derived from `breakdown * DefaultWeights` for internal consistency
- All 284 tests pass

### DIFF
--- a/plugins/dotnet/skills/csharp-scripts/SKILL.md
+++ b/plugins/dotnet/skills/csharp-scripts/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: csharp-scripts
-description: "Run single-file C# programs as scripts for quick experimentation, prototyping, and concept testing. Use when the user wants to write and execute a small C# program without creating a full project. Supports NuGet package references via #:package directives, Unix shebang execution, source-generated JSON for AOT compatibility, and automatic project conversion. File-based apps require .NET 10+ SDK; for older SDKs, falls back to temporary console project scaffolding with dotnet new console."
+description: Run single-file C# programs as scripts for quick experimentation, prototyping, and concept testing. Use when the user wants to write and execute a small C# program without creating a full project.
 ---
 
 # C# Scripts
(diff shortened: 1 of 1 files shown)


danmoseley:
I think this is ready to review/merge.

plugins/dotnet/skills/csharp-scripts/SKILL.md
```diff
@@ -162,3 +162,17 @@ Replace the generated `Program.cs` with the script content and run with `dotnet
 ## More info
 
 See https://learn.microsoft.com/en-us/dotnet/core/sdk/file-based-apps for a full reference on file-based apps.
+
+## Extended Reference (TEST PADDING - REVERT BEFORE MERGE)
```
JanKrivanek: Please remove :-)

PR #304, opened 2026-03-10
Purge dashboard entries older than 14 days

ViktorHofer:
When updating plugin JSON files, filter out Quality and Efficiency entries whose date is older than 14 days. If all entries are purged, delete the JSON file so it is excluded from the `components.json` manifest and no longer appears in the dashboard.

### Changes
- **`generate-benchmark-data.ps1`**: After appending new entries, purge any with a `date` older than 14 days. If no entries remain, delete the file (or skip writing it).
- **`dashboard.html`**: Updated footer text to reflect the 14-day retention window.

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -226,7 +226,7 @@ jobs:
             & ./eng/dashboard/generate-benchmark-data.ps1 @params
           }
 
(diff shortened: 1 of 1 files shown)


JanKrivanek (approved):
LGTM

The copilot comments seems valid though

PR #318, opened 2026-03-10
Disable scheduled workflows on forks

JanKrivanek:
### Motivation

The scheduled runs have no value on forks and unnecessarily consume resources and tokens

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -19,12 +19,13 @@ permissions:
 
 jobs:
   discover:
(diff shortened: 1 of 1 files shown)


.github/workflows/evaluation.yml
ViktorHofer, line 1: I think this was already handled correctly in this file. Not sure.

ViktorHofer (approved):

.github/workflows/evaluation.yml
JanKrivanek: Yeah - you are right. Check for explicit repo name is stronger check than just "not a fork"
Reverting to the original check here

PR #330, opened 2026-03-11
adding SIMD/TensorPrimitives to dotent-diag/analyzing-dotnet-performance

jeffschwMSFT:
Incorporated feedback, trained on 100+ scalar loop examples, moved to analyzing-dotnet-performance reference

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -80,6 +80,9 @@
 /plugins/dotnet-experimental/skills/exp-test-tagging/ @dotnet/dotnet-testing
 /tests/dotnet-experimental/exp-test-tagging/ @dotnet/dotnet-testing
 
+/plugins/dotnet-experimental/skills/exp-simd-vectorization/ @jeffschw @artl93
+/tests/dotnet-experimental/exp-simd-vectorization/ @jeffschw @artl93
+
 # dotnet-maui
 /plugins/dotnet-maui/ @Redth @jfversluis
 /tests/dotnet-maui/ @Redth @jfversluis


plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
+# SIMD Vectorization
+
+## Decision Gate
+
+1. **Check for TensorPrimitives first.** If one or more TensorPrimitives methods cover the operation → use them. If the `.csproj` does NOT already reference `System.Numerics.Tensors`, **add the package**, for example: `<PackageReference Include="System.Numerics.Tensors" />` (or use the versioning approach already used by your solution). Then replace the scalar loop with TP calls and stop. See the full API table below. Compose multiple TP calls when needed (e.g., finding both min and max → `TensorPrimitives.Min(span)` + `TensorPrimitives.Max(span)` as two calls). Do NOT write manual Vector128 code for operations TP already handles.
```
tannergooding: nit: It might be good to check `Span<T>` (and extensions) first, that way no additional dependency is needed. Then `TensorPrimitives` as a secondary.

Also noting if we use this skill in dotnet/runtime, we should remember to add a special condition for `System.Private.Corelib` since it is the foundational layer.

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
...
+
+2. **Scalar loop over contiguous array/span** of `byte`, `short`, `int`, `long`, `float`, `double`? → Implement with explicit `Vector128<T>` / `Vector256<T>` / `Vector512<T>` intrinsics using the patterns below.
```
tannergooding: nit: This misses the other supported types (`sbyte`, `ushort`, `uint`, `ulong`, `nint`, and `nuint`). We can also handle `char` via reinterpretation as `ushort` or `short`.

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
...
+
+3. **No contiguous numeric array processing** (dictionary lookups, tree traversals, linked lists, state machines, string formatting, small collections, enum comparisons, recursive algorithms, decimal arithmetic)? → Report `[NO SIMD OPPORTUNITY]` and write a **full paragraph** explaining WHY, referencing the specific code characteristics that prevent vectorization (e.g., "State machines require sequential branching on enum values — there are no contiguous numeric arrays to process in parallel, and each transition depends on the previous state"). This explanation is graded.
+
+## TensorPrimitives API Reference
+
+Use for any float/double array operation that has a matching API below. If the project doesn't already reference `System.Numerics.Tensors`, add it to the `.csproj`. Replace the entire manual loop with **one or more** `TensorPrimitives` calls as needed (prefer a single call when possible):
```
tannergooding: nit: This shouldn't be limited to just `float/double`. It should work for any of the primitive types provided the API meets the constraints.

For example, `Sum` is constrained `where T : System.Numerics.IAdditionOperators<T,T,T>, System.Numerics.IAdditiveIdentity<T,T>` and so works for any of the primitive types.

The same is true for something like `Min` which is constrained to `where T : System.Numerics.INumber<T>`.

`CosineSimilarity` on the other hand is constrained to `where T : System.Numerics.IRootFunctions<T>` and so only works for float/double, as those are the only primitive types that implement `IRootFunctions`

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
...
+| Multiply | `TensorPrimitives.Multiply(a, b, dst)` |
+| Divide | `TensorPrimitives.Divide(a, b, dst)` |
+| Element-wise Min | `TensorPrimitives.Min(a, b, dst)` |
+| Element-wise Max | `TensorPrimitives.Max(a, b, dst)` |
+
+### Three-span fused operations
+| Operation | API |
+|-----------|-----|
+| (a+b)*c | `TensorPrimitives.AddMultiply(a, b, c, dst)` |
+| a*b+c (FMA) | `TensorPrimitives.MultiplyAdd(a, b, c, dst)` |
```
tannergooding: This one needs to be `TensorPrimitives.FusedMultiplyAdd`

`AddMultiply(x, y, z)` does `(x + y) * z` while `MultiplyAdd(x, y, z)` does `(x * y) + z`

`FusedMultiplyAdd(x, y, z)` is what does `fma(x, y, z)` (which is the fused form of `(x * y) + z`).

Both `AddMultiply` and `MultiplyAdd` exist because it optimizes differently depending on where the dependency chain flows from (the addend or the multiplier)

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
...
+
+Use this when TensorPrimitives doesn't have a single API for the operation. This is required for byte-level operations, character class counting, range validation, bitwise bulk ops, cross-type conversions, and custom patterns.
+
+### Required imports
+```csharp
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
+```
+Never use `System.Runtime.Intrinsics.X86` or `.Arm` — cross-platform APIs only.
```
tannergooding: nit: It's not *never* use, particularly if there is some significant perf advantage.

Rather its we prefer singular portable code paths if there isn't significant advantage compared to the increased code complexity of using platform specific APIs.

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
...
+{
+    uint vec512Count = (uint)Vector512<T>.Count;
+    while (i + vec512Count <= length)
+    {
+        var vec = Vector512.LoadUnsafe(ref src, i);
+        // ... process vec ...
+        i += vec512Count;
+    }
+}
+if (Vector256.IsHardwareAccelerated && Vector256<T>.IsSupported)
```
tannergooding: nit: this fallthrough pattern ends up pessimizing the scalar case, which is hit for small inputs, as it causes up to 3 loops to be hit and not executed (possibly mispredicting)

Ideally you have the small input case only have one branch leading up to it if possible, which is how the TensorPrimitives operations work.

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,191 @@
...
+for (; i < length; i++)
+{
+    // ... scalar processing ...
+}
+```
+
+### Core SIMD operations
+- **Load/Store:** `Vector128.LoadUnsafe(ref src, offset)` / `.StoreUnsafe(ref dst, offset)`
+- **Arithmetic:** `+`, `-`, `*`, `/` operators on vector types
+- **FMA:** `Vector128.MultiplyAddEstimate(a, b, c)` — fused multiply-add
```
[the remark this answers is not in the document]
tannergooding: The copilot comment is correct. MultiplyAddEstimate is the faster of `fma(x, y, z)` or `(x * y) + z` and isn't strictly deterministic across hardware. It's for cases where perf matters more than determinism or the 1 ulp difference is irrelevant to the algorithm

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
jeffschwMSFT: good point about System.Private.Corelib.  this is 3rd party first, but if we find valuable for dotnet/runtime work, additional checks are needed

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,196 @@
...
+
+3. **Scalar loop over contiguous array/span** of `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `nint`, `nuint`, `float`, `double` (and `char` via reinterpretation as `ushort`)? → Implement with explicit `Vector128<T>` / `Vector256<T>` / `Vector512<T>` intrinsics using the patterns below.
+
+4. **No contiguous numeric array processing** (dictionary lookups, tree traversals, linked lists, state machines, string formatting, small collections, enum comparisons, recursive algorithms, decimal arithmetic)? → Report `[NO SIMD OPPORTUNITY]` and write a **full paragraph** explaining WHY, referencing the specific code characteristics that prevent vectorization (e.g., "State machines require sequential branching on enum values — there are no contiguous numeric arrays to process in parallel, and each transition depends on the previous state"). This explanation is graded.
```
tannergooding: Should this be "Non contiguous" instead of "No contiguous"?

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
```diff
@@ -0,0 +1,196 @@
...
+for (; i < length; i++)
+{
+    // ... scalar processing ...
+}
+```
+
+### Core SIMD operations
+- **Load/Store:** `Vector128.LoadUnsafe(ref src, offset)` / `.StoreUnsafe(ref dst, offset)`
+- **Arithmetic:** `+`, `-`, `*`, `/` operators on vector types
+- **Multiply-add (approximate):** `Vector128.MultiplyAddEstimate(a, b, c)` — performs a multiply-add with implementation-defined approximation; not guaranteed to be a strict IEEE fused multiply-add. For precise fused semantics, use hardware-specific intrinsics such as `System.Runtime.Intrinsics.X86.Fma.MultiplyAdd` when available.
```
tannergooding: We don't need the platform specific APIs here, you can just use `Vector128.FusedMultiplyAdd(a, b, c)` for precise fused semantics

plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/simd-vectorization.md
jeffschwMSFT: I debated this with the model, it felt that this language was clear to it.  I asked it for a compromise and we made a slight change

tannergooding: Gotcha. I personally found them to be polar opposite statements.

One is stating that it "should not perform simd on numeric arrays which are contiguous" the other stating that it "should not perf simd numeric arrays which are not contiguous"

PR #358, opened 2026-03-13
Add skill for using `dotnet format whitespace`

jjonescz:


danmoseley:
@jjonescz in the eval above, in the first case baseline (no skill) gives the same result but 30% faster (didn't need to read in the skill). In the second case, the skill doesn't activate at all.

edit: in the first case, with the skill added it actually didn't even complete apparently in the 120s -- not sure why it got graded anyway

If you click on the link it will give
> For PR 358 in dotnet/skills, download eval artifacts with gh run download 24213650400 --repo dotnet/skills --pattern "skill-validator-results-*" --dir ./eval-results, then fetch https://raw.githubusercontent.com/dotnet/skills/6a308d7d9b2dee2a72a39fcfc37babd26ff874e1/eng/skill-validator/src/docs/InvestigatingResults.md and follow it to analyze the results.json files. Diagnose each failure, suggest fixes to the eval.yaml and skill content, and tell me what to fix first.

this is handy to have AI look at the actual logs and you can ask it eg why did it not activate? But I think you might need to devise more eval scenario(s) that show why the skill is needed. One way to start might be to try locally -- without the skill -- and give it a task in Copilot CLI with some model or other and see what fails that you want to work. Then put that in an eval.

What I've learned working on skills for this repo is it's often surprisingly hard to beat the agent-without-skill results. You start at -100 points because it has to read and interpret the skill - and because it uses up context window too

For the ability to run '/evaluate' @ViktorHofer do we have a group to add Jan to?

ViktorHofer:
Just maintainer permissions are needed, no special group. You can reach out to @timheuer to get the required perms.

AbhitejJohn:
@jjonescz : Just checking in to see if this is still something you are looking to undraft and add.

jjonescz:
probably not, it seems copilot can do this usually on its own these days

PR #366, opened 2026-03-14
aot-compat: batch fixes by warning code instead of small batches

danmoseley:
## Batch aot-compat fixes by warning code instead of small batches

The current guidance tells the agent to fix 5-10 warnings then rebuild immediately. For large codebases like Azure.ResourceManager (461 files, 135+ JSON serialization call sites), this creates ~20-30 build-fix cycles, resulting in very long agent sessions (69 turns, 167 tool calls, ~17 minutes).

### Change

Replace "fix a small batch of 5-10, then rebuild" with "fix an entire warning code across the project before rebuilding." Target 2-4 total build-fix cycles instead of one per small batch.

### Why this is likely a better strategy for users too

The small-batch approach was defensive: catch mistakes early. But for an AI agent:

- **Most of the time is spent waiting for builds and re-reading output**, not fixing code. Each rebuild cycle grows the conversation context, making the agent slower and more likely to lose track late in the session.
- **Batching is how an experienced developer works.** You'd grep for all `JsonSerializer.Serialize` calls, create the context class, update them all, then rebuild once. You wouldn't rebuild after every 5 files.
- **If the agent makes a systematic mistake** (e.g., wrong type name), the next rebuild catches it immediately and the fix is also systematic. One extra cycle, not a disaster.

### Evaluation cost impact

This skill is the single largest token consumer in evaluation: 30% of all agent tokens across all plugins. In evaluation, the agent session accumulates context quadratically with turn count. Reducing turns from ~69 to ~20-25 could cut this skill's token usage by 60-70%.

### Suggestion

Run an `/evaluate` on this PR to compare before/after turn counts and scores.


danmoseley:
## Evaluation comparison: old vs new skill wording

Comparing `withSkill` metrics for the `dotnet-aot-compat` / `Make Azure.ResourceManager AOT-compatible` scenario.

| Metric | Old wording (PR #362 run) | New wording (this PR) | Change |
|--------|--------------------------|----------------------|--------|
| Turns | 73 | 81 | +11% |
| Tool calls | 201 | 221 | +10% |
| Tokens (withSkill) | 2,872,468 | 3,091,101 | +8% |
| Wall time | 1050s | 1050s | +0% |

Baseline (no skill) for reference:

| Metric | Old wording baseline | New wording baseline |
|--------|---------------------|---------------------|
| Turns | 108 | 78 |
| Tool calls | 215 | 170 |
| Tokens | 4,512,316 | 3,854,570 |

### Analysis

Turn count is roughly the same. The batching guidance did not significantly change agent behavior for this scenario.

Skill passed: False, improvement score: -0.1460433979067025

---
*Baseline run: [23077277174](https://github.com/dotnet/skills/actions/runs/23077277174) (PR #362, old wording, runs=1). New run: [23078509958](https://github.com/dotnet/skills/actions/runs/23078509958) (this PR, runs=3).*

danmoseley:
## Deep analysis: old vs new skill wording

### The headline: both versions time out, so the comparison is mostly noise

**Every single run — old and new, baseline and withSkill — hit the 1050s timeout.** The agent never finishes the task. We're comparing two incomplete attempts, which makes score differences largely meaningless.

### What the judge actually said

**Old wording (withSkill):** "Timed out after 1050 seconds having completed only a tiny fraction of the work... spent the vast majority of its time exploring and reading files (105+ view calls, 4 explore tasks) versus making actual changes (7 edits, 1 file creation). **It never ran a single build.**"

**New wording (withSkill):** "Timed out after 1050 seconds having completed only a small fraction of the needed work... attempted iterative builds but the process was haphazard and incomplete. The final build shows Build FAILED with error CS0653."

The new wording actually got the agent to *build* (an improvement in approach), but it still didn't finish. The old wording spent all its time reading files and never built at all.

### The numbers in context

| Metric | Old withSkill | New withSkill | Note |
|--------|--------------|---------------|------|
| Turns | 73 | 81 | Both hit timeout |
| Tool calls | 201 | 221 | More tools = agent was building, not just reading |
| Tokens | 2.87M | 3.09M | ~8% more (noise range for timed-out runs) |
| Wall time | 1050s | 1050s | Both timed out |
| Judge score | 1.0 | 1.7 | New slightly better (built, used annotations) |

The baseline (no skill) actually varied more between runs: 108 turns/4.5M tokens (old) vs 78 turns/3.9M tokens (new), showing high natural variance.

### Per-run variance is huge

The 3 runs of the new wording scored: **-0.20, +0.22, -0.03**. That's a swing from "skill made it worse" to "skill helped significantly" across runs of the same wording. This tells us **single-run comparisons are unreliable for this scenario**.

### Root cause: the scenario is too large for the timeout

The Azure.ResourceManager test case (461 files, 135+ JSON call sites) simply can't be completed in 1050 seconds by any approach — batching or not. The agent consistently times out regardless of strategy. The skill wording change we made is about *how to batch within the session*, but the session ends before any approach can finish.

### Recommendation

**Don't merge this PR based on these results** — the data is inconclusive due to timeouts masking any strategy difference. To properly evaluate the batching guidance:

1. **Increase the scenario timeout** for aot-compat (e.g., to 1800s or 2400s) so the agent can actually finish, then compare
2. **Or test on a smaller codebase** where the agent can complete within 1050s, to isolate whether batching actually changes behavior
3. **Or reduce the test case scope** — use a project with ~20 JSON call sites instead of 135+

The batching guidance is still likely better in principle (the judge noted the new wording got the agent to actually build, which is progress), but we can't prove it with a scenario that always times out.

PR #373, opened 2026-03-16
Add VSTest to MTP migration skill

Evangelink:
Replaces #346

### DIFF
--- a/plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
@@ -39,15 +39,6 @@ Migrate a .NET test solution from VSTest to Microsoft.Testing.Platform (MTP). Th
 - The solution mixes .NET and non-.NET test adapters (e.g., JavaScript or C++ adapters) — VSTest is required
 - Upgrading MSTest versions — use `migrate-mstest-v1v2-to-v3` or `migrate-mstest-v3-to-v4`
 
-## Inputs
-
-| Input | Required | Description |
-|-------|----------|-------------|
-| Project or solution path | Yes | The `.csproj`, `.sln`, or `.slnx` entry point containing test projects |
-| Test framework | Yes | MSTest, NUnit, or xUnit.net. Auto-detect from package references if not provided |
-| .NET SDK version | No | The SDK version used to build. Determines `dotnet test` integration mode. Auto-detect from `global.json` or `dotnet --version` |
-| CI system | No | Azure DevOps, GitHub Actions, or other. Determines pipeline update guidance |
-
 ## Workflow
 
 ### Step 1: Assess the solution
@@ -66,19 +57,7 @@ Migrate a .NET test solution from VSTest to Microsoft.Testing.Platform (MTP). Th
 
 > **Critical**: Always set MTP properties in `Directory.Build.props` at the solution or repo root — never per-project. This prevents inconsistent configuration where some projects use VSTest and others use MTP (an unsupported scenario).
 
-Create or update `Directory.Build.props` with the MTP properties. The exact properties depend on the framework (see Step 3):
-
-```xml
-<!-- Directory.Build.props -->
-<Project>
-  <PropertyGroup>
-    <!-- Framework-specific runner property goes here (Step 3) -->
-    <!-- dotnet test integration property goes here (Step 4) -->
-  </PropertyGroup>
-</Project>
-```
-
-> **Note**: The framework-specific runner properties (`EnableMSTestRunner`, `EnableNUnitRunner`) automatically set `<OutputType>Exe</OutputType>` on projects that reference the corresponding test framework. You do not need to set `OutputType` manually. MSTest.Sdk also sets `EnableMSTestRunner` automatically. For xUnit.net, `YTest.MTP.XUnit2` handles this automatically.
+> **Note**: MTP requires test projects to have `<OutputType>Exe</OutputType>`. Only `MSTest.Sdk` sets this automatically. For all other setups (MSTest NuGet packages with `EnableMSTestRunner`, NUnit with `EnableNUnitRunner`, xUnit.net with `YTest.MTP.XUnit2`), you must set `<OutputType>Exe</OutputType>` explicitly — either per-project or in `Directory.Build.props` with a condition that targets only test projects.
 
 ### Step 3: Enable the framework-specific MTP runner
 
@@ -91,6 +70,7 @@ Each framework has its own opt-in property. Add these in `Directory.Build.props`
 ```xml
 <PropertyGroup>
   <EnableMSTestRunner>true</EnableMSTestRunner>
+  <OutputType>Exe</OutputType>
 </PropertyGroup>
 ```
 
@@ -115,15 +95,22 @@ Requires `NUnit3TestAdapter` **5.0.0** or later.
 ```xml
 <PropertyGroup>
   <EnableNUnitRunner>true</EnableNUnitRunner>
+  <OutputType>Exe</OutputType>
 </PropertyGroup>
 ```
 
 #### xUnit.net
 
-Add a reference to `YTest.MTP.XUnit2` — this package provides MTP support for xUnit.net v2 projects without requiring an upgrade to xunit.v3:
+Add a reference to `YTest.MTP.XUnit2` — this package provides MTP support for xUnit.net v2 projects without requiring an upgrade to xunit.v3. You must also set `OutputType` to `Exe`:
 
 ```xml
 <PackageReference Include="YTest.MTP.XUnit2" Version="*" />
+```
+
+```xml
+<PropertyGroup>
+  <OutputType>Exe</OutputType>
+</PropertyGroup>
 ```
 
 > **Note**: `YTest.MTP.XUnit2` preserves the VSTest `--filter` syntax, so no filter migration is needed for xUnit.net v2. It also supports `--settings` for runsettings (xunit-specific configurations only), `xunit.runner.json`, TRX reporting via `--report-trx`, and `--treenode-filter`.
@@ -205,26 +192,32 @@ VSTest-specific arguments must be translated to MTP equivalents. Build-related a
 
 | VSTest `--filter` syntax | xUnit.net v3 MTP equivalent | Notes |
 |---|---|---|
-| `FullyQualifiedName~ClassName` | `--filter-class ClassName` | Substring match on class name |
+| `FullyQualifiedName~ClassName` | `--filter-class *ClassName*` | Wildcards required for substring match |
 | `FullyQualifiedName=Ns.Class.Method` | `--filter-method Ns.Class.Method` | Exact match on fully qualified method |
-| `Name=MethodName` | `--filter-method MethodName` | Match on method display name |
+| `Name=MethodName` | `--filter-method *MethodName*` | Wildcards for substring match |
 | `Category=Value` (trait) | `--filter-trait "Category=Value"` | Filter by trait name/value pair |
-| Complex `&`/`\|` expressions | `--filter-query "expr"` | Use [xUnit.net query filter language](https://xunit.net/docs/query-filter-language) |
+| Complex expressions | `--filter-query "expr"` | Uses [xUnit.net query filter language](https://xunit.net/docs/query-filter-language) |
 
-Example translation:
+`--filter-query` uses a path-segment syntax with four segments and an optional trait filter:
+
+```
+/<assemblyFilter>/<namespaceFilter>/<classFilter>/<methodFilter>[traitName=traitValue]
+```
+
+Each segment matches against: assembly name, namespace, class name, method name. Use `*` for "match all" in any segment. Documentation: https://xunit.net/docs/query-filter-language
 
 ```shell
 # VSTest
 dotnet test --filter "FullyQualifiedName~IntegrationTests&Category=Smoke"
 
-# xUnit.net v3 MTP — using individual filters
-dotnet test -- --filter-class IntegrationTests --filter-trait "Category=Smoke"
+# xUnit.net v3 MTP — using individual filters (AND behavior)
+dotnet test -- --filter-class *IntegrationTests* --filter-trait "Category=Smoke"
 
-# xUnit.net v3 MTP — using query language
-dotnet test -- --filter-query "class~IntegrationTests AND trait('Category', 'Smoke')"
+# xUnit.net v3 MTP — using query language (assembly/namespace/class/method[trait])
+dotnet test -- --filter-query "/*/*/*IntegrationTests*/*[Category=Smoke]"
 ```
 
-> **Note**: When combining `--filter-class` and `--filter-trait`, both conditions must match (AND behavior). For OR logic or complex expressions, use `--filter-query`.
+> **Note**: When combining `--filter-class` and `--filter-trait`, both conditions must match (AND behavior). For complex expressions, use `--filter-query` with the path-segment syntax. See the [xUnit.net query filter language docs](https://xunit.net/docs/query-filter-language) for full reference.
 
 ### Step 6: Install MTP extension packages (if needed)
 
@@ -265,7 +258,7 @@ If CI scripts use TRX reporting, crash dumps, or hang dumps, add the correspondi
     arguments: '--no-build --configuration Release'
 ```
 
-**If using DotNetCoreCLI@2 without .NET 10 native MTP**: Add `--` before MTP arguments:
+**If already using DotNetCoreCLI@2**: Update arguments per Step 5 translations. Remember the `--` separator on .NET 9 and earlier:
 
 ```yaml
 - task: DotNetCoreCLI@2
@@ -273,16 +266,6 @@ If CI scripts use TRX reporting, crash dumps, or hang dumps, add the correspondi
   inputs:
     command: 'test'
     arguments: '--no-build -- --report-trx --results-directory $(Agent.TempDirectory)'
-```
-
-**If using DotNetCoreCLI@2 with .NET 10 native MTP mode** (global.json configured): Pass MTP arguments directly:
-
-```yaml
-- task: DotNetCoreCLI@2
-  displayName: Run tests
-  inputs:
-    command: 'test'
-    arguments: '--no-build --report-trx --results-directory $(Agent.TempDirectory)'
 ```
 
 #### GitHub Actions
@@ -310,14 +293,6 @@ VSTest silently succeeds when zero tests are discovered. MTP fails with **exit c
 
 - Use environment variable: `TESTINGPLATFORM_EXITCODE_IGNORE=8`
 
-#### Console.InputEncoding preservation
-
-MTP always preserves the console codepage. If CI sets `chcp 65001` (UTF8) and tests start child processes with `CreateNoWindow = true`, those child processes may receive a UTF8 BOM that VSTest previously did not pass. Workaround:
-
-```csharp
-Console.InputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
-```
-
 ### Step 9: Remove VSTest-only packages (optional)
 
 Once migration is complete and verified, remove packages that are only needed for VSTest:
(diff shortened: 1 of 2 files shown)


plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+  Covers runner enablement, CLI argument translation, Directory.Build.props
+  and global.json configuration, CI/CD pipeline updates, and MTP extension
+  packages. DO NOT USE FOR: migrating between test frameworks
+  (MSTest/xUnit/NUnit), MSTest version upgrades (use migrate-mstest-* skills),
+  TFM upgrades, or UWP/WinUI test projects.
+---
+
+# VSTest → Microsoft.Testing.Platform Migration
+
+Migrate a .NET test solution from VSTest to Microsoft.Testing.Platform (MTP). MTP is an executable-first test platform that supports Native AOT, trimming, `dotnet run`, `dotnet watch`, and direct executable execution. The outcome is a solution where all test projects run on MTP, `dotnet test` works correctly, and CI/CD pipelines are updated.
```
Youssef1313: > MTP is an executable-first test platform that supports Native AOT, trimming, `dotnet run`, `dotnet watch`, and direct executable execution.

I'm not sure if this definition is helping AI for anything or if it's polluting the context with no value. Are we able to compare some evaluation with and without it?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+| Test framework | Yes | MSTest, NUnit, or xUnit.net. Auto-detect from package references if not provided |
+| .NET SDK version | No | The SDK version used to build. Determines `dotnet test` integration mode. Auto-detect from `global.json` or `dotnet --version` |
+| CI system | No | Azure DevOps, GitHub Actions, or other. Determines pipeline update guidance |
+
+## Workflow
+
+### Step 1: Assess the solution
+
+1. Identify the test framework for each test project:
+   - **MSTest**: References `MSTest`, `MSTest.TestFramework`, or uses `MSTest.Sdk`
```
Youssef1313: - Are we missing MSTest.TestAdapter here?
- Referencing MSTest.TestFramework alone doesn't make the project a test project. It will usually be just a library.
- For MSTest.Sdk, should we note that if `IsTestApplication` is false, this is not a test project?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+   - **NUnit**: References `NUnit` and `NUnit3TestAdapter`
```
Youssef1313: Should we simplify this to just this?


```suggestion
   - **NUnit**: References `NUnit3TestAdapter`
```

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+> **Critical**: Always set MTP properties in `Directory.Build.props` at the solution or repo root — never per-project. This prevents inconsistent configuration where some projects use VSTest and others use MTP (an unsupported scenario).
+
+Create or update `Directory.Build.props` with **all** required MTP properties up front. The exact properties depend on the framework (see Step 3), but `OutputType Exe` is always required:
+
+```xml
+<!-- Directory.Build.props -->
+<Project>
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
Youssef1313: In many cases this is not going to work well, as it will end up applying the OutputType to potentially non-test projects.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+    <!-- Framework-specific runner property goes here (Step 3) -->
+    <!-- dotnet test integration property goes here (Step 4) -->
+  </PropertyGroup>
+</Project>
+```
+
+If `Directory.Build.props` is shared with non-test projects, scope it with a condition:
+
+```xml
+<PropertyGroup Condition="'$(IsTestProject)' == 'true'">
```
Youssef1313: This is not going to work. Directory.Build.props is too early so we don't have IsTestProject available. (also IsTestProject is generally a VSTest-specific property)

And Directory.Build.targets is too late to set OutputType correctly.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+<PropertyGroup>
+  <EnableMSTestRunner>true</EnableMSTestRunner>
+</PropertyGroup>
+```
+
+Ensure the project references MSTest 3.2.0 or later. Recommend updating to the latest version.
+
+**Option B — MSTest.Sdk:**
+
+When using `MSTest.Sdk`, MTP is enabled by default — no `EnableMSTestRunner` or `OutputType Exe` property is needed (the SDK sets both automatically). The only action is: if the project has `<UseVSTest>true</UseVSTest>`, **remove it**. That property is what opts out of MTP.
```
Youssef1313: Should it be "out of VSTest"?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### NUnit
+
+Requires `NUnit3TestAdapter` **5.0.0** or later.
+
+1. Update `NUnit3TestAdapter` to 5.0.0+:
+
+```xml
+<PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
```
Youssef1313, line 90: Note: I recall that both NUnit and MSTest had some bugs with ordering of setting MSBuild properties etc that were fixed at some point. And so older versions of both adapters might not work well in some cases. So I'm not sure if we should find back the version with those bugs fixed and suggest that as a minimum version instead?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+```xml
+<PropertyGroup>
+  <EnableNUnitRunner>true</EnableNUnitRunner>
+</PropertyGroup>
+```
+
+#### xUnit.net
+
+Requires **xunit.v3** (v3 or later). If the project still uses xunit v2, it must be upgraded to v3 first.
```
Youssef1313: Should we have xunit 2 to xunit 3 migration as completely different skill? And go into much deeper details for the xunit.v3 breaking changes?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+The `dotnet test` integration depends on the .NET SDK version.
+
+#### .NET 10 SDK and later (recommended)
+
+Use the native MTP mode by adding a `test` section to `global.json`:
+
+```json
+{
+  "sdk": {
+    "version": "10.0.100"
```
Youssef1313, line 141: nit: use latest

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+> **Tip**: To preserve VSTest compatibility during a transition period, keep `xunit.runner.visualstudio` and `Microsoft.NET.Test.Sdk` alongside `xunit.v3`. Remove them once the migration is complete.
+
+### Step 4: Configure dotnet test integration
+
+The `dotnet test` integration depends on the .NET SDK version.
+
+#### .NET 10 SDK and later (recommended)
+
+Use the native MTP mode by adding a `test` section to `global.json`:
```
Youssef1313, line 136: IIRC global.json parsing doesn't support trailing commas.

The example below is **correct** in that it doesn't use trailing commas, but I'm not sure if we should explicitly state it to help LLMs further, or if it's not needed.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+  "sdk": {
+    "version": "10.0.100"
+  },
+  "test": {
+    "runner": "Microsoft.Testing.Platform"
+  }
+}
+```
+
+In this mode, `dotnet test` arguments are passed directly — no extra `--` separator is needed.
```
Youssef1313: I'm not sure if this makes it clearer to LLMs.

```suggestion
In this mode, `dotnet test` arguments are passed directly — For example, `dotnet test --report-trx`.
```

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### .NET 9 SDK and earlier
+
+Use the VSTest-bridge mode by adding this property in `Directory.Build.props`:
```
Youssef1313: Could it confuse this with Microsoft.Testing.Extensions.VSTestBridge which is a very different story? Is the following a clearer wording?


```suggestion
Use the VSTest mode of `dotnet test` command to run MTP test projects by adding this property in `Directory.Build.props`:
```

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+| `--blame-hang-dump-type <TYPE>` | `--hangdump-type <TYPE>` | Requires HangDump extension |
+| `--blame-hang-timeout <TIMESPAN>` | `--hangdump-timeout <TIMESPAN>` | Requires HangDump extension |
+| `--collect "Code Coverage;Format=cobertura"` | `--coverage --coverage-output-format cobertura` | Per-extension arguments |
+| `-d\|--diag <LOG_FILE>` | `--diagnostic` | |
+| `--filter <EXPRESSION>` | Framework-dependent (see below) | |
+| `-l\|--logger trx` | `--report-trx` | Requires `Microsoft.Testing.Extensions.TrxReport` NuGet package |
+| `--results-directory <DIR>` | `--results-directory <DIR>` | Same |
+| `-s\|--settings <FILE>` | `--settings <FILE>` | MSTest and NUnit still support `.runsettings`. MTP also supports `testconfig.json` |
+| `-t\|--list-tests` | `--list-tests` | Same |
+| `-- <RunSettings args>` | `--test-parameter` | Provided by VSTestBridge |
```
Youssef1313: Provided by VSTestBridge is probably too technical. Let's say applicable only to MSTest and NUnit instead?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+| `--blame-crash-dump-type <TYPE>` | `--crashdump-type <TYPE>` | Requires CrashDump extension |
+| `--blame-hang` | `--hangdump` | Requires `Microsoft.Testing.Extensions.HangDump` NuGet package |
+| `--blame-hang-dump-type <TYPE>` | `--hangdump-type <TYPE>` | Requires HangDump extension |
+| `--blame-hang-timeout <TIMESPAN>` | `--hangdump-timeout <TIMESPAN>` | Requires HangDump extension |
+| `--collect "Code Coverage;Format=cobertura"` | `--coverage --coverage-output-format cobertura` | Per-extension arguments |
+| `-d\|--diag <LOG_FILE>` | `--diagnostic` | |
+| `--filter <EXPRESSION>` | Framework-dependent (see below) | |
+| `-l\|--logger trx` | `--report-trx` | Requires `Microsoft.Testing.Extensions.TrxReport` NuGet package |
+| `--results-directory <DIR>` | `--results-directory <DIR>` | Same |
+| `-s\|--settings <FILE>` | `--settings <FILE>` | MSTest and NUnit still support `.runsettings`. MTP also supports `testconfig.json` |
```
Youssef1313: MTP also supports testconfig.json might confuse LLM more than it helps IMO. It might try to map stuff from XML format of runsettings directly to json and expect everything to work, which might not be the case?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+
+**MSTest and NUnit**: The `--filter` syntax is identical on both VSTest and MTP. No changes needed.
+
+**xUnit.net (breaking change)**: The VSTest `--filter` syntax is **not supported** on MTP. You must migrate to xUnit.net v3 native filter options. If your CI uses `--filter`, this is a required change.
+
+##### Simple filter translations
+
+| VSTest filter | xUnit.net MTP filter |
+|---------------|----------------------|
+| `--filter "FullyQualifiedName~MyClass"` | `--filter-class MyNamespace.MyClass` |
```
Youssef1313: Is `--filter-method *MyClass*` the more accurate form here?
And in that case we shouldn't probably say "MyClass". Instead pick up something random like `--filter "FullyQualifiedName~XYZ"` -> `--filter-method "*XYZ*"`?

This needs to be verified though.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+| `--filter "FullyQualifiedName~MyMethod"` | `--filter-method MyMethod` |
```
Youssef1313: Given my suggestion above, this line will be redundant.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+| `--filter "Category=Integration"` | `--filter-trait "Category=Integration"` |
```
Youssef1313: `--filter DisplayName=...` used to be valid for xunit with VSTest. Is there a filter for display name?

In addition, should we make this more generic and say `MyKey=MyValue`?

Generally I think this needs to be structured a bit in a way that maps VSTest filter key to the right command-line option, and maps VSTest filter operator to how the value of the option is provided (e.g, surround by asterisks for contains). Also if the operator is negated, then the `--filter-not-...` variant is used.

That way we are covering more cases without listing every single combination which hopefully should help LLMs.

We should also note that FullyQualifiedName is the default property when it's not explicitly written. I think `--filter "FQN"` is allowed, but I don't recall if VSTest treats it as "contains" or exact match.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+| `--filter "FullyQualifiedName!~Slow"` | `--filter-not-method Slow` |
+
+Multiple `--filter-*` options can be combined on the same command line. They are ANDed together:
```
Youssef1313: It was dependent on the exact option IIRC. Some stuff are ANDed and some other stuff are ORed, at least per my memory.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,396 @@
...
+<PackageReference Include="Microsoft.Testing.Extensions.TrxReport" Version="1.6.2" />
+
+<!-- Crash dump collection (replaces --blame-crash) -->
+<PackageReference Include="Microsoft.Testing.Extensions.CrashDump" Version="1.6.2" />
+
+<!-- Hang dump collection (replaces --blame-hang) -->
+<PackageReference Include="Microsoft.Testing.Extensions.HangDump" Version="1.6.2" />
+
+<!-- Code coverage (replaces --collect "Code Coverage") -->
+<PackageReference Include="Microsoft.Testing.Extensions.CodeCoverage" Version="17.13.0" />
```
Youssef1313, line 237: Should we also have instructions for coverlet, GitHubActionsTestLogger, spekt loggers for NUnit/JUnit/XUnit?

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — now reads:
- **MSTest**: References `MSTest` or `MSTest.TestAdapter`, or uses `MSTest.Sdk` (with `<IsTestApplication>` not set to `false`). Note: `MSTest.TestFramework` alone is a library dependency, not a test project.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — simplified to `- **NUnit**: References \`NUnit3TestAdapter\``

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — removed `OutputType Exe` from the Directory.Build.props template entirely. The skill now notes that the runner properties (`EnableMSTestRunner`, `EnableNUnitRunner`) automatically set `OutputType Exe` on projects that reference the corresponding test framework. For xUnit.net, `YTest.MTP.XUnit2` handles it automatically.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — removed the MTP definition sentence. The skill now just says "Migrate a .NET test solution from VSTest to Microsoft.Testing.Platform (MTP)." without the platform capabilities description.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — removed the `IsTestProject` condition block entirely. The runner properties handle `OutputType` automatically so the problem goes away.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Acknowledged — deferring the minimum safe version investigation for now. Will follow up separately.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — replaced the entire xunit v3 section. The skill now recommends `YTest.MTP.XUnit2` for xUnit.net v2 projects, which provides MTP support without requiring an upgrade to v3. xUnit v2→v3 migration will be a separate skill.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — changed to "forces the project to use VSTest instead of MTP".

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — applied the suggestion. Now reads: `Use the VSTest mode of \`dotnet test\` command to run MTP test projects by adding this property in \`Directory.Build.props\`:`

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — changed to "Applicable only to MSTest and NUnit".

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — removed along with the rest of the xunit filter section (now using `YTest.MTP.XUnit2` which preserves VSTest `--filter`).

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — entire xunit v3 section replaced. The skill now recommends `YTest.MTP.XUnit2` for xUnit.net v2. This comment about version is no longer applicable since the xunit.v3 PackageReference was removed.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — applied the suggestion. Now reads: `In this mode, \`dotnet test\` arguments are passed directly — for example, \`dotnet test --report-trx\`.`

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — added an explicit warning: `> **Important**: \`global.json\` does not support trailing commas. Ensure the JSON is strictly valid.`

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — the entire xunit-specific filter section has been removed. Since `YTest.MTP.XUnit2` preserves the VSTest `--filter` syntax, no filter migration is needed for xUnit.net. The filter section now just says: "**MSTest, NUnit, and xUnit.net (with `YTest.MTP.XUnit2`)**: The VSTest `--filter` syntax is identical on both VSTest and MTP. No changes needed."

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — removed the `testconfig.json` mention entirely. The line now just says: `MSTest and NUnit still support \`.runsettings\``

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — no longer applicable since the xunit filter section was removed (`YTest.MTP.XUnit2` preserves VSTest `--filter` syntax).

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: Fixed — the entire xunit filter section has been removed since `YTest.MTP.XUnit2` preserves the VSTest `--filter` syntax. No filter migration needed for any framework now.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Evangelink: I'll do a follow-up.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
Youssef1313: @Evangelink This is still not true. EnableMSTestRunner and EnableNUnitRunner don't automatically set OutputType Exe.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -0,0 +1,337 @@
...
+<!-- Directory.Build.props -->
+<Project>
+  <PropertyGroup>
+    <!-- Framework-specific runner property goes here (Step 3) -->
+    <!-- dotnet test integration property goes here (Step 4) -->
+  </PropertyGroup>
+</Project>
+```
+
+> **Note**: The framework-specific runner properties (`EnableMSTestRunner`, `EnableNUnitRunner`) automatically set `<OutputType>Exe</OutputType>` on projects that reference the corresponding test framework. You do not need to set `OutputType` manually. MSTest.Sdk also sets `EnableMSTestRunner` automatically. For xUnit.net, `YTest.MTP.XUnit2` handles this automatically.
```
Youssef1313: OutputType isn't set automatically by EnableMSTestRunner/EnableNUnitRunner. It's set automatically by MSTest.Sdk though.

Youssef1313: Also YTest.MTP.XUnit2 doesn't automatically set OutputType Exe and you need to manually set it.

plugins/dotnet-test/skills/migrate-vstest-to-mtp/SKILL.md
```diff
@@ -63,19 +57,7 @@ Migrate a .NET test solution from VSTest to Microsoft.Testing.Platform (MTP). Th
...
-<Project>
-  <PropertyGroup>
-    <!-- Framework-specific runner property goes here (Step 3) -->
-    <!-- dotnet test integration property goes here (Step 4) -->
-  </PropertyGroup>
-</Project>
-```
-
-> **Note**: The framework-specific runner properties (`EnableMSTestRunner`, `EnableNUnitRunner`) automatically set `<OutputType>Exe</OutputType>` on projects that reference the corresponding test framework. You do not need to set `OutputType` manually. MSTest.Sdk also sets `EnableMSTestRunner` automatically. For xUnit.net, `YTest.MTP.XUnit2` handles this automatically.
+> **Note**: MTP requires test projects to have `<OutputType>Exe</OutputType>`. Only `MSTest.Sdk` sets this automatically. For all other setups (MSTest NuGet packages with `EnableMSTestRunner`, NUnit with `EnableNUnitRunner`, xUnit.net with `YTest.MTP.XUnit2`), you must set `<OutputType>Exe</OutputType>` explicitly — either per-project or in `Directory.Build.props` with a condition that targets only test projects.
```
Youssef1313, line 60: Maybe it's of help that a condition in Directory.Build.props could potentially use `MSBuildProjectName` (as MSBuildProjectName is available very early). I think this might help LLM go into the right direction instead of attempting to do `Condition="$(IsTestProject)=='true'"` which won't work.

PR #378, opened 2026-03-16
Add writing-mstest-tests skill with eval scenarios

Evangelink:
Add a new skill for writing modern MSTest 3.x/4.x unit tests covering:
- Test class structure (sealed, AAA pattern, naming conventions)
- Modern assertion APIs (Assert.Throws, IsInstanceOfType, collection asserts)
- Data-driven tests (DataRow, DynamicData with ValueTuples)
- Async tests with cancellation and timeout patterns
- Test lifecycle (constructor init, TestContext injection)
- Anti-pattern detection and modernization

Includes 10 eval scenarios across 4 goal categories with fixture files for both modern MSTest and legacy anti-pattern test projects.

MSTest.Sdk version is recommended via global.json msbuild-sdks instead of inline in the project file, ensuring all test projects get bumped together.

### DIFF
--- a/eng/skill-validator/README.md
+++ b/eng/skill-validator/README.md
@@ -140,7 +140,7 @@ scenarios:
         - path: "input.txt"
           content: "file content to create before the run"
         - path: "data.csv"
-          source: "fixtures/sample-data.csv"  # relative to the eval.yaml directory (preferred); if no eval.yaml is available, relative to the skill directory
+          source: "fixtures/sample-data.csv"  # relative to skill dir
     assertions:
       - type: "output_contains"
         value: "expected text"


plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
```diff
@@ -0,0 +1,319 @@
+---
```
Youssef1313, line 1: nit: I personally prefer separating the skill validation changes in a different PR than the one introducing the skill

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Workflow
+
+### Step 1: Determine project setup
+
+Check the test project for MSTest version and configuration:
+
+- If using `MSTest.Sdk` (`<Sdk Name="MSTest.Sdk">`): modern setup, all features available
+- If using `MSTest` metapackage: modern setup (MSTest 3.x+)
+- If using `MSTest.TestFramework` + `MSTest.TestAdapter`: check version for feature availability
```
Youssef1313, line 39: I'm not sure what should be the takeaway from these.

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+    <PackageReference Include="MSTest" Version="3.8.2" />
+  </ItemGroup>
+</Project>
+```
+
+### Step 2: Write test classes following conventions
+
+Apply these structural conventions:
+
+- **Seal test classes** with `sealed` for performance and design clarity
```
Youssef1313, line 78: I don't think it will matter a lot for perf, if it matters at all. (and yes I'm aware of https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852). But it's also not bad to seal types when the intent is that they are not inheritable.

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+- Use separate test projects with naming convention `[ProjectName].Tests`
+
+```csharp
+[TestClass]
+public sealed class OrderServiceTests
+{
+    [TestMethod]
+    public void CalculateTotal_WithDiscount_ReturnsReducedPrice()
+    {
+        // Arrange
```
Youssef1313, line 91: I personally find it annoying to keep commenting on each phase/step of AAA.

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+#### Equality and null checks
+
+```csharp
+Assert.AreEqual(expected, actual);      // Value equality
+Assert.AreSame(expected, actual);       // Reference equality
+Assert.IsNull(value);
+Assert.IsNotNull(value);
+```
+
+#### Exception testing — use `Assert.Throws` instead of `[ExpectedException]`
```
Youssef1313, line 117: I'm not sure whether this helps or makes things more confusing (e.g, when using older versions of MSTest - we are guiding LLM to the wrong direction).

What I have in mind would be:

- Step 0: Create test project using latest version if a project doesn't already exist.
- Step 1: Determins MSTest version. If project was created in Step 0, then it's latest version already. Otherwise, the version is whatever version is already in use.
- Next steps are all kinda "branching" depending on the version determined in Step 1

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
Evangelink: Line below: `Recommend MSTest.Sdk or the MSTest metapackage for new projects`

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
Evangelink: I think it's better to always go close -> open, so start sealed and open if needs be.

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
Evangelink: For humans, I can understand, for LLM, I think it's cheap and it's a known pattern so I'll ignore this comment.

plugins/dotnet-test/skills/writing-mstest-tests/SKILL.md
Evangelink: LLM still tries to use `[ExpectedException]` so this is guiding it toward proper patterns. I would prefer to split skills as they grow/complexify rather than from start.

CONTRIBUTING.md
```diff
@@ -255,7 +255,7 @@ setup:
     - path: "input.txt"
       content: "inline file content"
     - path: "data.csv"
-      source: "fixtures/sample-data.csv"  # relative to skill directory
+      source: "fixtures/sample-data.csv"  # relative to eval directory, then skill directory
```
Evangelink:
```suggestion
      source: "fixtures/sample-data.csv"  # relative to skill directory
```

eng/skill-validator/README.md
```diff
@@ -140,7 +140,7 @@ scenarios:
         - path: "input.txt"
           content: "file content to create before the run"
         - path: "data.csv"
-          source: "fixtures/sample-data.csv"  # relative to skill dir
+          source: "fixtures/sample-data.csv"  # relative to the eval.yaml directory (preferred); if no eval.yaml is available, relative to the skill directory
```
Evangelink:
```suggestion
          source: "fixtures/sample-data.csv"  # relative to skill dir
```

PR #446, opened 2026-03-25
Create experimental plugin

Evangelink:


.github/CODEOWNERS
```diff
@@ -73,9 +73,9 @@
 /plugins/dotnet-test/ @dotnet/dotnet-testing
 /tests/dotnet-test/ @dotnet/dotnet-testing
 
-# dotnet-test-experimental (experimental test quality skills)
-/plugins/dotnet-test-experimental/ @dotnet/dotnet-testing
-/tests/dotnet-test-experimental/ @dotnet/dotnet-testing
+# dotnet-experimental (skills under active evaluation)
+/plugins/dotnet-experimental/ @dotnet/dotnet-testing
+/tests/dotnet-experimental/ @dotnet/dotnet-testing
```
ViktorHofer: Are you sure that this alias should be the owner of all skills in dotnet-experimental? You might want to only own that one skill that is currently in it.

Evangelink: Oh sorry nope it should be infra folks, thanks for catching that!

.github/CODEOWNERS
```diff
@@ -73,9 +73,9 @@
 /plugins/dotnet-test/ @dotnet/dotnet-testing
 /tests/dotnet-test/ @dotnet/dotnet-testing
 
-# dotnet-test-experimental (experimental test quality skills)
-/plugins/dotnet-test-experimental/ @dotnet/dotnet-testing
-/tests/dotnet-test-experimental/ @dotnet/dotnet-testing
+# dotnet-experimental (skills under active evaluation)
+/plugins/dotnet-experimental/ @JanKrivanek @ViktorHofer @Evangelink @ManishJayaswal @AbhitejJohn
+/tests/dotnet-experimental/ @JanKrivanek @ViktorHofer @Evangelink @ManishJayaswal @AbhitejJohn
```
ViktorHofer: Why do we set global owners for the entire plugin? I think what we discussed yesterday was that we will just use the same model as for non experimental skills. Meaning that when someone adds an experimental skill, they need to also own it.

ViktorHofer (approved):

PR #451, opened 2026-03-25
Improve investigation guide: PR number in prompt, document reject_tools, baseline-already-good pattern

danmoseley:
Improvements to `InvestigatingResults.md` and the evaluation workflow, based on experience using the guide to diagnose a real eval failure (PR #264).

## Changes

- **Add PR number to the copy-paste investigation prompt** in eval comments, so the agent knows which PR to analyze and where to post findings
- **Fix `improvementScore` field reference** to `isolatedImprovementScore` — the bare field name doesn't exist in the JSON schema
- **Document `reject_tools`, `timeout`, and `setup.files`** eval.yaml options in a new reference section — `reject_tools` was mentioned in a fix but never explained
- **Add failure pattern #8: "Baseline already good (no headroom)"** — covers the case where the model scores 4.5-5.0/5 without the skill, leaving no room for improvement
- **Add overfitting analysis** to the agent investigation flow (step 9) — the `overfittingResult` data was already being produced but the guide didn't explain how to use it
- **Add note about saving the example script as a `.py` file** rather than running via `python -c` (nested quote escaping issues)
- **Update priority list** to include the new pattern

### DIFF
--- a/eng/skill-validator/InvestigatingResults.md
+++ b/eng/skill-validator/InvestigatingResults.md
@@ -255,6 +255,32 @@ Most failing scenarios match 2–3 patterns simultaneously (e.g., timeout + toke
 6. **Rubric/judgment issues (#6, #7)** — once the runs are stable, tune the rubric.
 7. **Token overhead (#4)** — only optimize if quality is already good but the weighted score is marginally negative.
 
+## Improving the skill vs. gaming the eval
+
+When investigating failures, the goal is to **make the skill more useful to users** — not simply to make the eval score go up. Score improvement should be *evidence* of a better skill, not an end in itself.
+
+### Legitimate fixes (improve the skill)
+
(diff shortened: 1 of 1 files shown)


.github/workflows/evaluation.yml
```diff
@@ -155,7 +156,7 @@ jobs:
             Select-Object -First 1
 
           $hasInfraChanges = $changedFiles |
-            Where-Object { $_ -match '^eng/skill-validator/' } |
+            Where-Object { $_ -match '^eng/skill-validator/' -and $_ -ne 'eng/skill-validator/InvestigatingResults.md' } |
```
danmoseley, line 159: I'm determined to make it so I don't have to evlauate everything (or bypass) when touching just this md file. Maybe I should have put it somewhere else.

PR #535, opened 2026-04-17
add a new dotnet11 plugin and a new system text json skill to it.

ManishJayaswal:
Adding first of the several new skills for .NET11. These will cover the new/updated  APIs in .NET 11

### DIFF
--- a/.github/plugin/marketplace.json
+++ b/.github/plugin/marketplace.json
@@ -58,6 +58,11 @@
       "name": "dotnet-aspnet",
       "source": "./plugins/dotnet-aspnet",
       "description": "ASP.NET Core web development skills including middleware, endpoints, real-time communication, and API patterns."
+    },
+    {
+      "name": "dotnet11",
+      "source": "./plugins/dotnet11",
+      "description": "Skills for new .NET 11 APIs and language features."
     }
   ]
 }
--- a/eng/skill-validator/src/Evaluate/EvaluateCommand.cs
+++ b/eng/skill-validator/src/Evaluate/EvaluateCommand.cs
@@ -33,6 +33,7 @@ public static class EvaluateCommand
         var noiseSkillsDirOpt = new Option<string?>("--noise-skills-dir") { Description = "Directory containing skills to load as noise. Enables the noise test: re-runs scenarios with all noise skills loaded and measures degradation." };
         var noiseMaxDegradationOpt = new Option<double>("--noise-max-degradation") { Description = "Maximum acceptable average quality degradation (0-1) in noise test (only positive degradations count)", DefaultValueFactory = _ => 0.2 };
         var noiseMaxScenarioDegradationOpt = new Option<double>("--noise-max-scenario-degradation") { Description = "Maximum acceptable quality degradation (0-1) for any single noise-test scenario", DefaultValueFactory = _ => 0.4 };
+        var includeDotnet11Opt = new Option<bool>("--include-dotnet11") { Description = "Include skills in the dotnet11 plugin (excluded by default because it requires .NET 11 preview SDK)" };
 
         var command = new Command("evaluate", "Evaluate agent skills via LLM-based testing")
         {
@@ -59,6 +60,7 @@ public static class EvaluateCommand
             noiseSkillsDirOpt,
             noiseMaxDegradationOpt,
             noiseMaxScenarioDegradationOpt,
+            includeDotnet11Opt,
         };
 
         command.Add(RejudgeCommand.Create());
@@ -110,6 +112,7 @@ public static class EvaluateCommand
                 NoiseSkillsDir = parseResult.GetValue(noiseSkillsDirOpt),
                 NoiseDegradationLimit = parseResult.GetValue(noiseMaxDegradationOpt),
                 NoiseMaxScenarioDegradation = parseResult.GetValue(noiseMaxScenarioDegradationOpt),
+                IncludeDotnet11 = parseResult.GetValue(includeDotnet11Opt),
             };
 
             return await Run(config, cancellationToken);
@@ -198,6 +201,24 @@ public static class EvaluateCommand
             var searched = string.Join(", ", config.SkillPaths.Select(p => $"\"{Path.GetFullPath(p)}\""));
             Console.Error.WriteLine($"No skills or agents found in the specified paths: {searched}");
             return 1;
+        }
+
+        // Filter out dotnet11 plugin skills unless --include-dotnet11 is passed
+        if (!config.IncludeDotnet11)
+        {
+            var excluded = discoveredSkills.Where(s => IsDotnet11Plugin(s)).ToList();
+            if (excluded.Count > 0)
+            {
+                discoveredSkills = discoveredSkills.Where(s => !IsDotnet11Plugin(s)).ToList();
+                Console.WriteLine($"Skipping {excluded.Count} dotnet11 skill(s) (pass --include-dotnet11 to include them)");
+
+                if (discoveredSkills.Count == 0 && discoveredAgents.Count == 0)
+                {
+                    Console.Error.WriteLine("All discovered skills belong to the dotnet11 plugin which requires .NET 11 preview SDK. " +
+                        "Pass --include-dotnet11 to evaluate them.");
+                    return 1;
+                }
+            }
         }
 
         if (discoveredSkills.Count > 0)
@@ -1832,6 +1853,13 @@ public static class EvaluateCommand
         }
 
         return resolved.Count > 0 ? resolved : null;
+    }
+
+    private static bool IsDotnet11Plugin(SkillInfo skill)
+    {
+        var context = PluginDiscovery.FindPluginContext(skill);
+        return context is not null &&
+               string.Equals(context.Value.Plugin.Name, "dotnet11", StringComparison.OrdinalIgnoreCase);
     }
 
     internal static (Dictionary<string, (PluginInfo Plugin, List<SkillInfo> Skills)> Groups, List<string> Errors)
--- a/eng/skill-validator/src/Evaluate/EvaluateCommand.cs
+++ b/eng/skill-validator/src/Evaluate/EvaluateCommand.cs
@@ -206,13 +206,13 @@ public static class EvaluateCommand
         // Filter out dotnet11 plugin skills and agents unless --include-dotnet11 is passed
         if (!config.IncludeDotnet11)
         {
-            var excludedSkills = discoveredSkills.Where(s => IsDotnet11Plugin(s)).ToList();
-            var excludedAgents = discoveredAgents.Where(a => IsDotnet11PluginPath(a.Path)).ToList();
+            var excludedSkills = new HashSet<SkillInfo>(discoveredSkills.Where(s => IsDotnet11Plugin(s)));
+            var excludedAgents = new HashSet<AgentInfo>(discoveredAgents.Where(a => IsDotnet11PluginPath(a.Path)));
             var excludedCount = excludedSkills.Count + excludedAgents.Count;
             if (excludedCount > 0)
             {
-                discoveredSkills = discoveredSkills.Where(s => !IsDotnet11Plugin(s)).ToList();
-                discoveredAgents = discoveredAgents.Where(a => !IsDotnet11PluginPath(a.Path)).ToList();
+                discoveredSkills = discoveredSkills.Where(s => !excludedSkills.Contains(s)).ToList();
+                discoveredAgents = discoveredAgents.Where(a => !excludedAgents.Contains(a)).ToList();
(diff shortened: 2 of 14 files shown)


ManishJayaswal:
@sayedihashimi - FYI

ManishJayaswal:
@ViktorHofer  and @JanKrivanek - these skills need .NET 11 installed. However, it does not seem like validator actually installs any .NET or checks if it is installed and what version it is. Is the assumption here that latest .NET will always be available on the machine? I would prefer these evaluations to not run if .NET 11 is not installed but there should be a way to install latest preview of .NET11 through some option. Thoughts?

ViktorHofer:
Basically conditional evals. I think someone added trait support to eval.yaml but I would need to double check.

Today, in CI the evaluation.yml file installs the .NET SDK: https://github.com/dotnet/skills/blob/400661df09e807167bc5cd0509c4e0268f2af3d9/.github/workflows/evaluation.yml#L537-L540

The simplest solution might be just updating to a .NET 11 SDK

SamMonoRT:
@eiriktsarpalis - please review the STJ skill

ViktorHofer (commented):
Btw we already have a `dotnet-upgrade` plugin. Maybe that's better suited instead of the separate plugin?

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Typed `JsonTypeInfo` Access
+
+#### `JsonSerializerOptions.GetTypeInfo<T>()`
+
+Returns a strongly-typed `JsonTypeInfo<T>` for the specified type, using the
+options' configured type-info resolver.
+
+```csharp
+JsonTypeInfo<T> GetTypeInfo<T>()
```
eiriktsarpalis, line 45: Why not inline examples into each section? This alone doesn't seem to provide a whole lot of context on its own?

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
```diff
@@ -0,0 +1,117 @@
+---
+name: system-text-json-net11
+description: >
+  Provides guidance on new System.Text.Json APIs introduced in .NET 11.
```
eiriktsarpalis, line 4: What's the purpose of a skill that enumerates new features in a particular release of a library? If it's to upgrade specific coding patterns to take advantage of new APIs, these coding patterns should be enumerated explicitly, e.g. with *BEFORE* and *AFTER* sections.

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+  It covers typed JsonTypeInfo access via GetTypeInfo<T> and TryGetTypeInfo<T> on
+  JsonSerializerOptions, and the new JsonNamingPolicy.PascalCase static property.
+  Use when serializing or deserializing JSON in .NET 11 applications and needing
+  typed metadata access or PascalCase property naming.
+---
+
+# System.Text.Json — .NET 11
+
+New APIs added to `System.Text.Json` across .NET 11 releases.
```
eiriktsarpalis, line 14: Do note that there are whole load of STJ features still in the pipeline, so this document would need to be kept in sync.

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
ManishJayaswal: the purpose is mostly to introduce new features to the model (something that model was not trained on), not necessarily to upgrade.

eiriktsarpalis: Right, in that case wouldn't it make sense to exactly document what the new features are meant to solve? I don't see how simply enumerating new APIs will add much to the context.

ManishJayaswal: There have been several experiments that have shown that just listing out the new APIs and some examples on how to use them leads to significantly better results. Our own results and experiments with NuGet team shows that. Here is a blog post from google on this - https://developers.googleblog.com/en/closing-the-knowledge-gap-with-agent-skills/

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
ManishJayaswal: Yes, it is expected that we will update these before final release of .NET 11

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
ManishJayaswal: I played with inlining the examples as well as putting them in the end and the results were similar. I had decided to put it in the end so that, where needed, we can cover multiple APIs in the same example (to reduce total text).

tests/dotnet11/system-text-json-net11/eval.yaml
```diff
@@ -0,0 +1,102 @@
...
+      camelCase property names:
+
+      ```csharp
+      public record Person(string Name, int Age);
+      ```
+
+      Show me a minimal program targeting `net8.0` that produces output like
+      `{"name":"Jane","age":30}` and run it.
+    expect_activation: false
+    assertions:
```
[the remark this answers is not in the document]
ManishJayaswal: Rubric signal says to check that it does not load the system-text-json-net11 skill

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Target Framework
+
+```xml
+<TargetFramework>net11.0</TargetFramework>
+```
+
+## New APIs
```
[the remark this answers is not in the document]
ManishJayaswal: Fixed

eng/skill-validator/src/Evaluate/EvaluateCommand.cs
```diff
@@ -200,6 +203,24 @@ public static async Task<int> Run(ValidatorConfig config, CancellationToken canc
             return 1;
         }
 
+        // Filter out dotnet11 plugin skills unless --include-dotnet11 is passed
+        if (!config.IncludeDotnet11)
+        {
+            var excluded = discoveredSkills.Where(s => IsDotnet11Plugin(s)).ToList();
+            if (excluded.Count > 0)
+            {
+                discoveredSkills = discoveredSkills.Where(s => !IsDotnet11Plugin(s)).ToList();
```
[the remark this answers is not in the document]
ManishJayaswal: Fixed

eng/skill-validator/src/Evaluate/EvaluateCommand.cs
```diff
@@ -200,6 +203,27 @@ public static async Task<int> Run(ValidatorConfig config, CancellationToken canc
...
+        // Filter out dotnet11 plugin skills and agents unless --include-dotnet11 is passed
+        if (!config.IncludeDotnet11)
+        {
+            var excludedSkills = discoveredSkills.Where(s => IsDotnet11Plugin(s)).ToList();
+            var excludedAgents = discoveredAgents.Where(a => IsDotnet11PluginPath(a.Path)).ToList();
+            var excludedCount = excludedSkills.Count + excludedAgents.Count;
+            if (excludedCount > 0)
+            {
+                discoveredSkills = discoveredSkills.Where(s => !IsDotnet11Plugin(s)).ToList();
+                discoveredAgents = discoveredAgents.Where(a => !IsDotnet11PluginPath(a.Path)).ToList();
```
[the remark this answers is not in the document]
ManishJayaswal: Fixed.

ManishJayaswal:
@JanKrivanek and @ViktorHofer  - it seems like I need to merge the change in evaluation.yml before I can evaluate the new skill in the new plugin dotnet11.  

Why is this is happening -  I set the default in /evaluate to not run skills in dotnet11 plugins as they need .NET 11 preview installed. Otherwise, every run will need to install .NET 11 plus .NET 10 (unless we want to test only on .NET 11 previews, which is not ideal). They are only run when --dotnet11 flag is passed to /evaluate command. However, this is now a chicken and egg problem. I need to first merge evaluation.yml (which is included in this PR) but can't really evaluate this PR without that change as the workflow picks the evaluation.yml from main branch. 

Any concerns if we merge this one dotnet11 skill without evaluation as we have 7 more dotnet11 skills that we need to merge?

@AbhitejJohn - FYI

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Serializing or deserializing JSON in a .NET 11 (or later) project
+- Needing strongly-typed `JsonTypeInfo<T>` access instead of the untyped `JsonTypeInfo` overload
+- Wanting to safely check whether type metadata is available without catching exceptions (`TryGetTypeInfo<T>`)
+- Requiring PascalCase property naming during JSON serialization
+
+## When Not to Use
+
+- The project targets .NET 10 or earlier — these APIs are not available before .NET 11
+- Using a third-party JSON library (e.g., Newtonsoft.Json) instead of `System.Text.Json`
```
AbhitejJohn: Consider: I don't know if the model understands third-party in this context.

```suggestion
- Using a JSON library that is not `System.Text.Json` (e.g., Newtonsoft.Json)
```

AbhitejJohn (commented):
I'm a little confused by the dotnet11 flag for eval. What is our concern with always installing the dotnet 11 SDK? If our concern is that might affect the other evals then, can we segment that to a different pipeline perhaps? The way we have it today, its easy to regress these skills cause they aren't validated in CI by default.

ViktorHofer:
Yeah we should just update the SDK in this repo to .NET 11 and use that.

ViktorHofer (changes requested):
Please revert the changes in skill-validator. We shouldn't put dotnet version specific logic into that tool. This seems to be a specific demand to this repository. If we want to being able to filter on specific evals and only run those, then a trait system might be better. We might even already have that, not 100% sure.

ManishJayaswal:
@ViktorHofer  and @AbhitejJohn - since most customers would run these skills on .NET 10 (and very few on .NET 11 previews at this time), I was thinking that other skills should continue to be evaluated on .NET 10. We can certainly change SDK in this repo to .NET 11 previews but that is not what most customers would have.

ViktorHofer:
IMO the majority of our customer base isn't even on .NET 10 yet. They usually stick to the earliest in support LTS version (.NET 8). I think our key audience for these skills should be customers using bleeding edge technologies like .NET 11 Previews and up-to-date models like opus4.6 / opus4.7 / gpt5.x.

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
ManishJayaswal: fixed

ManishJayaswal:
Thanks @JanKrivanek for making .NET11 previews the default SDK. I have updated this PR to remove all changes to skill-validator and other changes related to the .NET 11 SDK. Now it is simply adding a new plugin (dotnet11) with one skill in it. @AbhitejJohn @JanKrivanek and @ViktorHofer - if you folks can approve it then I will merge it and then add the rest of dotnet11 skills to this plugin in separate PRs (so that they can be reviewed)

plugins/dotnet11/skills/system-text-json-net11/SKILL.md
```diff
@@ -0,0 +1,130 @@
+---
+name: system-text-json-net11
+description: >
+  Provides guidance on new System.Text.Json APIs introduced in .NET 11.
+  It covers typed JsonTypeInfo access via GetTypeInfo<T> and TryGetTypeInfo<T> on
```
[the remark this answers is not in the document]
ManishJayaswal: fixed

.github/plugin/marketplace.json
```diff
@@ -58,6 +58,11 @@
       "name": "dotnet-aspnet",
       "source": "./plugins/dotnet-aspnet",
       "description": "ASP.NET Core web development skills including middleware, endpoints, real-time communication, and API patterns."
+    },
+    {
+      "name": "dotnet11",
+      "source": "./plugins/dotnet11",
+      "description": "Skills for new .NET 11 APIs and language features."
     }
```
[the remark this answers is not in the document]
ManishJayaswal: fixed

ManishJayaswal:
> Please revert the changes in skill-validator. We shouldn't put dotnet version specific logic into that tool. This seems to be a specific demand to this repository. If we want to being able to filter on specific evals and only run those, then a trait system might be better. We might even already have that, not 100% sure.

@ViktorHofer - the requested change has been made. Can you approve this for merging?
@AbhitejJohn  - FYI

AbhitejJohn (approved):

ViktorHofer (approved):

JanKrivanek (approved):

PR #577, opened 2026-04-22
Fold F# .fsproj ordering into msbuild-antipatterns AP-05

T-Gro:
Replaces #248.

AP-05 tells users to drop explicit `<Compile Include>` lists — correct for C#, but silently wrong for any order-dependent compiler (F#, and therefore any mixed-language solution). A two-sentence exception in AP-05 (plus ported eval scenarios) covers it inline.


ViktorHofer:
@T-Gro I just removed the agentic workflow for msbuild. You can reset your changes to that file.

T-Gro:
ok, done

ViktorHofer (approved):

PR #579, opened 2026-04-23
Remove agentic workflows

ViktorHofer:
The agentic workflows were useful when CCA didn't support loading skills. That's now solved and supported. The agentic workflows come with a couple of issues (limited chars) which make it hard to use. Remove them in favor of the plugin system.


JanKrivanek (approved):
🥹

PR #629, opened 2026-05-07
Fix markdownlint CI workflow (startup_failure since inception)

danmoseley:
## Problem

The markdownlint CI workflow has **never worked** — all 117 runs since it was added in PR #534 (April 17) show `startup_failure`. The root cause is that it uses third-party actions (`tj-actions/changed-files` and `DavidAnson/markdownlint-cli2-action`) that are not in the dotnet org's allowed actions list.

### Workflow changes
Replace third-party actions with inline steps using only first-party actions already in the repo's allowlist:
- `actions/checkout` + `actions/setup-node` (both already approved, SHA-pinned)
- `git diff` for changed file detection (replaces `tj-actions/changed-files`)
- `npx markdownlint-cli2` for linting (replaces `DavidAnson/markdownlint-cli2-action`)
- When config or workflow files change, lint **all** `.md` files (not just changed ones)

This follows the same pattern used by `dotnet/runtime`'s working markdownlint workflow.

### Config changes
Suppress rules in `.markdownlint-cli2.jsonc` that conflict with established repo conventions:
- **MD001** (heading increment), **MD007** (list indentation), **MD014** (dollar signs in commands)
- **MD022** (blanks around headings), **MD026** (trailing punctuation in headings)
- **MD028** (blank lines in blockquotes), **MD029** (ordered list prefix style)
- **MD031** (blanks around fences), **MD032** (blanks around lists)
- **MD034** (bare URLs), **MD036** (emphasis as heading)
- **MD040** (code fence language), **MD056** (table column count)
- **MD058** (blanks around tables), **MD060** (table column style)

### Mechanical fixes across 9 files
- **MD009**: removed trailing whitespace (4 files)
- **MD012**: reduced consecutive blank lines (3 files)
- **MD038**: fixed space in code span (1 file)
- **MD047**: added missing trailing newline (3 files)

## Validation
- Ran `markdownlint-cli2` locally against **all** `.md` files in the repo: **0 errors**
- Both `actions/checkout` and `actions/setup-node` SHAs are already in the repo's GitHub Actions allowlist

PR #632, opened 2026-05-08
Fix markdownlint workflow: recursive pathspec and npx --yes

danmoseley:
Follow-up to #629. Addresses two review feedback items:

1. **Recursive pathspec**: Changed \-- '*.md'\ to \-- '**/*.md'\ in \git diff\ so markdown files in subdirectories are detected (not just root-level files).

2. **npx --yes flag**: Added \--yes\ to \
px markdownlint-cli2\ calls to avoid interactive install prompts in CI.


danmoseley:
Closing: both changes are unnecessary. git pathspec '*.md' already matches recursively (confirmed locally), and npx in CI (non-interactive) installs packages without prompting regardless of --yes flag.

PR #633, opened 2026-05-11
fix(groom): use MCP tools with min-integrity:none instead of unauthenticated gh CLI

JanKrivanek:
## Problem

PR #627 instructed the agent to use `gh api` for reading issue comments. However:

1. **`gh` CLI is NOT authenticated in the sandbox** — the framework strips credentials via the `Clean git credentials` step, and the system prompt explicitly says *"`gh` is not authenticated"*
2. Run 25492920921 (post-fix) confirmed: the agent tried `gh api`, found it unauthenticated, and called `missing_tool` — no grooming was performed

### Root Cause

Analysis of run 25481332080 (pre-fix, MCP-based) shows the MCP `issue_read` tool **did return comments** — only 1 item out of many was filtered by the integrity policy. The agent overreacted to the `[Filtered]` warning and called `report_incomplete` instead of working with the available data.

## Fix

| Change | Reason |
|--------|--------|
| Set `min-integrity: none` under `tools.github` | Disables integrity filtering. Safe because Step 2 only processes `github-actions[bot]` comments (author filter) |
| Revert to MCP `issue_read` with `get_comments` | The proper tool for reading comments inside the sandbox |
| Remove `gh` from bash tools | Not usable without auth |
| Add filtered-items guidance | Tells agent to continue with available data if `[Filtered]` appears |
| Document sandbox auth limitation | Prevents future attempts to use `gh` CLI for API calls |

Fixes #625


JanKrivanek:
@ViktorHofer - Would appreciate a quick review. The groom workflow has been broken since May 8 (PR #627 uses \gh api\ which doesn't work in the agentic sandbox — credentials are stripped). This fix reverts to MCP tools with \min-integrity: none\ to bypass the integrity filter. Auto-merge is enabled — just needs approval to unblock.

Evangelink (approved):

ViktorHofer (commented):
LGTM

PR #644, opened 2026-05-13
Add LSP support for .razor and .cshtml file extensions

davidwengier:
The roslyn-language-server package includes the Razor LSP bits now.

Fixes https://github.com/dotnet/razor/issues/13140

Package that adds this support is: https://www.nuget.org/packages/roslyn-language-server/5.8.0-1.26262.10

### DIFF
--- a/plugins/dotnet/lsp.json
+++ b/plugins/dotnet/lsp.json
@@ -12,7 +12,9 @@
         "--autoLoadProjects"
       ],
       "fileExtensions": {
-        ".cs": "csharp"
+        ".cs": "csharp",
+        ".razor": "aspnetcorerazor",
+        ".cshtml": "aspnetcorerazor"
       }
     }
   }


dbreshears (approved):

dibarbet (approved):

plugins/dotnet/lsp.json
```diff
@@ -12,7 +12,9 @@
         "--autoLoadProjects"
       ],
       "fileExtensions": {
-        ".cs": "csharp"
+        ".cs": "csharp",
```
jasonmalinowski, line 15: I could have imagined this was somehow mapping it back to the "csharp" key that's on line 3? That's not the case?

jasonmalinowski (approved):

plugins/dotnet/lsp.json
davidwengier: Line 3 is just the name of the LSP server, and seems to have no bearing on functionality. In testing mine was called "localcsharp" so I could tell which server the cli was using, and everything worked fine.

PR #661, opened 2026-05-18
Fix vally eval isolation: pin baseline to 0 skills, skilled to 1

Evangelink:
## Problem

vally evaluations were not isolating skills properly. In the failing run https://github.com/dotnet/skills/actions/runs/26021981688, the `dotnet-test` job's eval for `code-testing-agent` reported **`Found 77 skill(s)`** in the baseline run and **`Found 22 skill(s)`** in the skilled run — instead of the expected 0 and 1.

## Root cause

`vally eval` derives its skill set via `discoverSkills(skillRoot, ...)` where `skillRoot = opts.skillDir ?? cwd`:

- The **baseline** call omitted `--skill-dir`, so vally fell back to the repo root, walked up to `.vally.yaml`, and loaded every skill under `plugins/` (77 total).
- The CI workflow's **skilled** call passed `--skill-dir ./plugins/<plugin>/skills` (from `matrix.entry.skills_path`, which for scheduled/infra runs is plugin-level). That directory has no `SKILL.md` at its root, so vally's Step-4 child scan picked up every sibling skill in the plugin (22 for `dotnet-test`).

## Fix

Applied to both `eng/vally-adapter/run-vally-evals.sh` and `.github/workflows/vally-evaluation.yml`:

- **Baseline** now passes `--skill-dir <empty mktemp dir>` so vally discovers 0 skills.
- **Skilled** now uses `plugins/<plugin>/skills/<eval-name>` (the specific skill dir, not the plugin-level one) so vally matches Step-1 (`SKILL.md` at root) and discovers exactly 1 skill.
- Evals with no matching `SKILL.md` — i.e. the `agent.*` orchestrator evals that declare their own `environment.skills` — are explicitly **skipped** with a warning. Baseline-vs-skilled isolation is not meaningful for them since they require multi-skill setup by design.
- The CI workflow's `Find eval specs` step now narrows to a single eval when `workflow_dispatch` is invoked with a specific `skill` input, so single-skill dispatches don't accidentally run every eval in the plugin.
- Empty temp dirs are created per-eval and cleaned up after each run to avoid any cross-run contention under parallel execution.

## Expected outcome

After this PR, CI logs should show:

- `--- Baseline run ---` → `Found 0 skill(s)`
- `--- Skilled run ---` → `Found 1 skill(s): <skill-name>`

### DIFF
--- a/.github/workflows/vally-evaluation.yml
+++ b/.github/workflows/vally-evaluation.yml
@@ -112,8 +112,22 @@ jobs:
 
       - name: Find eval specs
         id: find-evals
+        env:
+          DISPATCH_SKILL: ${{ inputs.skill }}
         run: |
-          EVALS=$(find tests/${{ matrix.entry.plugin }} -name "eval.vally.yaml" -type f | sort)
+          # For workflow_dispatch targeting a specific skill, only run that one
+          # eval. Otherwise run every eval discovered under the plugin's tests
+          # directory.
+          if [ -n "$DISPATCH_SKILL" ]; then
+            CANDIDATE="tests/${{ matrix.entry.plugin }}/${DISPATCH_SKILL}/eval.vally.yaml"
+            if [ -f "$CANDIDATE" ]; then
+              EVALS="$CANDIDATE"
+            else
+              EVALS=""
+            fi
+          else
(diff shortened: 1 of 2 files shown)


Evangelink:
Superseded by a PR opened from the dotnet/skills branch directly (not from fork).

PR #664, opened 2026-05-18
skill-validator: remove the tools check

Evangelink:
Per Jan's suggestion on #660, this PR removes the tools check entirely instead of working around its brittle hardcoded `BuiltInTools` list.

## Why

The `ExternalDependencyChecker` carried a hand-maintained `BuiltInTools` set used to flag:

1. Non-built-in tool names in an agent's frontmatter `tools:` array (e.g. `terminal` → warning).
2. `#tool:foo/bar` references in skill / agent prose.

Both are brittle and low-value:

- The list goes stale the moment the agent runtime adds a tool (`terminal` was the trigger for #660 — four agents got warned for a tool that is perfectly valid).
- The runtime itself enforces which tools an agent can use; the validator does not need to second-guess it.
- Authors had to babysit `tool-ref:` / `agent-tool:` entries in `allowed-external-deps.txt` for legitimate uses, which is pure busywork.

## What changed

- Removed `BuiltInTools`, `ToolReferenceRegex`, and `ExternalDependencyChecker.CheckAgent`.
- Removed the agent loop in `CheckCommand.CheckExternalDeps`.
- Removed obsolete tests; `LoadAllowList_ParsesEntriesAndSkipsComments` now uses an `mcp-server:` example.
- Removed the `tool-ref:` / `agent-tool:` documentation and the two `tool-ref:msbuild:#tool:...` allowlist entries from `eng/allowed-external-deps.txt`.

The `script:` / `invokes:` / `mcp-server:` checks all stay — they don't depend on a hardcoded tool list and still catch real external dependencies.

## Verification

- `dotnet build` clean.
- `dotnet test` — all 533 tests pass.
- `check --plugin plugins/dotnet-test` and `check --plugin plugins/dotnet-msbuild` both pass cleanly with the existing allowlist; no more `terminal` warnings (making #660 unnecessary).

Supersedes #663 (which was opened from a fork and triggered the infra-from-fork bot).

Also addresses the two Copilot review comments from #663:

- Updated the `ExternalDependencyChecker` class-level XML doc to no longer claim it covers agents.
- Removed the now-unused `agents` parameter from `CheckCommand.CheckExternalDeps` (and updated the single call site).

### DIFF
--- a/eng/allowed-external-deps.txt
+++ b/eng/allowed-external-deps.txt
@@ -5,8 +5,6 @@
 # Types:
 #   script:SKILL-NAME:relative/path    - Script file in a skill's scripts/ directory
 #   invokes:SKILL-NAME                 - INVOKES pattern in a skill description
-#   tool-ref:NAME:#tool:reference      - Non-built-in #tool: reference in skill/agent content
-#   agent-tool:AGENT-NAME:tool-name    - Non-built-in tool in agent frontmatter tools array
 #   mcp-server:PLUGIN-NAME:server-name - MCP server declaration in plugin.json
 #
 # To get clean, remove entries as the underlying dependency is resolved.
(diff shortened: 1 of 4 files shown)


Evangelink:
There was a confusion about the scope of what we wanted to achieve with @JanKrivanek

PR #670, opened 2026-05-19
test-anti-patterns / test-smell-detection: activation-only rewrite (Variant A)

Evangelink:
## What

**Variant A** of a 3-PR experiment to fix plugin-mode activation failures for 	est-anti-patterns and 	est-smell-detection. Activation-only rewrite — both skills kept, descriptions reworked.

### Diagnosis

Inspecting prior /evaluate result JSON for PR #653 showed:

- **test-anti-patterns plugin mode**: 5/6 scenarios → NO skill invoked at all (model reads files with iew and answers from training). 1/6 → picks `assertion-quality` instead.
- **test-smell-detection plugin mode**: 2/3 expected-activation scenarios → NO skill invoked. 1/3 → picks `test-anti-patterns` instead.

Both skills activate fine in isolated mode. The failure is plugin-mode discrimination: the model has many overlapping siblings, the baseline (no skill) already scores 4.0/5, and the previous descriptions actively told the agent `DO NOT USE FOR ... use the other one` — which trained it to bypass both.

## Change (Variant A)

* Keep both skills (least radical).
* Drop every cross-skill mutual-exclusion clause from both descriptions.
* Lead with `INVOKE THIS SKILL when…` mandate phrasing.
* Expand USE-FOR triggers with exact eval-prompt phrasing.
* Shrink `DO NOT USE FOR` to real boundaries (writing / running / migrating).
* Pull in the eval-prompt tweaks from PR #653 so prompts use the new trigger phrasing.

## Companion variants

* **Variant B** (#TBD-narrow-specialist): repositions `test-smell-detection` as a narrow niche only fired by explicit `testsmells.org` / academic-19-smell-catalog requests; `test-anti-patterns` becomes the umbrella.
* **Variant C** (#TBD-merge): deletes `test-smell-detection` entirely; moves its catalog under `test-anti-patterns/references/`; ports its eval scenarios.

Compare `/evaluate` results across all three to pick the winner.

## Validation

* `skill-validator check --plugin plugins/dotnet-test` passes for both skills.

### DIFF
--- a/plugins/dotnet-test/skills/test-anti-patterns/SKILL.md
+++ b/plugins/dotnet-test/skills/test-anti-patterns/SKILL.md
@@ -1,18 +1,21 @@
 ---
 name: test-anti-patterns
 description: >
-  Detection-focused review of .NET test code for anti-patterns that
-  undermine reliability and diagnostic value.
-  USE FOR: audit test quality, review test code, find test anti-patterns,
-  tests pass but don't verify anything, flaky tests, ordering dependency,
-  duplicate tests, magic values, missing/no assertions, swallowed
-  exceptions, always-true assertions, over-mocking, test coupling, coverage
-  touching, coverage inflation.
-  DO NOT USE FOR: writing new tests (use writing-mstest-tests), direct
-  MSTest API rewrites or implementation-only fixes such as swapped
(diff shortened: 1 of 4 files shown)


Evangelink:
I went with variant B in #671

PR #671, opened 2026-05-19
test-anti-patterns / test-smell-detection: narrow-specialist re-pitch (Variant B)

Evangelink:
## What

**Variant B** of a 3-PR experiment to fix plugin-mode activation failures for `test-anti-patterns` and `test-smell-detection`. Narrow-specialist re-pitch.

### Diagnosis

(Same as Variants A and C — see PR #670 for full diagnosis.) Plugin-mode activation fails because the two skills overlap heavily, the baseline is already 4.0/5, and the previous descriptions told the agent to use the other one.

## Change (Variant B)

* `test-anti-patterns` becomes **the umbrella audit skill** for all pragmatic test-quality reviews (anti-patterns and smell-style prompts alike). Its description explicitly absorbs smell-audit triggers (`test smell audit`, `formal smell audit`, `rank by severity`, etc.).
* `test-smell-detection` is repositioned as a **narrow niche** — invoked only when the user explicitly asks for the `testsmells.org` / academic 19-smell catalog with citable smell names from the research literature.
* Eval prompts for `test-smell-detection` are rewritten so they actually exercise the narrow niche (explicit `testsmells.org` / 19-smell catalog requests).

## Companion variants

* **Variant A** (#670): activation-only rewrite, both skills keep overlap and INVOKE-THIS wording.
* **Variant C** (#TBD-merge): full merge — deletes `test-smell-detection` entirely.

Compare `/evaluate` results across all three to pick the winner.

## Validation

* `skill-validator check --plugin plugins/dotnet-test` passes for both skills.

PR #672, opened 2026-05-19
test-anti-patterns: merge test-smell-detection in (Variant C)

Evangelink:
## What

**Variant C** of a 3-PR experiment to fix plugin-mode activation failures for `test-anti-patterns` and `test-smell-detection`. Full merge — eliminates sibling competition by removing `test-smell-detection`.

### Diagnosis

(Same as Variants A and B — see PR #670 for full diagnosis.) Plugin-mode activation fails because the two skills overlap heavily, the baseline is already 4.0/5, and the previous descriptions told the agent to use the other one.

## Change (Variant C)

* `test-anti-patterns` becomes the **single umbrella audit skill** — its description absorbs every smell-audit trigger phrase and its body references the moved 19-smell catalog for deeper smell-audit needs.
* `plugins/dotnet-test/skills/test-smell-detection/` is **deleted**.
* `references/test-smell-catalog.md` is **moved** under `test-anti-patterns/references/` so the academic taxonomy is still on hand.
* The 4 `test-smell-detection` eval scenarios (smelly OrderProcessor, clean Calculator, integration UserRepository, decline-write) are **ported into the test-anti-patterns eval suite** so coverage is retained, with their fixtures copied under `tests/dotnet-test/test-anti-patterns/fixtures/`.
* `tests/dotnet-test/test-smell-detection/` is **deleted**.
* `plugins/dotnet-test/README.md` and the `test-quality-auditor` agent are updated to drop `test-smell-detection` references.

This is the **most radical** of the three variants: a breaking removal of a public skill, justified only if it materially improves plugin-mode activation.

## Companion variants

* **Variant A** (#670): activation-only rewrite, both skills kept.
* **Variant B** (#671): narrow-specialist — `test-smell-detection` survives as a niche only for explicit `testsmells.org` requests.

Compare `/evaluate` results across all three to pick the winner.

## Validation

* `skill-validator check --plugin plugins/dotnet-test` passes (21 skills, down from 22).


Evangelink:
I went with variant B in #671

PR #682, opened 2026-05-21
Docs: clarify required .NET SDK version

Bortlesboat:
CONTRIBUTING.md currently says '.NET 10 SDK or later', but the repo pins the SDK via global.json (currently a .NET 11 preview).\n\nThis updates the prerequisite line to reference global.json so contributors install the intended SDK and avoid mismatches when running the skill-validator locally.


AbhitejJohn:
@Bortlesboat : Good catch, this looks good if you'd like to publish it.

AbhitejJohn (approved):

PR #700, opened 2026-05-28
run-tests: add platform-detection keywords for activation

Evangelink:
The 'Detect test platform from Directory.Build.props' eval scenario regressed from **2.7/5 to 2.0/5** with `run-tests` showing **NOT ACTIVATED** in the skills-loaded column.

## Why this scenario failed to activate

The scenario prompt asks:

> *I want to run the integration tests in this project with dotnet test. Can you figure out which test platform it uses and run them?*

The previous description framed detection only as:
- `chooses the correct platform/SDK/framework syntax`
- `selecting VSTest vs Microsoft.Testing.Platform command syntax`

Both are about *syntax selection*, not about *detecting* which platform a project uses from its files. The key prompt phrase **"figure out which test platform it uses"** had no matching language in the description.

## Fix

Per `InvestigatingResults.md` guidance for *#5 Skill not activated* failures, surface the detection use case explicitly with the file names the scenario hinges on:

> `detecting test platform/framework from global.json, .csproj, Directory.Build.props`

This mirrors the prompt-shaping approach used in #688.

To stay within the 1,024-char description cap:
- `command syntax` -> `syntax`
- `including` -> `incl.`
- multi-TFM example: drop redundant `dotnet test` prefix

Folded description size: **948 -> 1,010 chars**.

## Validation

`skill-validator check --plugin plugins\dotnet-test` passes with no new warnings.

### DIFF
--- a/plugins/dotnet-test/skills/run-tests/SKILL.md
+++ b/plugins/dotnet-test/skills/run-tests/SKILL.md
@@ -2,19 +2,20 @@
 name: run-tests
 description: >
   Runs .NET tests with `dotnet test` and chooses the correct platform/SDK/framework
-  syntax. USE FOR: running, filtering, or troubleshooting `dotnet test`; detecting
-  test platform/framework from `global.json`, `.csproj`, `Directory.Build.props`;
-  selecting VSTest vs Microsoft.Testing.Platform syntax (incl. `--` separator
-  rules on .NET SDK 8/9 vs 10+); choosing the right filter syntax for MSTest /
-  xUnit / NUnit / TUnit (--filter, --filter-class, --filter-trait, --filter-query,
-  --treenode-filter); TRX and other reporting (--report-trx vs --logger trx);
-  blame/hang/crash diagnostics (--blame-hang-timeout, --blame-crash); running
-  tests against a single target framework when a project targets multiple TFMs
-  (e.g., `<TargetFrameworks>net8.0;net9.0</TargetFrameworks>`, `--framework <TFM>`);
-  and avoiding MTP/VSTest argument mixups (e.g., --logger trx on MTP,
-  --report-trx on VSTest, --blame on MTP).
-  DO NOT USE FOR: writing or generating test code, CI/CD pipeline
-  configuration, or debugging failing test logic.
+  syntax. USE FOR: running, filtering, or troubleshooting `dotnet test`; figuring
+  out which test platform (VSTest vs Microsoft.Testing.Platform) a project uses
+  from `global.json`, `.csproj`, `Directory.Build.props` and selecting the
+  matching syntax (incl. `--` separator rules on .NET SDK 8/9 vs 10+); choosing
+  the right filter syntax for MSTest / xUnit / NUnit / TUnit (--filter,
+  --filter-class, --filter-trait, --filter-query, --treenode-filter); TRX/reporting
+  (--report-trx vs --logger trx); blame/hang/crash diagnostics
+  (--blame-hang-timeout, --blame-crash); running tests against a single target
+  framework when a project targets multiple TFMs (e.g.,
+  `<TargetFrameworks>net8.0;net9.0</TargetFrameworks>`, `--framework <TFM>`); and
+  avoiding MTP/VSTest argument mixups (--logger trx on MTP, --report-trx on
+  VSTest, --blame on MTP).
+  DO NOT USE FOR: writing/generating test code, CI/CD config, or debugging
+  failing test logic.
 license: MIT
 ---
 
--- a/plugins/dotnet-test/skills/run-tests/SKILL.md
+++ b/plugins/dotnet-test/skills/run-tests/SKILL.md
@@ -2,16 +2,17 @@
 name: run-tests
 description: >
   Runs .NET tests with `dotnet test` and chooses the correct platform/SDK/framework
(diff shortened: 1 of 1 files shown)


Evangelink:
Second iteration: rephrased the platform-detection clause with prompt-shaped language to fix plugin-mode activation.

**Why the first fix wasn't enough**

After `021b0a343`, the `Detect test platform from Directory.Build.props` scenario activated in **isolated** mode but still **NOT ACTIVATED** in plugin mode (1.3/5 → 1.7/5). The scenario prompt is:

> *I want to run the integration tests in this project with dotnet test. Can you figure out which test platform it uses and run them?*

The previous clause used `detecting test platform/framework from \global.json\...\` — the verb `detecting` didn't bridge to the prompt's `figure out which test platform` when competing sibling skills (filter-syntax, migrate-vstest-to-mtp, mtp-hot-reload, etc.) were loaded in plugin mode.

**This commit**

- Replaces `detecting test platform/framework from...` with `figuring out which test platform (VSTest vs Microsoft.Testing.Platform) a project uses from \global.json\, \.csproj\, \Directory.Build.props\ and selecting the matching syntax\\. This mirrors the prompt verb exactly and folds the two adjacent USE FOR entries (detection + selection) into one stronger clause.
- Trims verbose tokens elsewhere (`TRX and other reporting` → `TRX/reporting`, `writing or generating` → `writing/generating`, `CI/CD pipeline configuration` → `CI/CD config`, drops `e.g., \\) to stay under the 1024-char folded description cap.

Folded description length: 1010 → 1007 / 1024.

`skill-validator check --plugin plugins/dotnet-test\\ is green.

Please `/evaluate` to confirm plugin-mode activation now succeeds.

Evangelink:
Third iteration: lead the description with platform detection as the primary value proposition.

**Root cause from eval artifact analysis** (run 26581320083)

The plugin agent actually correctly **read `Directory.Build.props` and identified MTP** without the skill (rubric: 5/5 + 5/5 on detection criteria) — but did **NOT load the skill**, so it missed the SDK 9 MTP `--` separator rule and the integration-test filtering requirement (rubric: 1/5 + 1/5 on syntax criteria). That capped plugin quality at 2/5.

The router declined to load the skill because the previous opening sentence `Runs .NET tests with \dotnet test\ and chooses the correct platform/SDK/framework syntax` sounded like base knowledge. The router weighted `I can answer this directly` higher than `maybe I need this skill`. The detection clause was buried as the 2nd USE FOR item, not weighted strongly.

**This commit**

- Lead the description with detection as primary capability:
  > `For \dotnet test\: figures out which test platform (VSTest vs Microsoft.Testing.Platform) a project uses from \Directory.Build.props\, \global.json\, and \.csproj\, then picks the matching command syntax.`
- Add a parallel USE FOR entry `identifying the test runner/platform from project files` (different verbs/nouns than the lead sentence) to widen prompt-phrasing coverage.
- Promote `\--\ separator rules on .NET SDK 8/9 vs 10+` from a parenthetical to a standalone USE FOR entry (specialized knowledge signal).

Folded description length: 1007 → 994 / 1024. `skill-validator check --plugin plugins/dotnet-test` is green.

Please `/evaluate` to confirm plugin-mode activation now succeeds.

Evangelink:
Fourth iteration: teach the skill to map user vocabulary (`integration tests`, `unit tests`...) to filter expressions.

**Why score stayed at 2.3/5 despite activation working**

Eval artifact analysis (run 26584111841) — both isolated and plugin now load the skill, and the agent correctly identifies MTP from `Directory.Build.props` (5/5 + 5/5). But:

- Rubric `Uses MSTest filter syntax for TestCategory`: **1/5** — agent ran plain `dotnet test` with no filter at all
- Rubric `Uses '--' separator before MTP filter args`: **1.7/5** — agent **mentioned** the rule in its summary but never **applied** it (no filter args to combine with the separator)

The prompt says `run the integration tests`, and the fixture has `[TestCategory(""Unit"")]` and `[TestCategory(""Integration"")]` methods, but the skill body had **no guidance on translating user vocabulary into a filter expression** — it just pointed at the \ilter-syntax\ skill. The agent defaulted to running the whole suite.

**This commit** adds three pieces of teaching to `plugins/dotnet-test/skills/run-tests/SKILL.md`:

1. **Workflow pre-step nudge** (after the detection-files line): `If the prompt names a subset of tests (integration tests, smoke tests, a specific class, a specific TFM), plan to apply the matching filter / --framework in Step 3 — do not run the whole suite.` This pushes the agent to commit to filtering **before** running anything.

2. **New subsection in Step 3**, `When the user names a test category, trait, or group`:
   - Table mapping framework → attribute → filter property (`TestCategory` for MSTest, `Category` for NUnit/TUnit, `Category` trait for xUnit)
   - Table of platform/SDK-correct invocations for the `run the integration tests` case across VSTest/MTP and SDK 8/9 vs 10+ (showing the `--` separator in the SDK 8/9 row)
   - Fallback to `FullyQualifiedName~<text>` when no matching attribute is found

3. **New validation checklist item**: `When the user named a test category/trait/group, the appropriate filter was applied (not 'run all tests')`.

Skill size: 3,935 → 4,487 BPE tokens (still `approaching comprehensive` range, below the threshold).

Validator green. Please `/evaluate` — expect significant improvement on this scenario and possibly on `Filter MSTest tests by category on VSTest` too (same vocabulary-mapping concept).

JanKrivanek (approved):

PR #703, opened 2026-05-29
Add MCP servers to dotnet-blazor plugin

javiercn:
## Summary
- add the Microsoft Learn MCP server to plugins/dotnet-blazor/plugin.json
- add the Playwright MCP server to plugins/dotnet-blazor/plugin.json
- allowlist both MCP server declarations in eng/allowed-external-deps.txt


plugins/dotnet-blazor/plugin.json
```diff
@@ -2,5 +2,19 @@
...
   "description": "Skills for Blazor development: component authoring, interactivity, and web application patterns.",
-  "skills": ["./skills/"]
+  "skills": ["./skills/"],
+  "mcpServers": {
+    "microsoft-docs": {
+      "type": "http",
+      "url": "https://learn.microsoft.com/api/mcp",
+      "headers": {},
+      "tools": ["*"]
+    },
```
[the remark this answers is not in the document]
javiercn: Fixed in c12dffff0. skill-validator now tolerates non-stdio MCP entries during plugin discovery and skips unsupported MCP types safely instead of crashing when a plugin contains HTTP MCP servers.

plugins/dotnet-blazor/plugin.json
```diff
@@ -2,5 +2,19 @@
...
+    "playwright": {
+      "type": "local",
```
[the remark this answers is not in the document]
javiercn: Fixed in df9625d0d. The Playwright MCP entry now uses `type: "stdio"`, which matches the evaluator's supported MCP transport.

plugins/dotnet-blazor/plugin.json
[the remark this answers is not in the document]
javiercn: Fixed in c12dffff0. skill-validator now tolerates non-stdio MCP entries during plugin discovery and skips unsupported MCP types safely instead of crashing when a plugin contains HTTP MCP servers.

plugins/dotnet-blazor/plugin.json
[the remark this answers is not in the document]
javiercn: Fixed in df9625d0d. The Playwright MCP entry now uses `type: "stdio"`, which matches the evaluator's supported MCP transport.

plugins/dotnet-blazor/plugin.json
```diff
@@ -2,5 +2,21 @@
...
+      "tools": ["*"]
+    },
+    "playwright": {
+    "playwright": {
+      "type": "stdio",
+      "command": "npx",
+      "args": ["@playwright/mcp@latest"],
+      "tools": ["*"]
+    }
+    }
```
javiercn: @copilot Format is wrong here. Fix it

danroth27 (approved):
I think we should move the MS Learn MCP server to the core dotnet plugin, but otherwise this looks fine to me.

plugins/dotnet-blazor/plugin.json
```diff
@@ -2,5 +2,27 @@
...
+      "headers": {},
+      "tools": [
+        "*"
+      ]
+    },
+    "playwright": {
+      "type": "stdio",
+      "command": "npx",
+      "args": [
+        "@playwright/mcp@latest"
```
AbhitejJohn: Can we specify a version here instead? Security wise latest might be problematic to use in case there's a compromised version released - SDL guidelines apparently.

plugins/dotnet-blazor/plugin.json
```diff
@@ -2,5 +2,27 @@
...
+  "mcpServers": {
+    "microsoft-docs": {
+      "type": "http",
+      "url": "https://learn.microsoft.com/api/mcp",
+      "headers": {},
+      "tools": [
+        "*"
+      ]
+    },
+    "playwright": {
```
AbhitejJohn, line 17: Looks like [this requires node](https://github.com/microsoft/playwright-mcp#requirements) to be installed, which is reasonable. I am curious what a user workflow looks like if someone doesn't have that pre-req. Do we recommend installing it or do we fail?

AbhitejJohn (approved):

plugins/dotnet-blazor/plugin.json
javiercn: You'll get an error in the Copilot CLI saying the MCP server wasn't able to start and you can see the output if you use `/mcp` similar experience in VS Code and (Visual Studio?). Most web developers will have node on their machine in any case.

plugins/dotnet-blazor/plugin.json
javiercn: Hmm, interesting. We can probably do this; it'll be a bit challenging because playwright gets updated every month and you want the latest version because it's paired with the browser versions. The playwright docs themselves give you this snippet as the way to set it up, but we can do a fixed version with something to update it.

@copilot+claude-opus-4.8 instead of using `latest` here. Create a github action that runs every week and updates plugin.json with the latest available version that has been published for more than 24h. The action should use npm to view the available versions and point only to stable versions. If a new version is available, the action will make the change and open a PR with the title `[dotnet-blazor] <<YYYY-MM-DD>> - Update playwright MCP version to <<XXX>>`. While you are at it, make sure you replace the version here too. (Make sure you use the script to validate it works).

AbhitejJohn: Thanks, the changes look good.

AbhitejJohn:
@javiercn : Looks like the token consumption increased without much of a change in quality, based on the evals. Would you mind taking a deeper look please?

AbhitejJohn:
Re-opening cause I know @javiercn wanted to take this in. I'll take a stab at understanding why the evals weren't great from the last run. @javiercn : Feel free to jump on if you get a chance as well.

javiercn:
@AbhitejJohn might be worth punting on this. I'm experimenting with the playwright-cli which is the recommended way forward for agents.

PR #718, opened 2026-06-03
Add migrate-xunit-to-mstest skill

Evangelink:
## Summary

Adds the **`migrate-xunit-to-mstest`** skill, which migrates .NET test projects from xUnit.net (v2 or v3) to MSTest v4 — attributes, assertions, fixtures, lifecycle, output, cancellation, parallelization, and companion packages.

Single skill (not two) for both v2 and v3 sources: ~90% of the conversion logic is shared, and v2-vs-v3 deltas are small and well-scoped — handled as conditional steps. Matches the precedent of `migrate-mstest-v1v2-to-v3`.

## Why parallelization gets front-loaded

xUnit parallelizes test classes by default; MSTest serializes by default. This is the single largest source of post-migration regressions: tests that depend on isolation by parallel scheduling, on the lack of it, or on shared static state can pass differently after the switch.

Step 1 inventories shared state and flaky tests. Step 11 enumerates three explicit choices and provides a translation table for `CollectionBehavior(DisableTestParallelization)`, `MaxParallelThreads`, `[Collection]`, and `[CollectionDefinition]`. The validation checklist treats unspecified parallelization as an error.

## Design constraints (validated with rubber-duck review)

- **Preserves the current test platform** (VSTest stays on VSTest; MTP stays on MTP). Bundling a platform migration would violate the test-migration agent's "never mix migration steps" rule.
- **`IClassFixture<T>` -> mechanical** map to `[ClassInitialize]` / `[ClassCleanup]`.
- **`ICollectionFixture<T>` requires judgement** — it conflates "shared instance across classes" with "serial execution". MSTest decouples them; do not silently widen scope to assembly-level.
- **Exception assertion semantic trap** — xUnit `Assert.Throws<T>` is exact-type and maps to **`Assert.ThrowsExactly<T>`**; xUnit `Assert.ThrowsAny<T>` maps to **`Assert.Throws<T>`**. The names invert.
- **MSTest v4 TFM gating** — supported: `net8.0`, `net9.0`, `net462`+, `netstandard2.0` (test library only), UAP/WinUI. **Stop** for `net5.0`-`net7.0`.

## Files

- `plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md` — 13-step workflow
- `plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md` — full attribute/assertion/fixture/lifecycle tables (progressive disclosure)
- `tests/dotnet-test/migrate-xunit-to-mstest/eval.yaml` — 12 scenarios (skill-validator format, inline `setup.files`)
- `tests/dotnet-test/migrate-xunit-to-mstest/eval.vally.yaml` + `fixtures/` — Vally-format mirror of the same scenarios (40 fixture files), matching the convention of every sibling migration skill
- `plugins/dotnet-test/README.md` — migration table row + intro bullet
- `plugins/dotnet-test/agents/test-migration.agent.md` — triage row + two multi-step rules (xUnit→MSTest, xUnit→MSTest+MTP)

## Evaluation scenarios (12)

1. xUnit v2 -> MSTest preserving VSTest
2. xUnit v3 -> MSTest preserving VSTest
3. Stop when TFM unsupported by MSTest v4 (`net7.0`)
4. `IClassFixture<T>` -> `[ClassInitialize]` / `[ClassCleanup]`
5. `ICollectionFixture<T>` -> explicit scope decision (`[DoNotParallelize]`)
6. `ITestOutputHelper` -> `TestContext`
7. Exception assertions — `Throws` vs `ThrowsAny` semantic trap
8. `MemberData` / `TheoryData` -> `DynamicData`
9. `[Fact(Skip)]` / `[Trait]` / `Timeout` -> `[Ignore]` / `[TestCategory]` / `[TestProperty]` / `Timeout`
10. **Parallelization preservation** -> `[assembly: Parallelize(ClassLevel)]`
11. xUnit v3 `TestContext.Current.CancellationToken` -> MSTest `TestContext.CancellationToken`
12. Project already on MSTest — no-op

## Static checks

`dotnet run --project eng/skill-validator/src -- check --plugin ./plugins/dotnet-test`

✅ All checks passed (24 skill(s), 11 agent(s), 1 plugin(s))

One soft warning: SKILL.md is 8.1k BPE tokens (above the "approaching comprehensive" threshold). Splitting was considered and rejected — the workflow is sequential and benefits from being one document; non-essential content already lives in `references/mapping-cheatsheet.md`.

### DIFF
--- a/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
@@ -62,7 +62,7 @@ For writing idiomatic MSTest code (modern assertion APIs, lifecycle patterns, da
 
 ## Workflow
 
-> **Commit strategy:** Commit after Step 2 (packages updated, builds broken), after Step 5 (attributes converted, most asserts fixed), and after Step 9 (fixtures/lifecycle rewritten, tests pass). Commit before fixing follow-up cleanup so reviewers can bisect.
+> **Commit strategy:** Commit after Step 2 (packages updated, builds broken), after Step 6 (attributes converted, asserts fixed), and after Step 8 (fixtures/lifecycle rewritten, tests pass). Commit before fixing follow-up cleanup so reviewers can bisect.
 
 ### Step 1: Assess the project
 
@@ -80,7 +80,7 @@ For writing idiomatic MSTest code (modern assertion APIs, lifecycle patterns, da
 5. Inventory high-risk patterns -- scan for these and flag them now so you can plan judgement steps later:
    - **Parallelization differences (Step 11)** -- xUnit parallelizes test classes by default; MSTest does not. This is the **single most common source of post-migration regressions**: tests that depended on isolation by parallel scheduling, on the lack of it, or on shared static state can pass differently. Decide the target parallelization model now -- do not leave it as the MSTest default by accident.
    - `ICollectionFixture<T>` / `[CollectionDefinition]` (scope concern -- see Step 8)
-   - Custom `DataAttribute` / custom `FactAttribute` / custom `TheoryAttribute` subclasses (manual conversion to `ITestDataSource` / `TestMethodAttribute` -- see Step 6)
+   - Custom `DataAttribute` / custom `FactAttribute` / custom `TheoryAttribute` subclasses (manual conversion to `ITestDataSource` / `TestMethodAttribute` -- see Step 5)
    - `Assert.Throws<T>` (xUnit semantics = exact type; maps to `Assert.ThrowsExactly<T>`, **not** `Assert.Throws<T>`)
    - `Record.Exception` / `Record.ExceptionAsync` (manual conversion)
    - `Assert.Raises*` / event assertions (no MSTest equivalent -- manual)
@@ -227,7 +227,7 @@ Reversing these flips the assertion semantics silently. Verify by name, not by v
 
 | xUnit | MSTest |
 |---|---|
-| Constructor (sync setup) | Keep constructor (MSTest also instantiates per test). Drop xUnit-only `ITestOutputHelper` param -- see Step 8 |
+| Constructor (sync setup) | Keep constructor (MSTest also instantiates per test). Drop xUnit-only `ITestOutputHelper` param -- see Step 9 |
 | `Dispose()` (sync teardown) | Keep `Dispose()` (MSTest supports `IDisposable`) **or** rewrite as `[TestCleanup] public void Cleanup() { ... }` |
 | `DisposeAsync()` (async teardown) | Keep `IAsyncDisposable.DisposeAsync()` **or** rewrite as `[TestCleanup] public async Task CleanupAsync() { ... }` |
 | `IAsyncLifetime.InitializeAsync` | `[TestInitialize] public async Task InitAsync() { ... }` |
--- a/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
@@ -45,6 +45,10 @@ This is a **cross-framework** migration. Do not bundle it with a version upgrade
 
 - **Always identify the current xUnit version first.** State whether the project is on xUnit v2 (`xunit` 2.x) or xUnit v3 (`xunit.v3` / `xunit.v3.*`) before recommending changes. This grounds the migration advice -- some breaking-change steps only apply to one version.
 - **Always preserve the current test platform.** If the project runs on VSTest, keep VSTest. If it runs on MTP (e.g., xUnit v3 native MTP, or `<UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>`), keep MTP. Recommend `migrate-vstest-to-mtp` as a separate follow-up only if the user asks for it.
+- **Explicitly communicate every judgement-call decision** before applying it -- otherwise the user cannot tell what changed semantically. In particular:
+  - **Fixture scope changes** (Step 8): state the source scope (class / collection / assembly) and the target scope you chose, plus what gets shared and what gets serialized. A silent widening from collection to assembly is the most common way this migration regresses tests.
+  - **Parallelization** (Step 11): state that **MSTest defaults to serial execution** (xUnit parallelizes classes by default), so an explicit `[assembly: Parallelize(...)]` is **required** to match xUnit's behaviour -- omitting it silently halves CI throughput.
+  - **`Assert.Throws<T>` -> `Assert.ThrowsExactly<T>`** (Step 6): mention the exact-type-vs-any-derived semantic flip so reviewers know the assertion was deliberately renamed, not just translated.
 - **Specific API mapping questions** (assertions, fixtures, output helper, etc.): jump to the relevant step. Do not run the full workflow.
 - **Full migration requests**: follow the workflow end-to-end.
 - **Focused fix requests** (specific compile error after a partial migration): address only that error using the mapping reference. Do not walk the full workflow.
@@ -90,7 +94,7 @@ For writing idiomatic MSTest code (modern assertion APIs, lifecycle patterns, da
 
 ### Step 2: Replace packages
 
-> Choose the package option that matches what the project uses today. Default to the `MSTest` metapackage unless the user already uses `MSTest.Sdk` elsewhere in the solution. If you adopt `MSTest.Sdk`, you must preserve the platform from Step 1.
+> Choose the package option that matches what the project uses today. **When the user says "preserve VSTest" -- or the existing project uses explicit `PackageReference`s -- default to Option A (`MSTest` metapackage).** Reach for Option B (`MSTest.Sdk`) only when the user explicitly asks to modernize the SDK or already uses `MSTest.Sdk` elsewhere in the solution; if you adopt it, you must preserve the platform from Step 1.
 
 **Remove** every xUnit package reference (from `.csproj`, `Directory.Build.props`, `Directory.Packages.props`):
 
@@ -167,13 +171,15 @@ Apply these rewrites to every C# test file. Class-level first, then method-level
 
 **Methods:**
 
+> **`[Ignore]` and `[Timeout]` are modifiers, not discovery attributes.** Always emit `[TestMethod]` *alongside* them -- a method with `[Ignore]` but no `[TestMethod]` is silently skipped by the test runner (no error, no skip count). Same for `[Timeout]`.
+
 | xUnit | MSTest |
 |---|---|
 | `[Fact]` | `[TestMethod]` |
 | `[Theory]` | `[TestMethod]` (parameterized; MSTest 3+ no longer needs `[DataTestMethod]`) |
 | `[Fact(DisplayName = "x")]` | `[TestMethod("x")]` (v3 of MSTest) or `[TestMethod(DisplayName = "x")]` (v4) |
-| `[Fact(Skip = "reason")]` | `[TestMethod]` + `[Ignore("reason")]` (the `[Ignore]` attribute alone does not discover a test) |
-| `[Fact(Timeout = 5000)]` | `[TestMethod]` + `[Timeout(5000)]` (modifier -- not a discovery attribute) |
+| `[Fact(Skip = "reason")]` | `[TestMethod]` + `[Ignore("reason")]` (both attributes required) |
+| `[Fact(Timeout = 5000)]` | `[TestMethod]` + `[Timeout(5000)]` (both attributes required) |
 | `[Trait("Category", "Unit")]` | `[TestCategory("Unit")]` |
 | `[Trait("Owner", "alice")]` | `[TestProperty("Owner", "alice")]` |
 
@@ -288,7 +294,9 @@ xUnit collections do two things simultaneously: (1) share a fixture instance acr
 - **Many classes, fixture is genuinely assembly-wide** (e.g., process-wide TestServer): hoist to `[AssemblyInitialize]` / `[AssemblyCleanup]` in a dedicated `AssemblySetup` class **and** confirm with the user that widening the scope is acceptable. Note that this changes parallelization semantics.
 - **Custom collection behavior or test-collection-orderer**: stop and flag for manual review.
 
-**Always explain the scope change** -- silently widening fixture scope across the assembly is the most common way this migration regresses tests.
+> **REQUIRED -- communicate the scope decision before applying it.** Silently widening fixture scope across the assembly is the most common way this migration regresses tests. Use this template (replace bracketed text):
+>
+> "The xUnit `[Collection(\"<name>\")]` shared a `<Fixture>` between **\<N\> classes** and serialized them. I am mapping that to: a static `Lazy<<Fixture>>` shared by each class's `[ClassInitialize]` (scope: **per-class, shared via static** -- not widened to assembly), plus `[DoNotParallelize]` on `<ClassA>` and `<ClassB>` to preserve the serialization. The alternative -- `[AssemblyInitialize]` -- would widen the fixture to every test in the assembly, which I rejected because \<reason\>."
 
 ### Step 9: Convert output and TestContext
 
@@ -368,6 +376,12 @@ public TestContext TestContext { get; set; } = null!;
 ```
 
 Use this when the suite was healthy on xUnit and you want zero behavioural change. It preserves "parallel across classes, serial within a class" exactly.
+
+> **REQUIRED -- explicitly tell the user why this attribute is needed.** When applying Choice A, include this sentence (verbatim or near-verbatim) in your final summary:
+>
+> "MSTest defaults to **serial** execution across classes (unlike xUnit, which parallelizes classes by default), so this `[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.ClassLevel)]` is **required** to match the project's previous xUnit parallel-class behaviour. Without it, the suite would still pass but run roughly one-class-at-a-time and CI throughput would drop."
+>
+> The user must understand this is **opt-in** under MSTest -- a silent omission looks like a no-op but is actually a behavioural regression.
 
 **Choice B -- Adopt MSTest's serial default:**
 
--- a/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
@@ -333,11 +333,40 @@ If the project pins MSTest < 3.6 (rare after Step 2), use property injection ins
 public TestContext TestContext { get; set; } = null!;
 ```
 
-**xUnit v3 `TestContext.Current`:**
+**xUnit v3 `TestContext.Current`** (`TestContext.Current` is **static** in xUnit v3; in MSTest you must use the **instance** `TestContext` obtained via the same constructor or property injection shown above):
 
-- `TestContext.Current.CancellationToken` -> `TestContext.CancellationToken` (MSTest 3.6+, on the **instance** TestContext from constructor or property injection)
-- `TestContext.Current.AddAttachment(name, path)` -> `TestContext.AddResultFile(path)`
-- `TestContext.Current.TestOutputHelper.WriteLine(...)` -> `TestContext.WriteLine(...)`
+- `TestContext.Current.CancellationToken` -> `_testContext.CancellationToken` (MSTest 3.6+)
+- `TestContext.Current.AddAttachment(name, path)` -> `_testContext.AddResultFile(path)`
+- `TestContext.Current.TestOutputHelper.WriteLine(...)` -> `_testContext.WriteLine(...)`
+
+> **REQUIRED for CancellationToken:** Add the constructor injection from above even if the class only uses `TestContext.Current.CancellationToken` (no `ITestOutputHelper`). Do **NOT** replace `TestContext.Current.CancellationToken` with a new `CancellationTokenSource` -- that loses the test-host's cancellation linkage and changes behavior under timeouts.
+
+```csharp
+// xUnit v3
+[Fact]
+public async Task WorkRespectsCancellation()
+{
+    var ct = TestContext.Current.CancellationToken;
+    await Task.Delay(1, ct);
+    Assert.False(ct.IsCancellationRequested);
+}
+
+// MSTest (note: Assert.False -> Assert.IsFalse from Step 6)
+[TestClass]
+public sealed class MyTests
+{
+    private readonly TestContext _testContext;
+    public MyTests(TestContext testContext) => _testContext = testContext;
+
+    [TestMethod]
+    public async Task WorkRespectsCancellation()
+    {
+        var ct = _testContext.CancellationToken;
+        await Task.Delay(1, ct);
+        Assert.IsFalse(ct.IsCancellationRequested);
+    }
+}
+```
 
 ### Step 10: Convert companion packages
 
--- a/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
@@ -166,7 +166,7 @@ Apply these rewrites to every C# test file. Class-level first, then method-level
 | `[Trait("Category", "Unit")]` | `[TestCategory("Unit")]` |
 | `[Trait("Owner", "alice")]` | `[TestProperty("Owner", "alice")]` |
 
-> Use `[TestCategory]` for the conventional category trait (it integrates with VSTest/MTP filter syntax); use `[TestProperty]` for arbitrary key/value metadata.
+> Use `[TestCategory]` for the conventional category trait (it integrates with VSTest/MTP filter syntax); use `[TestProperty]` for arbitrary key/value metadata. Both attributes target `Assembly`, `Class`, and `Method` -- so an xUnit `[Trait]` placed at any of those scopes (including `[assembly: Trait(...)]`) keeps the same scope under MSTest (see Step 12).
 
 ### Step 5: Convert data-driven tests
 
@@ -422,16 +422,22 @@ Delete `xunit.runner.json`. Port relevant settings:
 
 If the project uses xUnit traits in CI filter expressions (e.g., `--filter "Category=Unit"` with xUnit), the equivalent MSTest filter is `--filter "TestCategory=Unit"` (VSTest) or `--filter-trait "TestCategory=Unit"` (MTP). Update CI pipelines accordingly.
 
-### Step 12: Remove xUnit assembly attributes
+### Step 12: Convert xUnit assembly attributes
 
-Delete these from `AssemblyInfo.cs` / any `[assembly: ...]` declaration:
+Some xUnit assembly attributes have direct MSTest equivalents at assembly scope; others must be removed (and re-applied per class/method) or reimplemented against MSTest extensibility.
 
-- `[assembly: CollectionBehavior(...)]`
-- `[assembly: TestCaseOrderer(...)]`
-- `[assembly: TestCollectionOrderer(...)]`
+**Convert (assembly scope preserved):**
+
+- `[assembly: Xunit.Trait("Category", "v")]` -> `[assembly: TestCategory("v")]` -- `TestCategoryAttribute` targets `Assembly`, `Class`, and `Method`; assembly application propagates to every test.
+- `[assembly: Xunit.Trait("k", "v")]` (non-category key) -> `[assembly: TestProperty("k", "v")]` -- same `Assembly`/`Class`/`Method` targets.
+
+**Delete (no MSTest equivalent or now handled elsewhere):**
+
+- `[assembly: CollectionBehavior(...)]` -- replaced by `[assembly: Parallelize(...)]` (Step 11)
+- `[assembly: TestCaseOrderer(...)]` -- reimplement against MSTest extensibility; flag for manual conversion
+- `[assembly: TestCollectionOrderer(...)]` -- flag for manual conversion
 - `[assembly: TestFramework(...)]`
-- `[assembly: CaptureConsole]` (xUnit v3)
-- `[assembly: Xunit.Trait("k", "v")]` (replace with `[TestCategory]`/`[TestProperty]` on test classes/methods)
+- `[assembly: CaptureConsole]` (xUnit v3) -- MSTest does not capture console by default
 
 Custom orderers/test framework hooks must be reimplemented against MSTest's extensibility model (`TestMethodAttribute` subclasses, `ITestDataSource`, etc.) -- stop and flag for manual conversion if present.
 
--- a/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
@@ -165,7 +165,7 @@ Apply these rewrites to every C# test file. Class-level first, then method-level
 **Class:**
 
 - Add `[TestClass]` to every class that contained xUnit `[Fact]`/`[Theory]` methods (xUnit had no class-level requirement).
-- Mark the class `sealed` (MSTest idiom -- see `writing-mstest-tests`).
+- **Preserve the original class hierarchy.** xUnit projects often use base/derived test classes (shared setup, helper assertions, generic base fixtures); marking classes `sealed` would break that pattern. Sealing is an optional follow-up handled by `writing-mstest-tests`, not part of the mechanical migration.
 - Replace `using Xunit;` / `using Xunit.Abstractions;` with `using Microsoft.VisualStudio.TestTools.UnitTesting;`.
 
 **Methods:**
--- a/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
@@ -119,28 +119,38 @@ The `MSTest` metapackage pulls in `MSTest.TestFramework`, `MSTest.TestAdapter`, 
 ```xml
 <Project Sdk="MSTest.Sdk/4.1.0">
   <PropertyGroup>
-    <TargetFramework>net9.0</TargetFramework>
+    <!-- Keep the project's existing TargetFramework -- do NOT retarget during migration. -->
+    <TargetFramework>$(ExistingTargetFramework)</TargetFramework>
   </PropertyGroup>
 </Project>
 ```
 
-`MSTest.Sdk` defaults to **MTP** and does *not* include `Microsoft.NET.Test.Sdk`. To preserve a VSTest project, you must opt back in:
+`MSTest.Sdk` defaults to **MTP**. To preserve a VSTest project, opt back in with `<UseVSTest>true</UseVSTest>` -- the SDK then pulls in `Microsoft.NET.Test.Sdk` automatically (no extra `PackageReference` needed):
 
 ```xml
 <PropertyGroup>
   <UseVSTest>true</UseVSTest>
 </PropertyGroup>
-<ItemGroup>
-  <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />
-</ItemGroup>
 ```
+
+For solutions with several test projects, prefer pinning the `MSTest.Sdk` version in `global.json` so it lives in one place:
+
+```json
+{
+  "msbuild-sdks": {
+    "MSTest.Sdk": "4.1.0"
+  }
+}
+```
+
+With the pin in `global.json`, the project line simplifies to `<Project Sdk="MSTest.Sdk">`.
 
 When switching to `MSTest.Sdk`, also remove now-redundant properties: `<OutputType>Exe</OutputType>`, `<IsPackable>false</IsPackable>`, `<IsTestProject>true</IsTestProject>`, `<EnableMSTestRunner>`.
 
 ### Step 3: Update project configuration
 
 1. **Preserve the runner.** Confirm the platform decision from Step 1 still holds after Step 2. Common mistakes:
-   - Switching to `MSTest.Sdk` without `UseVSTest=true` silently flips a VSTest project to MTP. Add `Microsoft.NET.Test.Sdk` back.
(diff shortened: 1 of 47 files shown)


plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+
+**Add** MSTest v4 -- pick exactly one of:
+
+```xml
+<!-- Option A: metapackage (pulls in TestFramework + TestAdapter + Analyzers + Microsoft.NET.Test.Sdk) -->
+<PackageReference Include="MSTest" Version="4.1.0" />
+```
+
+```xml
+<!-- Option B: MSTest.Sdk (MTP-first; add Microsoft.NET.Test.Sdk if you need VSTest) -->
```
Evangelink: You don't need manual reference to `Microsoft.NET.Test.Sdk` for VSTest when using MSTest.Sdk, instead add the following property `<UseVSTest>true</UseVSTest>`

Evangelink:
Thanks for the review! Pushed two follow-up commits:

**[c169d65](https://github.com/dotnet/skills/pull/718/commits/c169d656b) — Fix step cross-references** (review bot findings)
- `SKILL.md` Step 1.5 `Custom DataAttribute` bullet pointed to Step 6 (assertions); now points to Step 5 (data-driven tests) where the mapping lives.
- `SKILL.md` Step 7 lifecycle table `Constructor` row pointed to Step 8 (fixtures); now points to Step 9 (output and TestContext) where `ITestOutputHelper` conversion is described.
- Also noticed the commit-strategy callout had the same drift: it said "after Step 5 (asserts fixed)" and "after Step 9 (fixtures/lifecycle rewritten)". Asserts are in Step 6 and fixtures/lifecycle wrap up at Step 8 — fixed.

**[7e6fc17](https://github.com/dotnet/skills/pull/718/commits/7e6fc17f8) — Map `[assembly: Xunit.Trait]` to `[assembly: TestCategory]` / `[assembly: TestProperty]` (@Evangelink)**

You're right — both `TestCategoryAttribute` and `TestPropertyAttribute` carry `AttributeUsage(AttributeTargets.Assembly | Class | Method)`, so an xUnit `[assembly: Trait]` translates 1:1 to the corresponding MSTest assembly attribute without losing scope.

- `references/mapping-cheatsheet.md` §8 now maps `[assembly: Xunit.Trait("Category", "v")]` -> `[assembly: TestCategory("v")]` and the non-category form -> `[assembly: TestProperty("k", "v")]` instead of recommending removal.
- `SKILL.md` Step 12 renamed from "Remove xUnit assembly attributes" to "Convert xUnit assembly attributes" with the trait mappings inline.
- The notes after the Step 4 attribute table and §1 of the cheatsheet now disclose that `[TestCategory]` / `[TestProperty]` target `Assembly`, `Class`, and `Method`, so users don't assume class/method-only scope.

## On vally evals
The skill already ships an `eval.vally.yaml` at `tests/dotnet-test/migrate-xunit-to-mstest/eval.vally.yaml` with 40 fixture files mirroring all 12 `eval.yaml` scenarios — discovered by the `dotnet-test` suite glob `tests/dotnet-test/*/eval.vally.yaml` in `.vally.yaml`. The `vally-evaluate: SKIPPED` status on the prior run just reflects that the evaluation pipeline is gated on `/evaluate` / `evaluate-now`; will trigger below.

`dotnet run --project eng/skill-validator/src -- check --plugin ./plugins/dotnet-test` → ✅ All checks passed locally.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+<Project Sdk="MSTest.Sdk/4.1.0">
```
Evangelink, line 348: I usually prefer to have the version defined in global.json:
{
    "msbuild-sdks": {
        "MSTest.Sdk": "4.1.0"
    }
}

but maybe this is easier for LLM

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+```
+
+## 10. Companion / extension libraries
+
+| xUnit companion | MSTest equivalent |
+|---|---|
+| `Xunit.SkippableFact` (`[SkippableFact]`, `Skip.If`, `Skip.IfNot`) | `[Ignore]` (compile-time) or `Assert.Inconclusive("reason")` (runtime). Remove the package |
+| `Xunit.Combinatorial` (`[CombinatorialData]`, `[CombinatorialValues]`) | No equivalent -- expand to explicit `[DataRow]` or `[DynamicData]` |
```
Evangelink: For MSTest, we can use https://github.com/Youssef1313/Combinatorial.MSTest

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+
+| xUnit | MSTest |
+|---|---|
+| *(no class attribute)* | `[TestClass]` (required) |
+| *(no class modifier)* | `sealed` (MSTest idiom -- see `writing-mstest-tests`) |
+| `[Fact]` | `[TestMethod]` |
+| `[Theory]` | `[TestMethod]` (MSTest 3+ unified; `[DataTestMethod]` still works but is not needed) |
+| `[Fact(DisplayName = "x")]` | MSTest 4: `[TestMethod(DisplayName = "x")]`; MSTest 3: `[TestMethod("x")]` |
+| `[Theory(DisplayName = "x")]` | Same as above on the `[TestMethod]` |
+| `[Fact(Skip = "reason")]` | `[Ignore("reason")]` |
```
Evangelink: The [TestMethod] is required in addition to [Ignore]

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+| `[Fact(Timeout = 5000)]` | `[Timeout(5000)]` |
```
Evangelink: [TestMethod] should also be applied.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+| `[Trait("Category", "Unit")]` | `[TestCategory("Unit")]` |
+| `[Trait("Owner", "alice")]` | `[TestProperty("Owner", "alice")]` |
+| `[Collection("Db")]` | Step 8 + Step 11: `[DoNotParallelize]` (serialization) + `[ClassInitialize]` (sharing) -- preserve scope explicitly |
+| Custom `FactAttribute` subclass | Custom `TestMethodAttribute` subclass overriding `ExecuteAsync` (MSTest v4). See `writing-mstest-tests` and `migrate-mstest-v3-to-v4` for `CallerInfo` constructor pattern |
+| Custom `TheoryAttribute` subclass | Same -- subclass `TestMethodAttribute`; expose data via `ITestDataSource` |
```
Evangelink, line 47: From some examples I have seen, we should also point out the condition attributes that can be a better pick.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+
+> Use `[TestCategory]` for conventional category traits (recognized by VSTest/MTP filter syntax: `--filter "TestCategory=Unit"` or `--filter-trait "TestCategory=Unit"`). Use `[TestProperty]` for arbitrary key/value metadata that does not need filter integration.
```
Evangelink: TestProperty can be filtered

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+## 2. Data-driven tests
+
+| xUnit | MSTest |
+|---|---|
+| `[InlineData(1, 2)]` | `[DataRow(1, 2)]` |
+| `[InlineData(1, DisplayName = "case 1")]` | `[DataRow(1, DisplayName = "case 1")]` |
+| `[InlineData(null)]` | `[DataRow(null)]` |
+| `[MemberData(nameof(Cases))]` returning `IEnumerable<object[]>` | `[DynamicData(nameof(Cases))]` returning `IEnumerable<object[]>` |
+| `[MemberData(nameof(Cases), MemberType = typeof(X))]` | `[DynamicData(nameof(Cases), typeof(X))]` |
+| `[MemberData(nameof(Cases))]` returning `TheoryData<int, string>` | `[DynamicData(nameof(Cases))]` returning `IEnumerable<object[]>` or `IEnumerable<(int, string)>` (MSTest 3.7+ ValueTuple support) |
```
Evangelink: We also support TestDataRow<T> (see https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-writing-tests-data-driven#supported-data-source-types)

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,352 @@
...
+### 3.9 Skip / inconclusive
+
+> xUnit `Assert.Skip*` is **runtime** (decided inside the test body). MSTest `[Ignore]` is **compile-time** (decided at discovery). They are not interchangeable -- mapping `SkipUnless` to `[Ignore]` will permanently exclude the test on machines where it should have run.
+
+| xUnit | MSTest |
+|---|---|
+| `[Fact(Skip = "reason")]` | `[Ignore("reason")]` |
+| `Assert.Skip("reason")` (xUnit v3) | `Assert.Inconclusive("reason")` |
+| `Assert.SkipWhen(condition, "reason")` (xUnit v3) | `if (condition) Assert.Inconclusive("reason");` |
+| `Assert.SkipUnless(condition, "reason")` (xUnit v3) | `if (!condition) Assert.Inconclusive("reason");` |
```
Evangelink: It's not a direct replacement but using Condition attribute could prevent the need for this assert

Evangelink:
Thanks for the detailed review! Second round of fixes is in `b962a08`.

**MSTest semantics**
- `[Ignore]` and `[Timeout]` are *modifiers* — they don't discover a test on their own. `[Fact(Skip = ...)]` / `[Fact(Timeout = N)]` now map to `[TestMethod] + [Ignore(...)]` / `[TestMethod] + [Timeout(N)]` everywhere (SKILL.md Step 4, cheatsheet §1). Eval scenario 9 rubric tightened to require both.
- `TestPropertyAttribute` IS filterable (`--filter "Owner=alice"`) — corrected the note that wrongly described it as non-filterable metadata. Both filter syntaxes shown.
- `TestCategoryAttribute` / `TestPropertyAttribute` are valid at assembly scope too — already fixed in `7e6fc17`.

**Condition attributes (MSTest 3.10+)**
- `[OSCondition]`, `[CICondition]`, `[ArchitectureCondition]`, `[NonParallelizableCondition]` now called out as the preferred alternative to `Assert.SkipUnless` / `Assert.Inconclusive` for environmental gates. Added explicit mapping rows + Common Pitfalls note.

**Data sources**
- `TestDataRow<T>` added as a strongly-typed alternative alongside the existing `IEnumerable<object[]>` / `DynamicData` options (SKILL.md Step 5, cheatsheet §2).

**MSTest.Sdk**
- Removed the incorrect "add `Microsoft.NET.Test.Sdk` PackageReference back" instruction — `MSTest.Sdk` + `<UseVSTest>true</UseVSTest>` pulls it in automatically (Step 2 Option B, Step 3, Common Pitfalls).
- Added the recommended `global.json` `msbuild-sdks` pin pattern.
- Replaced hard-coded `<TargetFramework>net9.0</TargetFramework>` in the Option B example with a placeholder + comment (Copilot finding).

**Companion ports**
- `Xunit.Combinatorial` → `Combinatorial.MSTest` (Youssef1313's community port with matching attribute surface), SKILL.md Step 10 + cheatsheet §10.

Validator green locally (24 skills, 11 agents). Re-running `/evaluate` next.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,484 @@
...
+   - **Supported**: `net8.0`, `net9.0`, `net462`+, `netstandard2.0` (test library only), `uap10.0.16299`, `net8.0-windows10.0.18362.0` (WinUI), `net9.0-windows10.0.17763.0` (modern UWP).
+   - **Unsupported**: .NET Core 3.1, `net5.0`-`net7.0`. **STOP** and ask the user to upgrade the TFM first, or migrate to MSTest v3 (then use `migrate-mstest-v3-to-v4` after a TFM bump).
+5. Inventory high-risk patterns -- scan for these and flag them now so you can plan judgement steps later:
+   - **Parallelization differences (Step 11)** -- xUnit parallelizes test classes by default; MSTest does not. This is the **single most common source of post-migration regressions**: tests that depended on isolation by parallel scheduling, on the lack of it, or on shared static state can pass differently. Decide the target parallelization model now -- do not leave it as the MSTest default by accident.
+   - `ICollectionFixture<T>` / `[CollectionDefinition]` (scope concern -- see Step 8)
+   - Custom `DataAttribute` / custom `FactAttribute` / custom `TheoryAttribute` subclasses (manual conversion to `ITestDataSource` / `TestMethodAttribute` -- see Step 6)
```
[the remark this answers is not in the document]
Evangelink: Fixed in c169d65 — bullet now points to Step 5.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,484 @@
...
+- `Assert.Raises<T>` / `Assert.PropertyChanged` -> manual event subscription + flag check
+- `Record.Exception` / `Record.ExceptionAsync` -> `try/catch` returning the exception (or `Assert.ThrowsExactly<T>` if you know the type)
+
+### Step 7: Convert lifecycle
+
+**Constructor / `IDisposable` / `IAsyncDisposable` / `IAsyncLifetime`:**
+
+| xUnit | MSTest |
+|---|---|
+| Constructor (sync setup) | Keep constructor (MSTest also instantiates per test). Drop xUnit-only `ITestOutputHelper` param -- see Step 8 |
```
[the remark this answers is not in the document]
Evangelink: Fixed in c169d65 — row now references Step 9.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Good catch — fixed in b962a08. Cheatsheet §9 and SKILL.md Step 2 Option B / Step 3 now use `<UseVSTest>true</UseVSTest>` only and drop the manual `Microsoft.NET.Test.Sdk` PackageReference. Common Pitfalls table updated too.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Added in b962a08 — cheatsheet §9 now shows the `global.json` `msbuild-sdks` pin as the recommended pattern, with the simplified `<Project Sdk="MSTest.Sdk">` line.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Switched in b962a08 — both SKILL.md Step 10 and cheatsheet §10 now point at `Combinatorial.MSTest` (Youssef1313 community port) instead of `Xunit.Combinatorial`.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Fixed in b962a08. `[Ignore]` is a modifier — Step 4 + cheatsheet §1 now show `[Fact(Skip)]` → `[TestMethod] + [Ignore(...)]`. Also tightened the eval scenario 9 rubric to require both.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Fixed in b962a08 — same treatment for `[Fact(Timeout = N)]` → `[TestMethod] + [Timeout(N)]`. Eval scenario 9 rubric tightened to require both.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Added in b962a08 — cheatsheet §3.9 + §1 + SKILL.md Step 6 / Step 10 / Common Pitfalls now call out the MSTest 3.10+ condition attributes (`[OSCondition]`, `[CICondition]`, `[ArchitectureCondition]`, `[NonParallelizableCondition]`) as the preferred alternative to `Assert.SkipUnless` / `Assert.Inconclusive`, with explicit mapping examples.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Fixed in b962a08 — the note after the trait tables in both SKILL.md and cheatsheet §1 now correctly states `TestProperty` is filterable (`--filter "Owner=alice"`).

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,490 @@
...
+The `MSTest` metapackage pulls in `MSTest.TestFramework`, `MSTest.TestAdapter`, `MSTest.Analyzers`, and `Microsoft.NET.Test.Sdk` -- so VSTest discovery (`vstest.console`, classic `dotnet test`) still works.
+
+**Option B -- `MSTest.Sdk`:**
+
+```xml
+<Project Sdk="MSTest.Sdk/4.1.0">
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+  </PropertyGroup>
+</Project>
```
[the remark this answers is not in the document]
Evangelink: Fixed in b962a08 — replaced the hard-coded `<TargetFramework>net9.0</TargetFramework>` with a placeholder + comment so the example preserves the existing TFM.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Added in b962a08 — cheatsheet §2 and SKILL.md Step 5 now list `TestDataRow<T>` as a strongly-typed option with per-row metadata, with link to the docs.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Addressed in b962a08 — cheatsheet §3.9 now leads with the condition attributes as the preferred alternative for environmental gates (with mapping examples), keeping `Assert.Skip*` as the runtime fallback for dynamic predicates.

Evangelink:
Pushed d10bc59b to address the Convert xUnit v3 TestContext.Current.CancellationToken scenario.

**Diagnosis (from inspecting the eval artifacts):**

The 1.0/5 score wasn't an activation bug -- in plugin mode the skill *was* activated and used dit/bash/grep/view, but still scored 1.0. The judge rejected outputs that:

- Replaced TestContext.Current.CancellationToken with a freshly-created CancellationTokenSource (rubric: `Does NOT introduce CancellationTokenSource as a fake replacement`)
- Failed to inject MSTest's instance TestContext (rubric: `Uses MSTest's instance TestContext (constructor injection or property)`)
- Forgot to also rename Assert.False -> Assert.IsFalse, so dotnet test failed

The previous wording at Step 9 was just three one-line mappings under `xUnit v3 \TestContext.Current\:`; no example, no explicit "do not fabricate a token source" guardrail.

**Fix:**

- Added a full xUnit v3 -> MSTest worked example for the CancellationToken case at Step 9, showing the constructor-injection scaffold + the Assert.False -> Assert.IsFalse follow-through.
- Added a **REQUIRED** callout: add the constructor injection even when the class has no ITestOutputHelper, and **never** swap for a new CancellationTokenSource.
- Mirrored the CancellationTokenSource warning in 
eferences/mapping-cheatsheet.md row 6.

The NOT ACTIVATED marker on the isolated run looked like an SDK race (events.jsonl shows skill.invoked *after* session.shutdown), not a description-match issue -- isolated activation is high-variance (CV=85%) on this scenario. Making the skill more directly actionable should also reduce the chance the agent stalls on its first turn just absorbing the skill.

/evaluate

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,379 @@
...
+| `[Theory(DisplayName = "x")]` | Same as above on the `[TestMethod]` |
+| `[Fact(Skip = "reason")]` | `[TestMethod]` + `[Ignore("reason")]` (the `[Ignore]` attribute alone does not discover a test -- you still need `[TestMethod]`) |
+| `[Fact(Timeout = 5000)]` | `[TestMethod]` + `[Timeout(5000)]` (same -- `[Timeout]` is a modifier, not a discovery attribute) |
+| `[Trait("Category", "Unit")]` | `[TestCategory("Unit")]` |
+| `[Trait("Owner", "alice")]` | `[TestProperty("Owner", "alice")]` |
+| `[Collection("Db")]` | Step 8 + Step 11: `[DoNotParallelize]` (serialization) + `[ClassInitialize]` (sharing) -- preserve scope explicitly |
+| Custom `FactAttribute` subclass | Custom `TestMethodAttribute` subclass overriding `ExecuteAsync` (MSTest v4). See `writing-mstest-tests` and `migrate-mstest-v3-to-v4` for `CallerInfo` constructor pattern |
+| Custom `TheoryAttribute` subclass | Same -- subclass `TestMethodAttribute`; expose data via `ITestDataSource` |
+
+> Both `[TestCategory]` and `[TestProperty]` are **filterable** at runtime and target `Assembly`, `Class`, and `Method`:
```
Evangelink: TestProperty cannot be applied at asm level (TestCategory can)

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,544 @@
...
+With the pin in `global.json`, the project line simplifies to `<Project Sdk="MSTest.Sdk">`.
+
+When switching to `MSTest.Sdk`, also remove now-redundant properties: `<OutputType>Exe</OutputType>`, `<IsPackable>false</IsPackable>`, `<IsTestProject>true</IsTestProject>`, `<EnableMSTestRunner>`.
+
+### Step 3: Update project configuration
+
+1. **Preserve the runner.** Confirm the platform decision from Step 1 still holds after Step 2. Common mistakes:
+   - Switching to `MSTest.Sdk` without `UseVSTest=true` silently flips a VSTest project to MTP. Add `<UseVSTest>true</UseVSTest>` to the project (the SDK pulls in `Microsoft.NET.Test.Sdk` automatically -- no manual `PackageReference` needed).
+   - Leaving `<UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>` in `Directory.Build.props` will keep MTP on -- that is correct only if the project was already on MTP.
+2. Remove xUnit-only properties from `.csproj` / `Directory.Build.props`: anything referencing `xunit.runner.*`, `<CaptureConsoleOutput>`, `<UseRoslynCompilers>` set just for xUnit, etc.
```
Youssef1313: I'm not sure these are real properties.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,544 @@
...
+> **Commit strategy:** Commit after Step 2 (packages updated, builds broken), after Step 6 (attributes converted, asserts fixed), and after Step 8 (fixtures/lifecycle rewritten, tests pass). Commit before fixing follow-up cleanup so reviewers can bisect.
+
+### Step 1: Assess the project
+
+1. Locate every test project. Read `.csproj`, `Directory.Build.props`, `Directory.Packages.props`, and `global.json`.
+2. Identify the **xUnit version**:
+   - `xunit` 2.x (+ `xunit.assert` / `xunit.core` / `xunit.abstractions`) -> **xUnit v2**
+   - `xunit.v3` / `xunit.v3.*` -> **xUnit v3**
+3. Identify the **current test platform** (this dictates what to keep, not what to change):
+   - `xunit.runner.visualstudio` (v2) or no MTP signals (v3) and no `<UseMicrosoftTestingPlatformRunner>` -> **VSTest**
```
Youssef1313: UseMicrosoftTestingPlatformRunner doesn't affect anything other than `dotnet run`. I don't think we should use that as a signal to assess whether a project is MTP or VSTest.

The matrix of possibilities here is a bit complicated.

- xunit.v3 inside Test Explorer is always MTP, unless opted-out (by disabling the server project capability or by using xunit.v3.mtp-off)
- xunit.v3 in dotnet test:
    - Using the VSTest-based dotnet test -> it will attempt to run with the partly-baked implementation. Except for when using .NET 10+ **and** MTP v2 which will error.
    - Using the MTP-based dotnet test via global.json -> always MTP.
- **IMPORTANT:** Please double check on the above in case I'm mis-remembering some of the details.
- All above should really be unrelated to UseMicrosoftTestingPlatformRunner.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,544 @@
...
+   - xUnit v3 with `xunit.v3.mtp-v*` / `xunit.v3.core.mtp-v*`, or `<UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>`, or `YTest.MTP.XUnit2` (v2 MTP shim) -> **MTP**
```
Youssef1313: Same here about UseMicrosoftTestingPlatformRunner.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,544 @@
...
+   - Use the `platform-detection` skill if you are unsure.
```
Youssef1313: If we already have such skill, why to re-state everything here again? Shouldn't we just suggest the agent to go to the existing skill?

Please also double check that the existing skill is correct with regards to the points I mentioned above.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,544 @@
...
+
+**Option A -- `MSTest` metapackage (recommended for incremental migrations):**
+
+```xml
+<ItemGroup>
+  <PackageReference Include="MSTest" Version="4.1.0" />
+</ItemGroup>
+```
+
+The `MSTest` metapackage pulls in `MSTest.TestFramework`, `MSTest.TestAdapter`, `MSTest.Analyzers`, and `Microsoft.NET.Test.Sdk` -- so VSTest discovery (`vstest.console`, classic `dotnet test`) still works.
```
Youssef1313, line 116: The metapackage is commonly problematic for MTP due to code coverage issues that I don't know if they have been fixed.

In short, Microsoft.NET.Test.Sdk brings VSTest's Microsoft.CodeCoverage transitively and that can mess up things in certain situations.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,544 @@
...
+
+With the pin in `global.json`, the project line simplifies to `<Project Sdk="MSTest.Sdk">`.
+
+When switching to `MSTest.Sdk`, also remove now-redundant properties: `<OutputType>Exe</OutputType>`, `<IsPackable>false</IsPackable>`, `<IsTestProject>true</IsTestProject>`, `<EnableMSTestRunner>`.
+
+### Step 3: Update project configuration
+
+1. **Preserve the runner.** Confirm the platform decision from Step 1 still holds after Step 2. Common mistakes:
+   - Switching to `MSTest.Sdk` without `UseVSTest=true` silently flips a VSTest project to MTP. Add `<UseVSTest>true</UseVSTest>` to the project (the SDK pulls in `Microsoft.NET.Test.Sdk` automatically -- no manual `PackageReference` needed).
+   - Leaving `<UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>` in `Directory.Build.props` will keep MTP on -- that is correct only if the project was already on MTP.
```
Youssef1313: I'm not sure what this wants to say.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,379 @@
...
+> - `[TestCategory("Unit")]` -> `--filter "TestCategory=Unit"` (VSTest) / `--filter-trait "TestCategory=Unit"` (MTP)
+> - `[TestProperty("Owner", "alice")]` -> `--filter "Owner=alice"` (VSTest) / `--filter-trait "Owner=alice"` (MTP)
+>
+> Use `[TestCategory]` for the conventional category trait; use `[TestProperty]` for arbitrary key/value metadata. An `[assembly: Trait(...)]` in xUnit can be migrated to `[assembly: TestCategory(...)]` / `[assembly: TestProperty(...)]` (see Section 8).
+>
+> **Conditional skips** (xUnit `[Trait("OS", "Windows")]` patterns that gate execution): MSTest 3.10+ offers dedicated condition attributes -- `[OSCondition]`, `[CICondition]`, `[ArchitectureCondition]`, `[NonParallelizableCondition]` -- which are usually a better fit than overloading `[TestCategory]` for environmental gating. See Section 3.9.
```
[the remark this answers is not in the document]
Evangelink: Fixed in 60d21b6. Removed [ArchitectureCondition] and [NonParallelizableCondition] from the SkipWhen mapping and Common Pitfalls table -- only [OSCondition] and [CICondition] ship in MSTest 3.10+ (verified against microsoft/testfx). For non-parallel intent use [DoNotParallelize]; for architecture gating fall back to runtime Assert.Inconclusive.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
Evangelink: Fixed in 60d21b6. Verified that TestPropertyAttribute targets only [Class | Method] (no AttributeTargets.Assembly) -- [assembly: TestProperty(...)] does not compile. Updated the trait note in SKILL.md Step 4, cheatsheet section 1, Step 12 assembly attributes table, and cheatsheet section 8 to clarify that a non-category [assembly: Trait] must collapse to [assembly: TestCategory(value)] or be pushed down to per-class [TestProperty].

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
Evangelink: Fixed in 60d21b6. <CaptureConsoleOutput> is an xunit.runner.json setting (not an MSBuild property) and <UseRoslynCompilers> is unrelated. Removed both from Step 3 and now point users at Step 11 for the xunit.runner.json -> .runsettings port table.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
Evangelink: Fixed in 60d21b6. Step 1 bullet 3 no longer uses <UseMicrosoftTestingPlatformRunner> as a VSTest-vs-MTP signal. The bullet now delegates to the platform-detection skill and explicitly says the property only affects 'dotnet run', not Test Explorer / 'dotnet test'.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
Evangelink: Fixed in 60d21b6 together with the L78 thread -- the two adjacent bullets collapsed into a single platform-detection delegation.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
Evangelink: Fixed in 60d21b6. Step 1 bullet 3 now delegates to platform-detection. Re: double-checking platform-detection itself -- you're right that its current UseMTP-runner-as-signal claim is wrong; I'll address that in a separate PR scoped to that skill so this PR stays focused on the xUnit migration surface.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
Evangelink: Fixed in 60d21b6. Added an MTP code-coverage caveat to Step 2 Option A explaining that Microsoft.NET.Test.Sdk pulls in VSTest's Microsoft.CodeCoverage transitively, which can interfere with the MTP collector. Recommend Option B (MSTest.Sdk without UseVSTest) for MTP projects that need code coverage.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
Evangelink: Fixed in 60d21b6. Step 3 bullet 1 now states explicitly that <UseMicrosoftTestingPlatformRunner> only affects the 'dotnet run' entry point and is not a runner switch for Test Explorer / 'dotnet test'.

Evangelink:
Round 3 review feedback addressed in 60d21b6.

Highlights:
- Removed fictional MSTest condition attributes (only [OSCondition] and [CICondition] ship in 3.10+).
- Corrected [TestProperty] scope (Class/Method only -- not Assembly).
- Step 1 platform detection now delegates to the platform-detection skill and stops using <UseMicrosoftTestingPlatformRunner> as a runner signal (it only affects `dotnet run`).
- Step 2 Option A gained an MTP code-coverage caveat re: `Microsoft.NET.Test.Sdk` pulling VSTest's `Microsoft.CodeCoverage` transitively.
- Step 3 dropped references to non-existent xUnit MSBuild properties.

Follow-up: the platform-detection skill itself contains the same incorrect UseMTP-runner-as-signal claim. I'll fix that in a separate PR to keep this one scoped to the xUnit migration.

/evaluate

plugins/dotnet-test/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md
```diff
@@ -0,0 +1,379 @@
...
+| `Assert.DoesNotMatch("\\d+", str)` | `Assert.DoesNotMatchRegex(@"\d+", str)` |
+| `Assert.Equal("a", "A", ignoreCase: true)` | `Assert.AreEqual("a", "A", ignoreCase: true)` |
+
+### 3.6 Collection
+
+| xUnit | MSTest |
+|---|---|
+| `Assert.Contains(item, collection)` | `Assert.Contains(item, collection)` |
+| `Assert.DoesNotContain(item, collection)` | `Assert.DoesNotContain(item, collection)` |
+| `Assert.Contains(collection, x => predicate)` | `Assert.Contains(collection.First(x => predicate), collection)` or `Assert.IsTrue(collection.Any(x => predicate))` |
```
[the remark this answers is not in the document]
Evangelink: Fixed in 2d84d3f. Replaced the buggy Assert.Contains(collection.First(x => predicate), collection) translation with just Assert.IsTrue(collection.Any(x => predicate)). You're right that First() throws InvalidOperationException on no-match (changing the failure mode) and feeding it back into Assert.Contains is circular.

plugins/dotnet-test/skills/migrate-xunit-to-mstest/SKILL.md
```diff
@@ -0,0 +1,546 @@
...
+
+### Step 4: Convert test classes and methods
+
+Apply these rewrites to every C# test file. Class-level first, then method-level.
+
+**Class:**
+
+- Add `[TestClass]` to every class that contained xUnit `[Fact]`/`[Theory]` methods (xUnit had no class-level requirement).
+- Mark the class `sealed` (MSTest idiom -- see `writing-mstest-tests`).
+- Replace `using Xunit;` / `using Xunit.Abstractions;` with `using Microsoft.VisualStudio.TestTools.UnitTesting;`.
```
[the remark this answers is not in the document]
Evangelink: Fixed in 2d84d3f. SKILL.md Step 4 and cheatsheet section 1 no longer instruct mechanically sealing every class. You're right that xUnit base/derived test class patterns would break -- sealing is now framed as an optional follow-up handled by `writing-mstest-tests`, not part of the mechanical migration.

JanKrivanek (approved):
It looks solid.

It's quite huge - but that feels justified for this scenario

PR #760, opened 2026-06-12
Add authoring-github-workflows skill + actionlint CI gate (prevent workflow-YAML breakage)

YuliiaKovalova:
## Why

Eval on `main` broke for every run after #746 merged, with GitHub reporting *"This run likely failed because of a workflow file issue"* and **no jobs starting**. Root cause: an **unquoted** `${{ }}` expression in `run-name` containing `#`. In YAML, ` #` starts a comment, so the expression was silently truncated to `...format('Evaluate PR` — an unterminated `${{`. The file still parsed as valid YAML (so `yaml.safe_load` and review passed), but GitHub Actions rejected it.

This PR makes sure **we never see that class of bug again**, with two layers:

### 1. Knowledge — a skill
`.agents/skills/authoring-github-workflows/SKILL.md` (alongside the existing repo contributor meta-skills like `create-skill`) documents:
- The `#`-as-comment trap and the quoting rule for expression scalars.
- Other characters that force quoting.
- Why YAML-only validation (`yaml.safe_load`, `yamllint`) is **not** enough — only `actionlint` understands the Actions expression grammar.
- The failure signature (*"workflow file issue"*, no jobs).

### 2. Enforcement — an actionlint CI gate
`.github/workflows/actionlint.yml` runs a pinned `actionlint` on every PR/push that touches `.github/workflows/`:
- Lints only **hand-authored** workflows; generated gh-aw files (`DO NOT EDIT` headers, e.g. `*.lock.yml`, `agentics-maintenance.yml`) are skipped — they're compiled artifacts.
- Runs with `-shellcheck= -pyflakes=` so the gate targets workflow/expression correctness, not pre-existing shell/Python style noise.
- `.github/actionlint.yaml` declares the repo's custom runner labels (`ubuntu-slim`, `windows-11-arm`).

## Verification
- `actionlint` exits **0** on all 14 hand-authored workflows (including the fixed `evaluation.yml`).
- Reverting the quote reproduces `got unexpected EOF while lexing end of string literal [expression]` and `actionlint` exits **1** — confirming the gate would have caught the original bug.

## Note
This PR is **stacked on #759** (the one-line `evaluation.yml` hotfix) so the new gate is green here. The fix commit is identical, so it merges cleanly whether #759 lands first or this PR does. If you prefer, merge #759 for the urgent fix and this PR for the durable prevention.


.agents/skills/authoring-github-workflows/SKILL.md
```diff
@@ -0,0 +1,123 @@
+---
+name: authoring-github-workflows
```
JanKrivanek, line 2: There is a slight overlap of area with https://github.com/dotnet/skills/blob/main/.github/agents/agentic-workflows.agent.md- I'd point out explicitly in the frontmatter and in the body that this skill is for _syntactical_ guidance, while the https://github.com/dotnet/skills/blob/main/.github/agents/agentic-workflows.agent.md is for _semantical and functional_ guidance. That should help those 2 working nicely together

JanKrivanek (approved):
Thanks!

PR #767, opened 2026-06-15
code-testing-agent: pin down behavior in generated tests

Evangelink:
## Summary

Adds explicit guidance for producing tests that **pin down behavior**, not just exercise code paths. Coverage is necessary but not sufficient — a passing test that does not actually verify the contract gives false confidence. This PR teaches the `code-testing-agent` skill and `code-testing-implementer` agent to write tests that would fail under a plausible code mutation, with eval rubrics that check it.

## Motivation

Comparing two recent runs of a polyglot test-generation benchmark (44 instances across Python, Go, JS, Ruby, C), the `code-testing-generator` agent matched the stock generic agent on whether tests *compiled and passed* (mutation_pass) but underperformed on whether tests actually verified the spec (rubrics_pass). Inspection of the produced tests revealed three recurring failure modes — all language-agnostic and well-known in testing literature:

1. **Orthogonal decomposition.** When a function handled quoting, escaping, and spaces, the agent wrote one test per property and never one test combining them. The bugs lived in the intersection.
2. **Degenerate fixtures.** The agent set the parameter *under test* to a no-op value (e.g. `scrollback=0` while testing scroll history, `capacity=1` while testing eviction), then asserted on the resulting noop. Tests passed but proved nothing about the feature being tested.
3. **Trivial assertions.** `Assert.IsNotNull(result)`, `expect(x).toBeDefined()`, `assert result is not None` — only the existence of a return value was checked. Side-effects, related state, and the actual transformation went unasserted.

These are universal anti-patterns documented in mutation-testing literature (Offutt et al.), the observability/behavior-radius literature (Binder, *Testing OO Systems*, 1999), and combinatorial-testing literature (Kuhn et al.). They affect any test-generation workflow in any language.

### Skill prompt (the conventions doc loaded by every code-testing-agent invocation)

`plugins/dotnet-test/skills/code-testing-agent/unit-test-generation.prompt.md` — new section `Write Tests That Pin Down Behavior` between `Quality over Quantity` and `Parameterization`. Five named principles, ~40 lines, no language-specific assertion APIs:

* **Mutation thinking** — would the test fail if the code had a plausible bug?
* **Property intersections** — exercise combinations, not only coordinate axes
* **Behavior radius** — assert on at least one secondary observable per test
* **Fixture realism** — never set the parameter under test to a degenerate value
* **Quick self-review** before declaring a test method done

### Implementer agent (the sub-agent that actually writes test files)

`plugins/dotnet-test/agents/code-testing-implementer.agent.md` — new `Test depth (cross-language invariants)` subsection appended to Step 4, mirroring the same five principles at the point in the workflow where test files are produced. Cross-references the prompt file for the full discussion.

### Eval rubrics (so we can detect regressions)

`tests/dotnet-test/code-testing-agent/eval.yaml` and `tests/dotnet-test/code-testing-agent/eval.vally.yaml` — adds depth-oriented rubric items to each of the three existing scenarios:

* **ContosoUniversity (.NET)** — minimal `IsNotNull`-only assertions; per-controller secondary-observable check on the underlying `SchoolContext`
* **python-flask-tasks (Python)** — minimal `is not None`-only assertions; at least one combined-property TaskService validation test
* **typescript-vitest-cart (TypeScript)** — minimal `toBeDefined`/`toBeTruthy`-only assertions; at least one intersection test combining discount + tax + shipping

The rubric additions are LLM-judged, consistent with the rest of the scenario rubrics, and require no new fixtures.

## Non-goals

* Not benchmark-specific — every change is framed in universal test-craft principles. No mention of any specific eval, prompt, or rubric vocabulary outside of the eval files themselves.
* Not a new skill — leverages files that already exist in every code-testing-agent execution path.
* Not changing the existing rules (happy/edge/error paths, mock externals) — depth is *additive* on top of them.
* Not modifying `code-testing-generator.agent.md` — that file gets a *separate* PR for the pre-completion self-review gate (test-gap-analysis + assertion-quality), which is a different concern.

## Validation

* `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin plugins/dotnet-test` — passes; the SKILL.md token count for code-testing-agent is unchanged (the new content lives in a separate file in the skill bundle).

### DIFF
--- a/plugins/dotnet-test/agents/code-testing-implementer.agent.md
+++ b/plugins/dotnet-test/agents/code-testing-implementer.agent.md
@@ -65,15 +65,7 @@ These rules apply to every language and override any pattern an existing test fi
 
 #### Test depth (cross-language invariants)
 
-Coverage is necessary but not sufficient — a test that exercises code without verifying its behavior gives false confidence. Every test you write must *pin down behavior* so it would fail under a plausible bug. Apply the principles in the `code-testing-agent` skill's `unit-test-generation.prompt.md` → "Write Tests That Pin Down Behavior" section:
-
-- **Mutation thinking** — each assertion would fail under at least one plausible code mutation (`>` flipped to `>=`, a `null`/`None`/`nil` check removed, an `&&` swapped for `||`, an off-by-one boundary). If the test would still pass under such a mutation, the assertion is too weak — replace `IsNotNull`/`toBeDefined` with a concrete expected value or add a follow-up assertion that pins the contract.
-- **Property intersections** — when the code under test handles multiple independent properties (quoted vs. unquoted, ASCII vs. escaped, present vs. absent), add at least one test that exercises **several properties together**, not only one test per property. Bugs live at intersections.
-- **Behavior radius** — assert on at least one *secondary* observable per test (related state, log output, neighboring fields, history buffer, retry counter, event subscription), not only the primary return value. Bugs where the return looks right but a side-effect contract is broken require this.
-- **Fixture realism** — never set the parameter under test to a degenerate value: don't test scroll with `scrollback=0`, eviction with `capacity=1`, retries with `maxRetries=0`, or iteration ordering with a single-element collection.
-- **Avoid tautological assertions** — never assert that a value you just wrote can be read back unchanged on an identity/round-trip operation; that proves only that storage works, not that the code under test produces the right value. Assert on the *transformation* the code is supposed to perform.
-
-These rules apply to every language and framework. They are not a substitute for the existing rules above (happy/edge/error paths, mock external dependencies) — they are a depth requirement that sits on top of them.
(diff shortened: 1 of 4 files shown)


JanKrivanek (commented):
This overlaps with guidance in https://github.com/dotnet/skills/pull/768
While it is in different agent/files and has more details - it might be overly verbose (compared to the whole length and style of the edited files).

tl;dr; - can it be made much more conscise?

Evangelink:
@JanKrivanek good point — condensed in fa1433c44. The five multi-paragraph subsections in `unit-test-generation.prompt.md` are now a single compact bullet list (mutation thinking, no tautologies, property intersections, behavior radius, fixture realism) plus a one-line self-review, matching the terse style of the surrounding file. The `code-testing-implementer.agent.md` block no longer restates it all — it's a short pointer to that section, which also removes the overlap with #768. Net: +8/-45 lines.

JanKrivanek (approved):
Thanks for iterating on this!

PR #796, opened 2026-06-20
MSRC validation: contained Vally token proof

rootd4ddy:
Closed by reporter. Validation path did not produce the expected approved PoC evidence.


rootd4ddy:
@dotnet-policy-service agree

PR #836, opened 2026-06-26
Add TFM-aware buildTransitive->build forwarding guidance to msbuild skills

YuliiaKovalova:
## Summary

Captures the lesson learned in microsoft/testfx#9431 about TFM-aware `buildTransitive/` → `build/` forwarding.

A NuGet build-extension `buildTransitive/*.props` should forward through the corresponding `build/*.props` (ownership chain `buildTransitive → build → shared`) rather than importing `buildMultiTargeting/` directly. The subtlety: `build/*.props` are frequently packed **per-TFM** under `build/<tfm>/`, while `buildMultiTargeting/` is not. So a forwarder living in `buildTransitive/<tfm>/` **must include the TFM segment** — and must derive `<tfm>` from the file's own folder, **not** `$(TargetFramework)`, because NuGet nearest-match can serve a consumer a different asset folder. Dropping the segment resolves to a non-existent package-root `build/Pkg.props` and fails with `MSB4019` for transitive consumers.

## Changes

- **extension-points/SKILL.md** — new "Forwarding chain: `buildTransitive/` → `build/` → shared" section, including the TFM-segment requirement and the derive-from-own-folder expression.
- **msbuild-antipatterns/SKILL.md** (AP-13) — added the forwarding/TFM note (the other relevant build rule).
- **msbuild-code-review.agent.md** — added a Category-4 review check for the forwarding chain + TFM-correct forwarding.

Reference: microsoft/testfx#9431 (Rule E-1).


YuliiaKovalova:
Addressed the review feedback in d38fb2e: replaced the ambiguous `buildTransitive/*.props|targets` with the clearer ".props/.targets forwarders" wording.

YuliiaKovalova:
## Eval analysis + quality iteration

I downloaded the eval artifacts (run 28239731107) and analyzed `results.json`. Summary of findings and the fix I pushed (`b45beec`).

### The failing verdicts are noise/overhead-driven, not content regressions
The **pairwise A/B judgments** (the more reliable signal — direct comparison of baseline vs skilled output) favor the skill on both "failing" scenarios:

- **extension-points / "Fix extension point anti-patterns"**: `overallWinner: "tie"` — *"Both responses identified the same issues, applied the same fixes… essentially identical."* The absolute scores (baseline 5.0 → skilled 4.3) are judge scoring variance — the skilled run's own rubric reasoning is all positive ("canonical patterns", "clear and correct"). This is **pattern #8 (baseline already good, no headroom)** + **#7 (judge regression on close calls — "usually noise, re-run")** from `InvestigatingResults.md`. `perRunScores` CV = 81%.
- **msbuild-antipatterns / "Review MSBuild files"**: `overallWinner: "skill"` — the skilled output actually **won** ("Response B is slightly better… checked the fsharp subdirectory for completeness"). CV = 102%.
- The F# "NOT ACTIVATED" scenarios fail purely on the **token penalty** (pattern #4) — inherent skill-loading overhead, weighted only 0.05.

### What I changed
Even though the dips are noise, the `extension-points` "Forwarding chain" addition is *irrelevant* to the anti-pattern-fixing scenario, so its length was pure overhead/distraction there. I **tightened it from +26 to +11 net lines** (and trimmed the AP-13 note) — dropping the redundant non-TFM example and the self-evident derivation explanation, while keeping the chain rule, the `MSB4019` cause, and the TFM-derivation expression. This directly reduces the token overhead flagged by pattern #4 without losing substance.

### Recommendation
Given CV of 81–102%, a single run is dominated by LLM non-determinism. Re-running with more runs should stabilize the signal (the guide suggests 7–10 for noisy scenarios).

/evaluate --runs 7

YuliiaKovalova:
Addressed both review nits in 0b7f356: updated the range to "AP-16 through AP-22" (the reference doc now includes AP-22), and switched the prose to a forward-slash `build/MyPackage.props` to match the `build/<tfm>/` convention in that section.

YuliiaKovalova:
### Re: the three F# `msbuild-antipatterns` scenarios (footnotes [2]/[3]/[4])

These three are **pre-existing, structural** failures — not regressions from this PR, and not fixable by its content:

- **All three are `NOT ACTIVATED` with unchanged quality** (`5.0 → 5.0`, `4.0 → 4.0`). `msbuild-antipatterns` correctly does **not** activate for F# coding tasks (add a module / fix file order / add a signature file). That's the desired behavior.
- They fail **only** on the token footnote (`-0.6%` / `-4.1%` / `-4.3%`) — the inherent cost of loading the plugin's skill menu in the skilled arm vs the no-skill baseline. The `~50k`-token delta is the plugin/skill footprint; this PR's forwarding note is a negligible fraction of it.
- **They were already ❌ in the first eval on `d38fb2e`** (before any of my tightening), with the identical NOT-ACTIVATED + token pattern. They fail the same way on `main`.

What I *did* do to minimize footprint (this lever is mine to pull, even though it won't flip the deterministic token-penalty rows):
- Tightened the `extension-points` "Forwarding chain" section (+26 → +11 net lines).
- Condensed the AP-13 note to a one-line cross-reference (`8be4ea9`), removing duplication with `extension-points`. Net result: scenario [2]'s skilled token count already dropped (`167k → 158k`, footnote moved `-3.8% → -0.6%`).

The remaining red on [3]/[4] is deterministic token overhead on control scenarios where the skill correctly stays dormant — it can only be moved by shrinking the **overall** plugin's skill-menu footprint (out of scope here) or by the harness's token weighting / run count. Re-running won't help these two (they're not variance-driven). Flagging for maintainer awareness.

Evangelink (approved):

PR #843, opened 2026-06-29
Port dotnet-blazor evals to Vally format

javiercn:
## Summary

Ports all 9 `dotnet-blazor` skill evaluations to the newer **Vally** format (`eval.vally.yaml`), bringing the plugin in line with the already-migrated `dotnet-*` plugins. The legacy `eval.yaml` files are kept alongside (dual-format), matching how other migrated plugins operate during the transition.

Also registers a `dotnet-blazor` suite in the root `.vally.yaml` so the shadow Vally pipeline discovers these specs.

## Skills ported (27 stimuli total)

| Skill | Stimuli |
|---|---|
| author-component | 5 |
| collect-user-input | 2 |
| configure-auth | 2 |
| coordinate-components | 2 |
| create-blazor-project | 3 |
| fetch-and-send-data | 2 |
| plan-ui-change | 5 |
| support-prerendering | 2 |
| use-js-interop | 4 |

## Conversion details

Prompts, assertions, and rubric items are preserved verbatim. Mapping applied:

- `scenarios` -> `stimuli`; per-scenario `timeout` -> top-level `config.timeout` (max across the file).
- Assertions -> graders: `output_contains` -> `output-contains` (config.substring), `output_matches` -> `output-matches` (config.pattern), `file_contains` -> `file-contains` (config.path + value), `file_exists` -> `file-exists` (config.path); each stimulus also gets `prompt` and `pairwise` graders.
- Legacy `setup.commands` (the `dotnet new blazor ...` scaffolding) -> `environment.commands`.

## Verification

- All 9 files parse as YAML and pass `vally lint` (only the standard "scoring defaults applied" and `config` deprecation warnings shared by every existing eval).
- Parity check confirms 0 dropped scenarios/assertions/rubric items and identical prompts vs the legacy specs.
- Ran the `author-component` eval end-to-end (baseline + skilled + adapt). It passed the shadow improvement threshold, confirming the converted spec executes correctly through the full Vally -> adapter pipeline.

### DIFF
--- a/.vally.yaml
+++ b/.vally.yaml
@@ -28,6 +28,11 @@ suites:
     evals:
       - "tests/dotnet-aspnetcore/*/eval.vally.yaml"
 
+  dotnet-blazor:
+    description: Blazor evals
+    evals:
+      - "tests/dotnet-blazor/*/eval.vally.yaml"
+
   dotnet-data:
     description: Data/EF Core evals
     evals:
--- a/tests/dotnet-blazor/author-component/eval.vally.yaml
+++ b/tests/dotnet-blazor/author-component/eval.vally.yaml
@@ -1,0 +1,246 @@
+name: author-component
+description: Evaluates the dotnet-blazor/author-component skill
+type: capability
+config:
+  timeout: 3m
+stimuli:
+  - name: Author a data-loading search component
+    prompt: |
+      Write a Blazor component called ProductSearch that will be used on a catalog browsing page. Here are the service contracts the application already registers in DI:
+
+      ```csharp
+      public record Product(int Id, string Name, decimal Price, string ImageUrl);
+
+      public interface IProductService
+      {
+          /// <summary>
+          /// Searches products by query string. Throws on network failure.
+          /// Respects the cancellation token — callers should cancel when results are no longer needed.
+          /// </summary>
+          Task<List<Product>> SearchAsync(string query, CancellationToken ct);
+      }
+      ```
+
+      The parent page will use the component like this:
+
+      ```razor
+      <ProductSearch OnProductSelected="HandleSelection" />
+      ```
+
+      The component should:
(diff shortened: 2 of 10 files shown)


javiercn:
@AbhitejJohn quick question on the **Skill Coverage Report** bot here. This PR only ports the `dotnet-blazor` evals to the Vally format (`eval.vally.yaml`) and doesn't touch any `SKILL.md` or the legacy `eval.yaml` files. The coverage workflow fires because the PR edits `tests/**`, and it's reporting low coverage (0–20%) with a bunch of uncovered `[CodePattern]` items (e.g. `[Parameter]`, `[Authorize]`, `[CascadingParameter]`).

From what I can tell, `Measure-SkillCoverage.ps1` analyzes the legacy `eval.yaml` assertions/rubric, not the new `eval.vally.yaml`, so these gaps look pre-existing rather than something this PR introduced. A few questions:

1. Should I just ignore this report for a pure eval-format-port PR like this one?
2. Is the coverage tool expected to learn about `eval.vally.yaml` as part of the skill-validator → Vally transition (otherwise coverage will read as artificially low for every migrated plugin)?
3. **Are we expected to port existing evals to Vally?** I did `dotnet-blazor` here (it was the only plugin still on legacy-only `eval.yaml`), but I want to confirm whether porting the remaining/legacy evals is the intended direction, or whether new evals should be authored in Vally going forward while existing ones stay as-is.

Thanks!

AbhitejJohn (approved):

AbhitejJohn:
@javiercn : Thanks for adding vally evals. Yes we'd want to have vally evals for all our plugins. We should be able to move to Vally soon. At the moment, if we can keep the eval scripts for both Vally and Skill Validator similar that would be easier when we migrate and disable Skill Validator.

PR #847, opened 2026-06-30
Enable VS Code subagent fan-out for the dotnet-test code-testing agents

0101:
## Problem

The `dotnet-test` plugin's `code-testing-*` agents implement a Research -> Plan -> Implement pipeline that delegates to subagents. In VS Code, the published agents carried only `name` / `description` / `license` frontmatter — no subagent metadata (`tools:` / `agents:`) — so VS Code never delegated and `code-testing-generator` ran the whole pipeline inline. (The GitHub Copilot CLI was unaffected.)

## Changes

All under `plugins/dotnet-test/`:

- **Orchestrators** (`code-testing-generator`, `code-testing-implementer`): add an `agents:` list of the subagents they delegate to, plus an explicit `tools: ["agent", "read", "search", "edit", "execute"]` that includes the `agent` (runSubagent) tool.
- **Workers** (`code-testing-researcher`, `code-testing-planner`, `code-testing-builder`, `code-testing-tester`, `code-testing-fixer`, `code-testing-linter`): add an explicit role-scoped `tools: ["read", "search", "edit", "execute"]`.
- Soften the generator's three `runSubagent({ agent, prompt })` blocks to tool-agnostic delegation prose.
- `README.md`: document the per-user VS Code setting `chat.subagents.allowInvocationsFromSubagents` (off by default) needed for the nested `code-testing-implementer` -> builder/tester/fixer/linter layer; the CLI has no such gate.

### Why explicit tool aliases, not `tools: ["*"]`

VS Code has no all-tools wildcard for `tools:`, and a subagent's `tools:` **overrides** (not extends) the toolset it would otherwise inherit — so a catch-all entry would leave subagents with no file/terminal access in VS Code. The named portable aliases (`read`, `search`, `edit`, `execute`, `agent`) map to real tools in **both** VS Code and the Copilot CLI; `agents:` is ignored by the CLI.

## Validation

- **VS Code (Insiders 1.127):** selecting `code-testing-generator` on a multi-module Python project produced real fan-out — `runSubagent` to `code-testing-researcher` -> `code-testing-planner` -> `code-testing-implementer`, with the subagents performing their own file I/O (`research.md`, `plan.md`, test files). 72 generated tests pass; no "subagents lack file tools" warnings.
- **Copilot CLI:** the same plugin via `--plugin-dir` delegated to the `code-testing-*` subagents (researcher / planner / implementer / tester), skills still loaded (`code-testing-extensions`, `test-gap-analysis`), and the generated tests pass. An all-tools baseline run used only tools within the enumerated set, confirming no CLI capability is restricted.

### DIFF
--- a/plugins/dotnet-test/README.md
+++ b/plugins/dotnet-test/README.md
@@ -101,6 +101,14 @@ These are pipeline stages invoked automatically by the agents above (`user-invoc
 | **code-testing-fixer** | code-testing-implementer | Fixes compilation errors in source or test files |
 | **code-testing-linter** | code-testing-implementer | Runs code formatting and linting |
 
+> **VS Code — enabling full multi-level fan-out:** The pipeline delegates in two levels: `code-testing-generator` → researcher / planner / implementer, and `code-testing-implementer` → builder / tester / fixer / linter. VS Code gates *nested* delegation (a subagent spawning its own subagents) behind a setting that is **off by default**, so the first level runs out of the box but the second one does not. For large scopes — many files or modules, where parallel build/test/fix/lint workers help — enable it in your VS Code settings:
+>
+> ```jsonc
+> "chat.subagents.allowInvocationsFromSubagents": true
+> ```
+>
+> Without it, `code-testing-implementer` still builds, tests, fixes, and lints — it just does that work inline instead of delegating to the worker subagents, so results are unaffected. The GitHub Copilot CLI has no such gate and always fans out.
+
 ## Prerequisites
 
 ### For polyglot skills and agents
--- a/plugins/dotnet-test/agents/code-testing-builder.agent.md
+++ b/plugins/dotnet-test/agents/code-testing-builder.agent.md
@@ -6,6 +6,7 @@ description: >-
   errors, verifying project builds successfully.
 name: code-testing-builder
 user-invocable: false
+tools: ["read", "search", "edit", "execute"]
 license: MIT
 ---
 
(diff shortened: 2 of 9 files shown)


0101:
Good catch — addressed in 6bd7c07.

You're right that the `tools:` allowlists omitted the `skill` tool while every `code-testing-*` agent's prompt instructs calling skills (`code-testing-extensions` for per-language guidance, plus `test-gap-analysis` / `assertion-quality`). Since `tools:` is an override, that omission can stop the agents from loading those skills.

I verified this empirically:
- **VS Code:** in my earlier run the transcript showed the agents had only `read` / `search` / `edit` / `execute` tools and **no skill tool or skill advertisement** — the `code-testing-extensions` skill never loaded (the run produced passing tests only because the target was trivial Python).
- **Copilot CLI:** the `skill` tool *was* invoked even without being listed, so the CLI was unaffected — but adding it is correct and harmless there.

**Fix:** added `skill` to all eight agents:
- orchestrators (`code-testing-generator`, `code-testing-implementer`): `["agent", "skill", "read", "search", "edit", "execute"]`
- workers (researcher/planner/builder/tester/fixer/linter): `["skill", "read", "search", "edit", "execute"]`

Re-verified in the CLI after the change: full fan-out (researcher → planner → implementer/tester), `code-testing-extensions` is invoked, and the generated tests pass (126/126 on the multi-module sample). A VS Code re-validation is underway.

Evangelink (approved):

PR #848, opened 2026-06-30
Add dotnet-aspnetcore REST API authoring skills (controllers + Minimal APIs)

javiercn:
# ASP.NET Core REST API authoring skills (`dotnet-aspnetcore`)

Adds a disjoint set of eighteen skills that improve how the agent writes ASP.NET Core REST APIs with controllers and Minimal APIs. Each skill targets a genuine, measured gap that both a strong and a mid-tier model miss, and each ships with a convention-checking eval that proves it helps both without regressing either.

## How these were built and measured

The skills were developed gap-first: for every candidate, the agent was first measured **without** the skill on natural, underspecified prompts (the kind a developer would actually write, with no hint of the convention), then **with** it. A skill is kept only when it helps both models with no meaningful regression; where the model is already strong unaided, no skill is added.

- **Domain**: a realistic ARM-style control-plane API (`MessagingApi`: a messaging namespace aggregate owning queues, topics, subscriptions, authorization rules, with soft delete, row versions, and last-modified timestamps). None of this domain appears in any skill body; skills stay domain-agnostic.
- **Models**: every A/B was run on **Claude Sonnet** and **GPT-5.4**, 3 runs per arm (5 for the borderline case), scored 1-5 per rubric criterion by an LLM judge.
- **Prompts never name the skill or cue the convention**, so the baseline is honest. Rubrics check outcomes, not the skill's vocabulary.

## Skills added (18) and eval results

Mean overall score (1-5), baseline → with-skill, natural prompts:

| Skill | Track | Sonnet | GPT-5.4 | Gap it closes |
| --- | --- | --- | --- | --- |
| `controller-concurrency` | controller | 3.0 → **5.0** | 2.5 → **4.5** | Dual validators: adds the time-based `Last-Modified` validator (`If-Modified-Since`/`If-Unmodified-Since`) alongside the `ETag` both models already did |
| `minimal-api-concurrency` | minimal | 2.5 → **5.0** | 2.83 → **4.33** | Same dual-validator gap, Minimal API form; ordinal header comparison |
| `structure-api-business-logic` | shared | 3.0 → **5.0** | 3.5 → **4.83** | Move logic into an HTTP-agnostic service returning a `Result` with an error kind (no exceptions/tuples across the service boundary); thin endpoints |
| `author-controller-endpoints` | controller | 3.56 → **4.67** | 3.78 → **4.67** | `ActionResult<T>`, correct status codes, `CreatedAtAction` + `Location`, `ProducesResponseType`, operation exposure |
| `author-minimal-api-endpoints` | minimal | 2.67 → **5.0** | 3.17 → **4.33** | Strongly typed `Results<>` unions, `TypedResults`, `MapGroup`, 201+`Location`, `ValidationProblem` |
| `controller-data-access` | controller | 3.67 → **5.0** | 4.0 → **4.83** | Keyset pagination over a deterministic total order (unique-key tiebreaker), next page as an RFC 8288 `Link` header, `AsNoTracking` + projection |
| `minimal-api-data-access` | minimal | 4.0 → **5.0** | 3.67 → **4.83** | Keyset pagination over a deterministic total order (unique-key tiebreaker), next page as an RFC 8288 `Link` header, `AsNoTracking` + projection |
| `test-apis-with-webapplicationfactory` | shared | 4.4 → **5.0** | 4.6 → **4.8** | Drive the real pipeline via `WebApplicationFactory`, replace the registered `DbContext`, isolate the database per test, assert status and body |
| `authorize-api-endpoints` | shared | 2.33 → **4.56** | 1.89 → **3.78** | Authorize through the framework, not inline checks: declarative policy + handler reading `context.Resource` (HttpContext) route/endpoint metadata for route/claim rules; imperative `AuthorizeAsync` with the loaded entity for ownership; named policies, OR-semantics, 401/403/404, middleware order |
| `minimal-api-parameter-binding` | minimal | 2.0 → **5.0** | 2.0 → **5.0** | Bind a custom/value type from the route or query via a static `TryParse`/`BindAsync` so parsing and validation happen at the boundary and invalid input is 400 before the handler runs |
| `minimal-api-endpoint-filters` | minimal | 1.0 → **5.0** | 2.0 → **5.0** | Use endpoint filters only for concerns that need the bound (validated) arguments or the result and that no built-in already covers: rewrite a bound argument before the handler, and transform the result (field selection); route validation to `AddValidation`, other cross-cutting work to middleware |
| `rate-limiting` | shared | 2.0 → **4.67** | 2.0 → **4.33** | Throttle with the built-in rate limiter (`AddRateLimiter`/`UseRateLimiter`) partitioned per caller (claim/API key/IP), 429 + `Retry-After`, deliberate algorithm — not a hand-rolled counter |
| `long-running-operations` | shared | 3.0 → **5.0** | 3.67 → **5.0** | Model slow work as an async operation: `202 Accepted` + a distinct pollable operation-status resource (running/succeeded/failed), not a blocked request |
| `change-tracking-delta` | shared | 2.17 → **4.33** | 2.17 → **4.0** | Incremental sync: fetch only added/updated/removed since a watermark. Monotonic change marker (rowversion, order-preserving; timestamp fallback), soft-delete tombstones, total order + `MIN_ACTIVE_ROWVERSION` boundary, `multipart/mixed` of `application/http` (200 upsert / 204 removed) keeping payloads unchanged, `Link: rel="deltaLink"`, 410 on stale token |
| `structured-logging` | shared | 3.0 → **5.0** | 3.0 → **4.0** | Message templates with named placeholders (not interpolation), `BeginScope` correlation, levels by severity, exception-as-argument, no secrets/PII |
| `patch-partial-updates` | shared | 3.67 → **5.0** | 3.67 → **4.67** | PATCH that changes only sent fields and distinguishes explicit null (clear) from absent (leave), via JSON Merge Patch over `JsonElement`/`JsonNode` (or `JsonPatchDocument` / tri-state `Optional`) |
| `design-collection-api` | shared | 2.33 → **4.67** | 3.0 → **5.0** | Design a complete collection API and reconcile it with the data model: decide each cross-cutting axis (stable order, pagination, filtering, field selection, delta-readiness, concurrency, status codes) and add the columns/indexes each needs (order key + id tiebreaker, rowversion watermark, soft-delete, concurrency token), routing each capability to its own skill. Closes the both-model gap where a naive "list the resource" request omits filtering and change tracking and ignores the persistence columns they require |
| `filter-and-select` | shared | 3.33 → **4.33** | 4.0 → **4.33** | Bounded, opt-in filtering (allow-list + Stripe-style bracket operators) and field selection (sparse fieldsets) translated to EF `Where`/`Select`: `400` on unknown field/operator, operands coerced to the field type, the keyset sort key kept in the projection so paging stays stable, and amplification guards (gate leading-wildcard `Contains` in favor of sargable `StartsWith`, cap predicate count and page size) |

No rubric criterion regresses on either model for any kept skill.

### Note on the authorization skill

`authorize-api-endpoints` was added after a second, probe-driven round. Both models default to inline `if (User…)` checks and do not reach for the authorization framework at all: the baseline for the framework criteria (requirement + handler, named-policy opt-in, imperative `AuthorizeAsync` for loaded-entity rules, `context.Succeed` OR-semantics) scored ~1/5 on both models, making this the largest single gap found. The skill leads with the rule of thumb that decides the mechanism: **where does the decision data live?** Route values and claims are decided declaratively by a handler that reads `context.Resource` as the `HttpContext` (and custom endpoint metadata); stored entity state (an `OwnerId`) is decided imperatively with `IAuthorizationService.AuthorizeAsync` after loading.

### Note on de-cueing the prompts

`minimal-api-parameter-binding` and `minimal-api-endpoint-filters` were initially measured as "no gap" because their probe prompts **cued the technique** (for example "use a strongly typed value in the handler signature" or "apply this across the route group without repeating"). A cued prompt hands the baseline the same guidance the skill provides, so the baseline looked strong (5.0/5.0). Rewriting the prompts to describe only the user-facing goal collapsed the baselines to ~2.0, exposing genuine both-model gaps that the skills then close. Every prompt in these evals describes a goal, never a technique the skill teaches.

## Skills considered and not built (14)

Every candidate below was measured the same way (natural prompts, both models, LLM-judged rubric). A skill is kept only when it helps **both** models with no meaningful regression. These were dropped; this list is a record so they are not re-attempted without new evidence.

### Already strong on both models — no gap to close

A skill here would only spend context tokens and risk regressing a model that is already at or near ceiling.

| Candidate | Sonnet | GPT-5.4 | What it would have taught / why dropped |
| --- | --- | --- | --- |
| `design-api-operations` | 5.0 | 4.0 | Resource-operation modeling: nesting child resources under aggregates, lifecycle transitions as explicit intent-revealing operations. Sonnet at ceiling; GPT-5.4 only scattered modest gaps. Distinct from the shipped `design-collection-api`, which covers cross-cutting collection completeness and data-model reconciliation, not resource modeling |
| `model-controller-payloads` | 5.0 | 5.0 | Request/response DTOs, avoiding entity leakage, over-posting, and navigation cycles, projecting summaries. Done unprompted |
| `model-minimal-api-payloads` | 5.0 | 4.83 | Same, Minimal API form. Payload modeling already strong |
| `options-validation` | 5.0 | 4.67 | `IOptions<T>` with `ValidateDataAnnotations` / `ValidateOnStart`. Already applied |
| `transactions-outbox` | 5.0 | 5.0 | Transactional outbox / reliable dispatch of side effects. Produced unprompted from "notify reliably even when the downstream is sometimes unavailable" |
| `polymorphic-json` | 5.0 | 5.0 | `[JsonPolymorphic]` / `[JsonDerivedType]` with an explicit discriminator. Already done |
| `health-checks` | 5.0 | 4.67 | `AddHealthChecks` with liveness/readiness endpoints. Already done |
| `opentelemetry` | 5.0 | 5.0 | Vendor-neutral OpenTelemetry tracing/metrics wiring. Already done without a vendor-neutral hint |
| `http-header-forwarding` | 5.0 | 4.67 | `ForwardedHeaders` middleware when running behind a reverse proxy. Already done |
| `consume-apis` (.NET client) | 4.67 | 4.0 | Client-side consumption: follow `rel=next` to completion, persist the delta link and resync on `410`, send `If-Match` and handle `412`, honor `Retry-After` on `429`, poll `202`, use `IHttpClientFactory`. Strong when the contract is spelled out **and** when it is only discoverable in an API-reference doc with a goal-only task (de-cued), so the high score is genuine competence, not cueing |
| `consume-apis` (JS client) | 4.5 | 4.33 | Same conventions from a `fetch` client (`AbortController`, backoff, `Link` parsing). Already strong |

### Asymmetric or borderline — one skill cannot help both models

A real gap exists on only one model (the other is at ceiling), or the two models are weak on different halves of the problem, so a single skill cannot help both without regressing one.

| Candidate | Sonnet | GPT-5.4 | Why dropped |
| --- | --- | --- | --- |
| `http-resilience` | 5.0 | 2.5 | Polly / `Microsoft.Extensions.Http.Resilience` retry, timeout, circuit breaker. A real GPT-5.4 gap, but Sonnet is at ceiling, so a skill risks regressing Sonnet |
| `api-error-handling` | 2.33 | 4.33 | Consistent `ProblemDetails` via `IExceptionHandler` / `UseExceptionHandler`. A real Sonnet gap, but GPT-5.4 is already strong. Asymmetric |
| `ef-bulk-query-shaping` | 3.5 | 3.67 | Bulk `ExecuteUpdate`/`ExecuteDelete` and `AsSplitQuery` to avoid cartesian explosion. Each model is strong on the half the other is weak on, so there is no consistent both-model gap |

## Cross-model validation (a third family: MAI-Code-1-Flash)

To check the skills are not overfit to the two frontier models they were tuned on, a spot-check ran a smaller, faster third-family model (MAI-Code-1-Flash) on a representative slate.

**The skills lift the weaker model too** (baseline → with skill, natural prompts):

| Skill | MAI-Code-1-Flash | Sonnet / GPT-5.4 (reference) |
| --- | --- | --- |
| `authorize-api-endpoints` | 1.00 → **4.20** (+3.20) | 2.33→4.56 / 1.89→3.78 |
| `minimal-api-parameter-binding` | 2.00 → **4.67** (+2.67) | 2.0→5.0 / 2.0→5.0 |
| `design-collection-api` | 2.00 → **3.33** (+1.33) | 2.33→4.67 / 3.00→5.00 |
| `filter-and-select` | 2.50 → **3.33** (+0.83) | 3.33→4.33 / 4.00→4.33 |

The weaker model starts lower and gains the most where the gap is largest (authorization, parameter binding), so the guidance is not specific to frontier models.

**The "no gap" drops are mostly robust across tiers, with one exception.** Re-measuring the dropped candidates' baselines on MAI-Code-1-Flash:

| Dropped area | MAI base | Sonnet / GPT-5.4 | Holds across tiers? |
| --- | --- | --- | --- |
| `polymorphic-json` | 4.67 | 5.0 / 5.0 | yes |
| `transactions-outbox` | 4.67 | 5.0 / 5.0 | yes |
| `model-minimal-api-payloads` | 4.83 | 5.0 / 4.83 | yes |
| `design-api-operations` | 4.17 | 5.0 / 4.0 | yes |
| `consume-apis` (.NET client) | 4.00 | 4.67 / 4.0 | yes |
| `health-checks` | 2.33 | 5.0 / 4.67 | **no — weak-model gap** |

Five of six drops hold: the weaker model also handles those areas well, so they are genuinely no-gap rather than frontier-only. The exception is `health-checks` — the frontier models add liveness/readiness checks unprompted, but MAI-Code-1-Flash mostly does not (2.33). It stays dropped for the models these skills target, but is recorded here as a **weak-model-tier candidate**: a `health-checks` skill would likely help smaller/cheaper models even though it is unnecessary for frontier ones.

## Notable findings baked into the skills

- **The genuine cross-model gaps are the non-obvious conventions**, not the basics: the dual ETag + `Last-Modified` validators, the service + `Result` structure, bounded pagination, and per-test database isolation. Endpoint and payload basics are already strong on both models.
- **Scope the eval to the skill's responsibility.** Early versions of the endpoint evals bundled domain-modeling rubrics (uniqueness, soft delete, parent existence) that the models already nail; the extra surface caused attention-theft that looked like a regression. Rescoping each eval to its skill's actual concern turned both endpoint skills into clean both-model wins.
- **Don't cue the convention in the prompt.** A cued `test-apis` prompt inflated the baseline to 5.0/4.67 and hid the real gap; a de-cued prompt exposed a real GPT-5.4 weakness in replacing and isolating the `DbContext`, which the skill then closes.
- **Two concrete skill fixes** came out of the measurements: removing a hard-delete example that cued the wrong delete semantics in `author-controller-endpoints`, and switching `test-apis` from a class-shared factory to a per-test factory so each test gets a clean in-memory database.

## Layout

- Skills: `plugins/dotnet-aspnetcore/skills/<name>/SKILL.md`
- Evals + fixtures: `tests/dotnet-aspnetcore/<name>/` (`eval.vally.yaml` + a `fixture/MessagingApi` project)

Opening as a draft for review of the skill content and the eval methodology before finalizing.


tests/dotnet-aspnetcore/structured-logging/eval.yaml
```diff
@@ -0,0 +1,25 @@
...
+    environment:
+      files:
+        - src: ./fixture/MessagingApi
+          dest: MessagingApi
+    graders:
+      - type: file-contains
+        config:
+          path: MessagingApi/**/*.cs
+          value: Logger
+      - type: prompt
```
webreidi, line 19: Add second grader:

```
- type: file-contains
        config:
          path: MessagingApi/**/*.cs
          value: BeginScope

```

tests/dotnet-aspnetcore/structured-logging/eval.yaml
```diff
@@ -0,0 +1,25 @@
...
+    rubric:
+      - Log calls use structured message templates with named placeholders (for example logger.LogInformation("... {NamespaceId} {TenantId}", id, tenant)) so the values are captured as queryable properties, not string-interpolated or concatenated into the message text
```
webreidi, line 21:
```suggestion
      - Log calls use structured message templates with named placeholders (for example logger.LogInformation("... {NamespaceId} {NamespaceName}", id, name)) so the values are captured as queryable properties, not string-interpolated or concatenated into the message text
```

tests/dotnet-aspnetcore/structured-logging/eval.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - Values needed for searching (namespace id, tenant, operation) are emitted as named log properties rather than baked into free-form text
```
webreidi, line 22:
```suggestion
      - Values needed for searching (namespace id or name, and operation) are emitted as named log properties rather than baked into free-form text; the implementation does not invent unavailable tenant data

```

tests/dotnet-aspnetcore/structured-logging/eval.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - Log lines belonging to the same request/operation are correlated, for example via a logging scope (ILogger.BeginScope) or a correlation identifier attached to the entries
+      - Log levels are used appropriately (information for normal operations, warning/error for failures) rather than everything at one level
```
webreidi, line 24:
```suggestion
      - Successful create, update, and delete operations log at Information, while expected not-found/conflict paths use Warning and unexpected exceptions preserve the exception object at Error rather than everything using one level
```

tests/dotnet-aspnetcore/structured-logging/eval.yaml
```diff
@@ -0,0 +1,25 @@
+name: structured-logging
+description: Probe — diagnosable structured logging
+type: capability
+config:
+  timeout: 16m
+stimuli:
+  - name: Diagnose production issues by namespace, tenant, operation
```
webreidi, line 7:
```suggestion
  - name: Diagnose production issues by namespace and operation
```

tests/dotnet-aspnetcore/structured-logging/eval.yaml
```diff
@@ -0,0 +1,25 @@
...
+    prompt: |
+      In the MessagingApi project (controllers), add logging to the namespace create, update, and delete operations that will genuinely help an on-call engineer diagnose problems in production. Use MessagingDbContext.
```
webreidi, line 9:
```suggestion
      In the MessagingApi project (controllers), add logging to the namespace create, update, and delete operations that will genuinely help an on-call engineer diagnose problems in production. Use MessagingDbContext. Use identifiers that actually exist in the project, such as the namespace name or id; do not invent a tenant field. Include enough operation context to distinguish create, update, and delete and correlate the lines for one request.
```

plugins/dotnet-aspnetcore/skills/structured-logging/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+  ILogger message templates with named properties instead of interpolated strings; correlating log lines
+  for one request/operation with BeginScope or a correlation id; choosing log levels; logging exceptions;
+  keeping secrets and PII out of logs; LoggerMessage source generation on hot paths.
+  DO NOT USE FOR: distributed tracing/metrics wiring (that is OpenTelemetry); global exception-to-response
+  mapping (use IExceptionHandler); audit persistence in the database.
+license: MIT
+---
+
+# Structured Logging
+
```
webreidi, line 16:
```suggestion

For controller operations, inject `ILogger<ControllerType>` and treat one action as one operation. Open a
scope after the request has been bound and the resource identity is known, then log the operation and its
outcome inside that scope. Use identifiers already present in the request or entity; do not invent a tenant,
user, or correlation field that the model does not provide.

```

webreidi:
```csharp
[ApiController]
[Route("api/namespaces")]
public sealed class NamespacesController(
  MessagingDbContext db,
  ILogger<NamespacesController> logger) : ControllerBase
{
  [HttpDelete("{name}")]
  public async Task<IActionResult> Delete(string name, CancellationToken ct)
  {
    var operationId = HttpContext.TraceIdentifier;
    using var scope = logger.BeginScope(new Dictionary<string, object?>
    {
      ["Operation"] = "DeleteNamespace",
      ["NamespaceName"] = name,
      ["OperationId"] = operationId
    });

    var entity = await db.Namespaces.SingleOrDefaultAsync(
      x => x.Name == name && !x.IsDeleted, ct);
    if (entity is null)
    {
      logger.LogWarning("Namespace not found for deletion");
      return NotFound();
    }

    entity.IsDeleted = true;
    await db.SaveChangesAsync(ct);
    logger.LogInformation("Namespace deleted");
    return NoContent();
  }
}
```

Apply the same shape to create and update: log a normal-operation event after a successful save, and log a
warning for expected misses or conflicts. Log unexpected exceptions at `Error` with the exception object and
rethrow or let the application's exception handler translate them. Do not log a full request body, tags,
connection strings, access tokens, or exception messages as template arguments.

plugins/dotnet-aspnetcore/skills/structured-logging/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+    logger.LogInformation("Charging payment");
+    // both lines carry OrderId and TenantId
+}
+```
+
+The framework already attaches request-level correlation (the trace identifier) to logs; add a scope for the identifiers that matter to *your* operation.
+
+## Levels, exceptions, and content
+
+- Use levels by severity: `Information` for normal flow, `Warning` for recoverable problems, `Error` for failures. Do not log everything at one level.
```
webreidi, line 50:
```suggestion
- Use levels by severity: `Information` for successful normal flow, `Warning` for expected misses or recoverable conflicts, and `Error` for unexpected failures. Do not log everything at one level.
```

plugins/dotnet-aspnetcore/skills/structured-logging/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+- Log an exception by passing it as the **first argument**, not by string-formatting it: `logger.LogError(ex, "Failed to create order {OrderId}", order.Id);`.
+- Never log secrets, tokens, full PII, or request/response bodies. Log identifiers and outcomes, not payloads.
```
webreidi, line 52:
```suggestion
- Put stable operation and resource identifiers in the scope or event template. A scope should not replace the event message: the event still says what happened and whether it succeeded.
- Never log secrets, tokens, full PII, or request/response bodies. Log identifiers and outcomes, not payloads.
```

webreidi (changes requested):
Overall, the plug-in is now too large, and all harnesses will ignore it because it has exceeded the context limit allowed for a plug-in. So, you'll need to review and either remove skills that aren't providing optimal value or reduce the descriptions of the skills to get within the limit without reducing their effectiveness.

PR #853, opened 2026-07-02
Document SetTargetFramework bin/obj clash on single-targeting ProjectReferences

ViktorHofer:
## Summary

Documents that `SetTargetFramework` metadata must **not** be set on a `ProjectReference` pointing to a **non-multi-targeting** (single-targeting) project.

Setting `SetTargetFramework="TargetFramework=<tfm>"` on such a reference injects `TargetFramework` as a **global property** on the referenced project. For a single-targeting project that property is **path-neutral** — the project already resolves to `bin\<config>\<tfm>\` / `obj\<config>\<tfm>\` on its own — so it does not change the output path; it only forks a *distinct* MSBuild project instance `(project, {TargetFramework=<tfm>})`. Meanwhile the solution/graph builds the same project as `(project, {})`. Both share the same `OutputPath`/`IntermediateOutputPath`, so the project is **built twice** to the same location — a bin/obj clash under parallel builds.

The P2P protocol itself correctly omits the `TargetFramework` global property for non-multi-targeting references; `SetTargetFramework` overrides that safe default and reintroduces the clash.

Motivated by https://github.com/dotnet/sdk/pull/55109#pullrequestreview-4614809313.

## Changes

- **`msbuild-antipatterns`**: new **AP-23** with BAD/GOOD examples, mechanism, detection, and when `SetTargetFramework` is legitimate (multi-targeting only); added to the quick-reference checklist; updated the SKILL.md pointer to "AP-16 through AP-23".
- **`check-bin-obj-clash`**: new "Common Causes and Fixes" subsection cross-referencing AP-23, a note on the `TargetFramework` row in the global-properties table, and an extended USE FOR clause for discoverability.


ViktorHofer:
@JanKrivanek can you please approve? I already discussed the changes with the team.

ViktorHofer:
@YuliiaKovalova @Evangelink can someone please approve this change?

YuliiaKovalova (approved):

Issue #649, opened 2026-05-14
Add dotnet11 skill for new process API's


Meir017:
based on this https://devblogs.microsoft.com/dotnet/process-api-improvements-in-dotnet-11/

AbhitejJohn:
Thanks for sharing, there's been a few skills tied to dotnet11 we've been looking at [here](https://github.com/dotnet/skills/pull/535) and [here](https://github.com/dotnet/skills/pull/610). None of them cover the process API updates though. Would be nice to have those as well - we'd love contributions. 
/cc: @SamMonoRT and @eiriktsarpalis .

SamMonoRT:
@adamsitnik - who authored the new Process APIs

adamsitnik:
Hello @Meir017 

It's great idea  and it's definitely on my TODO list for .NET 11. 

After first early adopters started using the new APIs I've received some feedback and I want to provide even few more "one-liners" like https://github.com/dotnet/runtime/issues/128453. When I get all of them merged, I am going to publish the new skill. So it won't be just "how to use the new APIs to get the job done" but also "how to solve common anti-patterns with the new APIs".

Thanks,
Adam

PR #866, opened 2026-07-07
feat(dotnet11): add process-api-net11 skill for new process APIs

shaikhsharukh:
Resolves issue #649 by adding a new skill `process-api-net11` for the new `System.Diagnostics.Process` APIs introduced in .NET 11.

This skill guides coding agents on using the new convenience one-liners, reliable deadlock-free output reading, and parent-child process lifetime properties. Evals have been included and verified using the skill-validator.


shaikhsharukh:
@dotnet-policy-service agree

rickgonzalez:
A short narrated explainer of this PR (dev testing my tool of public repos) - I hope you find it helpful. https://www.lenzon.ai/viewer/cmrb1zapc000o14os7auramuh?voice=google-chirp3

AbhitejJohn:
> 🤖 _Posted by **GitHub Copilot** on behalf of the repo maintainers._

Thanks for this contribution, @shaikhsharukh! 🎉 I verified every API in this skill against [`dotnet/runtime`](https://github.com/dotnet/runtime) `main` (which **is** the .NET 11 branch — `eng/Versions.props` shows `ProductVersion 11.0.0`; no `release/11.0` branch exists yet), and the good news is **these are all genuinely shipping .NET 11 APIs** — `Process.Run`/`RunAsync`, `RunAndCaptureText`/`Async`, `StartAndForget`, the `ReadAll*` family, and the new `ProcessStartInfo` properties. Nice job picking a real, useful new surface. 👍

Before we run evals and move toward merge, though, the `SKILL.md` documents several of these APIs inaccurately versus the [public ref assembly](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.Process/ref/System.Diagnostics.Process.cs). Since a skill trains agents to emit code, these would lead to non-compiling output. Could you correct the following?

**1. `ReadAll*` return types are wrong.** They return tuples / streaming types, not plain strings/arrays:

| Documented | Actual |
|---|---|
| `string ReadAllText()` | `(string StandardOutput, string StandardError) ReadAllText(TimeSpan? timeout = default)` |
| `byte[] ReadAllBytes()` | `(byte[] StandardOutput, byte[] StandardError) ReadAllBytes(...)` |
| `string[] ReadAllLines()` | `IEnumerable<ProcessOutputLine> ReadAllLines(...)` |
| `Task<string[]> ReadAllLinesAsync()` | `IAsyncEnumerable<ProcessOutputLine> ReadAllLinesAsync(...)` |
| `Task<string> ReadAllTextAsync()` | `Task<(string, string)> ReadAllTextAsync(...)` |

**2. `InheritedHandles` — invented type.** The doc uses `ProcessHandleInheritance InheritedHandles`, but that type doesn't exist. The real property is `IList<SafeHandle>? InheritedHandles`.

**3. Static method signatures.** `Run`, `RunAndCaptureText`, and `StartAndForget` don't take `(string arguments = "", ProcessStartInfo? startInfo = null)`. Actual: arguments is `IList<string>?` with an optional `TimeSpan? timeout` (`Run` also has a `bool silent` parameter); `StartAndForget` has no `startInfo` trailing parameter.

**4. Example 1 won't compile.** `Process.RunAndCaptureTextAsync("git", "--version")` — a bare string can't bind to `IList<string>?`. Use `["--version"]` (or pass a `ProcessStartInfo`).

**5. Example 3 won't compile.** `string[] lines = await process.ReadAllLinesAsync(); foreach (var line in lines)` — `ReadAllLinesAsync()` returns `IAsyncEnumerable<ProcessOutputLine>`, so it needs `await foreach`, and each item is a `ProcessOutputLine`, not a `string`.

For reference, Example 2 (`KillOnParentExit = true`) and the `ProcessTextOutput`/`ProcessExitStatus` usage in Example 1 are correct as written. ✅

Once the signatures and examples are aligned with the ref assembly, we'll kick off evals and continue the review. Thanks again for contributing! 🙏

shaikhsharukh:
I have aligned the skill definitions and examples with the `dotnet/runtime` ref assembly:

1. **`ReadAll*` Return Types:** Corrected the signatures to match the exact tuple return types (`(string StandardOutput, string StandardError)` for `ReadAllText`/`Async`, `(byte[] StandardOutput, byte[] StandardError)` for `ReadAllBytes`/`Async`) and streaming types (`IEnumerable<ProcessOutputLine>` / `IAsyncEnumerable<ProcessOutputLine>` for `ReadAllLines`/`Async`).
2. **`InheritedHandles` Property:** Changed the property type from the hypothetical `ProcessHandleInheritance` to `IList<SafeHandle>?`.
3. **Static Method Signatures:** Aligned the static `Run`, `RunAndCaptureText`, and `StartAndForget` methods to accept `IList<string>? arguments` and appropriate timeout/silent parameters.
4. **Example 1 Compilation:** Modified the call to pass `["status"]` as a collection expression instead of a bare string parameter.
5. **Example 3 Compilation:** Updated the ping example to use `await foreach` over `IAsyncEnumerable<ProcessOutputLine>` and access `line.Content` / `line.StandardError`.

AbhitejJohn (commented):
### Follow-up review — API accuracy re-verified + eval-hardening suggestions

Thanks for the fixes, @shaikhsharukh. I re-checked everything and ran a deeper evaluation pass. Summary below; concrete `suggestion`s are inline on `eval.yaml`.

**1. API accuracy — all confirmed :white_check_mark: (proven)**
I re-verified every signature in `SKILL.md` against the current `dotnet/runtime` `main` reference assemblies. All five items from the earlier review are now correct: the tuple-returning `ReadAllText`/`ReadAllTextAsync` → `(string StandardOutput, string StandardError)`, streaming `ReadAllLinesAsync` → `IAsyncEnumerable<ProcessOutputLine>`, `RunAndCaptureText(Async)` → `ProcessTextOutput`, and `KillOnParentExit`. Content looks accurate.

**2. Does the skill actually help? — yes, in plugin/deployment mode (proven, scoped)**
I hardened the eval to use intent-only prompts (never naming the API) and ran it with skill-validator (`--runs 5`, judge = `claude-opus-4.6`). In the **whole-plugin configuration** (how the skill actually ships), it produced large, low run-to-run variance (CV 1–6%) gains on its target APIs vs a no-skill baseline:

| Scenario | Baseline | With skill |
|---|---|---|
| Run + capture (`RunAndCaptureText`) | 1.8 / 5 | **4.2 / 5** |
| Full read (`ReadAllText` tuple) | 1.0 / 5 | **4.2 / 5** |
| Stream lines (`ReadAllLinesAsync`) | 1.4 / 5 | **4.2 / 5** |

The no-skill baseline reached for the pre-.NET-11 pattern (`StandardOutput.ReadToEndAsync()` + `Task.WhenAll`); the judge noted it *"didn't even attempt to research what new APIs .NET 11 offers."* So the skill is teaching genuinely non-obvious APIs — that's its value.

**3. Honest caveat — the blended eval verdict is _not_ green**
Being transparent: the validator's official pass metric doesn't clear the 0.10 bar (this skill scored 0.065). The plugin-mode wins above are real, but the *isolated* single-skill arm shows only small lift and the skill roughly doubles output tokens — both of which the weighted score penalizes. So it's a **strong deployment-mode signal with a failing blended score**; flagging for maintainers to weigh rather than claiming a clean pass.

**4. Suggested eval improvements (inline `suggestion`s)**
- Strengthen the run+capture assertions to require **both** `.StandardOutput` and `.StandardError` and reject the old event model.
- Add two scenarios that hit the harder, wrong-signature-prone surface — the tuple-returning `ReadAllText` and streaming `ReadAllLinesAsync` / `await foreach`. These are where the skill adds the most measurable value.
- Add a subtle `net10.0` non-activation negative (the new APIs are net11-only).
- The existing `KillOnParentExit` scenario is fine as coverage, but note it shows ~0 marginal lift — the base model already emits it correctly — so it isn't what demonstrates the skill's value.

---
cc @lewing @sammonort — would appreciate your read on point 3: the deployment-mode gains are real and stable, but the blended eval score doesn't clear the gate. Is the plugin-mode evidence enough to accept, or do you want the eval reworked to a clean green first?

tests/dotnet11/process-api-net11/eval.yaml
```diff
@@ -0,0 +1,57 @@
...
+      I want to make sure I don't run into any OS pipe deadlock issues, and I want to write it in the cleanest way possible using new .NET 11 features.
+      Show me the C# program and the `.csproj` XML targeting `net11.0` for this tool. Print both in your response.
+    assertions:
+      - type: "exit_success"
+      - type: "output_matches"
+        pattern: "Process\\.RunAndCaptureText(Async)?"
+      - type: "output_matches"
+        pattern: "net11\\.0"
+      - type: "output_not_matches"
+        pattern: "RedirectStandardOutput\\s*=\\s*true"
```
AbhitejJohn, line 20: The prompt asks for **both** stdout and stderr, so the assertions should prove both members of `ProcessTextOutput` are read — and reject the old event-based plumbing too.

```suggestion
      - type: "output_matches"
        pattern: "Process\\.RunAndCaptureText(Async)?"
      - type: "output_matches"
        pattern: "net11\\.0"
      # The prompt asks for BOTH stdout AND stderr — require both members.
      - type: "output_matches"
        pattern: "\\.StandardOutput\\b"
      - type: "output_matches"
        pattern: "\\.StandardError\\b"
      - type: "output_not_matches"
        pattern: "RedirectStandardOutput\\s*=\\s*true"
      - type: "output_not_matches"
        pattern: "BeginOutputReadLine|OutputDataReceived"
```

tests/dotnet11/process-api-net11/eval.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - type: "exit_success"
+      - type: "output_matches"
+        pattern: "KillOnParentExit\\s*=\\s*true"
+      - type: "output_matches"
+        pattern: "net11\\.0"
+    rubric:
+      - "Uses the new KillOnParentExit property on ProcessStartInfo set to true"
+      - "Targets net11.0"
+    timeout: 180
+
```
AbhitejJohn, line 45: Consider adding two scenarios that target the harder, wrong-signature-prone surface — this is where the skill demonstrably lifts quality (baseline 1.0 → 4.2 in my runs). The tuple-return and the streaming shape are the two things a model gets wrong without the skill.

```suggestion

  # --- Scenario: deadlock-free full read via instance ReadAllText (tuple return) ---
  - name: "Deadlock-free full-output read via instance ReadAllText in .NET 11"
    prompt: |
      In a .NET 11 app I start a child process myself with `Process.Start(...)` because I
      need to configure `ProcessStartInfo` first. After it starts I want to read its ENTIRE
      standard output AND standard error to completion, without the classic deadlock where
      one buffer fills while I'm draining the other. Give me the simplest first-party .NET 11
      way to get both back at once. Show a minimal program and the `net11.0` `.csproj`.
    assertions:
      - type: "exit_success"
      - type: "output_matches"
        pattern: "net11\\.0"
      - type: "output_matches"
        pattern: "ReadAllText(Async)?\\s*\\("
      # Correct: destructure the (StandardOutput, StandardError) tuple (sync or async form).
      - type: "output_matches"
        pattern: "\\)\\s*=\\s*(await\\s+)?\\w+\\.ReadAllText(Async)?\\s*\\("
      # Wrong: assigning the tuple-returning call to a single string.
      - type: "output_not_matches"
        pattern: "string\\s+\\w+\\s*=\\s*(await\\s+)?\\w+\\.ReadAllText(Async)?\\s*\\("
    rubric:
      - "Uses the new instance method Process.ReadAllText/ReadAllTextAsync"
      - "Destructures the (string StandardOutput, string StandardError) tuple — not a single string"
      - "Does not fall back to manual StandardOutput.ReadToEndAsync() as the primary mechanism"
      - "Targets net11.0"
    timeout: 180

  # --- Scenario: stream tagged output lines via ReadAllLinesAsync ---
  - name: "Stream tagged process output lines in .NET 11"
    prompt: |
      I have a long-running child process in a .NET 11 app. As it runs I want to handle each
      output line the moment it is produced — not buffer everything to the end — and for each
      line I need to know whether it came from stdout or stderr. I want the cleanest first-party
      .NET 11 approach, with no event-callback plumbing and no deadlock risk. Show a minimal
      program and the `net11.0` `.csproj`.
    assertions:
      - type: "exit_success"
      - type: "output_matches"
        pattern: "net11\\.0"
      - type: "output_matches"
        pattern: "ReadAllLines(Async)?\\s*\\("
      - type: "output_matches"
        pattern: "await\\s+foreach"
      - type: "output_matches"
        pattern: "\\.Content\\b|ProcessOutputLine"
      - type: "output_matches"
        pattern: "\\.StandardError\\b"
      - type: "output_not_matches"
        pattern: "OutputDataReceived|BeginOutputReadLine"
    rubric:
      - "Uses Process.ReadAllLinesAsync consumed with `await foreach` over IAsyncEnumerable<ProcessOutputLine>"
      - "Reads ProcessOutputLine.Content and distinguishes stdout vs stderr via .StandardError"
      - "Does NOT use the old OutputDataReceived/BeginOutputReadLine event model; targets net11.0"
    timeout: 180
```

tests/dotnet11/process-api-net11/eval.yaml
```diff
@@ -0,0 +1,57 @@
...
+      Show me a minimal C# program and the `.csproj` XML targeting `net8.0` that starts this process using the traditional Process.Start. Print both in your response.
+    expect_activation: false
+    assertions:
+      - type: "exit_success"
+      - type: "output_matches"
+        pattern: "net8\\.0"
+      - type: "output_matches"
+        pattern: "Process\\.Start\\("
+      - type: "output_not_matches"
+        pattern: "Process\\.RunAndCaptureText"
```
AbhitejJohn: Broaden the non-activation guard so it catches any of the new .NET 11 Process APIs leaking into a net8 answer, not just `RunAndCaptureText`.

```suggestion
      - type: "output_not_matches"
        pattern: "RunAndCaptureText|ReadAllTextAsync|ReadAllLinesAsync|KillOnParentExit"
```

tests/dotnet11/process-api-net11/eval.yaml
```diff
@@ -0,0 +1,57 @@
...
+    rubric:
+      - "Solves the task using standard pre-.NET 11 APIs (Process.Start)"
+      - "Does NOT load or reference the process-api-net11 skill"
+      - "Targets net8.0"
+    timeout: 180
```
AbhitejJohn, line 119: A subtle non-activation negative is worth adding: the new APIs are net11-only, so a `net10.0` capture task should keep the skill from applying them (guards against over-activation on the "capture output" intent).

```suggestion
    timeout: 180

  # --- Scenario: subtle non-activation — capture output on .NET 10 (APIs are net11-only) ---
  - name: "Non-activation: capture process output on .NET 10"
    prompt: |
      I'm on a `.NET 10` project (`net10.0`). I need to run `git` with the argument `status`
      and capture its standard output into a string. Show a minimal C# program and the
      `.csproj` XML targeting `net10.0`, and print both.
    expect_activation: false
    assertions:
      - type: "exit_success"
      - type: "output_matches"
        pattern: "net10\\.0"
      - type: "output_not_matches"
        pattern: "RunAndCaptureText|ReadAllTextAsync|ReadAllLinesAsync"
      - type: "output_matches"
        pattern: "RedirectStandardOutput|ReadToEnd"
    rubric:
      - "Solves the task with pre-.NET 11 APIs (RedirectStandardOutput + ReadToEnd/ReadToEndAsync)"
      - "Recognizes the new .NET 11 Process convenience APIs are unavailable on net10.0 and does not use them"
      - "Does NOT load the process-api-net11 skill; targets net10.0"
    timeout: 180
```

shaikhsharukh: Fixed. Applied the suggestion in full.

tests/dotnet11/process-api-net11/eval.yaml
shaikhsharukh: Fixed. Applied the suggestion in full.

tests/dotnet11/process-api-net11/eval.yaml
shaikhsharukh: Fixed. Applied the suggestion in full.

tests/dotnet11/process-api-net11/eval.yaml
shaikhsharukh: Fixed. Applied the suggestion in full.

shaikhsharukh:
I have applied all of the suggested evaluation improvements:

1. **Hardened run+capture assertions:** Updated to verify that both `.StandardOutput` and `.StandardError` are accessed, and blocked usage of the legacy event-based `BeginOutputReadLine` / `OutputDataReceived` patterns.
2. **Added `ReadAllText` (tuple return) scenario:** Created a new scenario verifying that agents properly destructure the `(StandardOutput, StandardError)` tuple when reading process output via instance methods.
3. **Added `ReadAllLinesAsync` scenario:** Added a streaming scenario verifying that agents consume `IAsyncEnumerable<ProcessOutputLine>` using `await foreach` and read `.Content` / `.StandardError` properly.
4. **Broadened non-activation guards:** Updated the .NET 8 negative scenario to block leakages of any of the new .NET 11 APIs (`RunAndCaptureText`, `ReadAllTextAsync`, `ReadAllLinesAsync`, `KillOnParentExit`).
5. **Added .NET 10 negative scenario:** Added a subtle negative scenario verifying that agents correctly fallback to pre-.NET 11 redirect and stream-reader APIs when targeting `net10.0`.

AbhitejJohn:
@eiriktsarpalis could you please review this PR? You did the deepest review on the closely analogous #535 (new dotnet11 plugin skill + evals), so your feedback here would be much appreciated.

eiriktsarpalis:
It might be best if @adamsitnik took a look.

SamMonoRT:
@adamsitnik will review once back from vacation. Please don't merge without his approval.

AbhitejJohn:
@adamsitnik , @SamMonoRT : Gentle ping on reviewing the changes here.

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+---
+
+# Process API Improvements — .NET 11
+
+New APIs added to `System.Diagnostics.Process` in .NET 11 simplify process management, eliminate boilerplate, and prevent common deadlock patterns when capturing output.
+
+## When to Use
+
+- Running or orchestrating external processes in a .NET 11 (or later) project.
+- Needing to start a process, wait for it to exit, and capture its output/error streams without risking deadlocks (`Process.RunAndCaptureTextAsync`).
```
adamsitnik: nit: we provide sync and async overloads

```suggestion
- Needing to start a process, wait for it to exit, and capture its output/error streams without risking deadlocks (`Process.RunAndCaptureText[Async]`).
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+- Wanting to ensure child processes are automatically terminated when the parent process exits (`KillOnParentExit`).
+- Requiring trimmer-friendly and NativeAOT-compatible process creation via `SafeProcessHandle`.
```
adamsitnik: The wording needs a bit more specific. `Process` itself is NativeAOT-compatible, it's just that with `SafeProcessHandle` the user gets the smallest possible size on disk (so it's more of an optimization)

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+- Requiring fine-grained control over handle inheritance (`InheritedHandles`) or starting detached processes (`StartDetached`).
+
+## When Not to Use
+
+- The project targets .NET 10 or earlier — these APIs are not available before .NET 11.
+- Running simple shells where custom execution code is unnecessary.
```
adamsitnik: I am not a native speaker, so please take it with a grain of salt. But overall I do believe that the new APIs simplify the code a lot, so I would recommend them even for simple shells.

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+- The default `Process.Start()` is sufficient and does not require output capturing or advanced lifecycle rules.
+
+## Target Framework
+
+```xml
+<TargetFramework>net11.0</TargetFramework>
+```
+
+## New APIs & Convenience Methods
```
adamsitnik: Each of these methods is a new API, some are also convenience methods. I would just call it `New APIs`

```suggestion
## New APIs
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+
+### High-Level Convenience APIs (Static Methods)
+
+#### `Process.Run` / `Process.RunAsync`
+Starts a process and waits for it to exit, returning the exit status. Does not capture standard output or error.
```
adamsitnik: We need to mention that it also allows the users to discard the output/error by providing `silent: true` and internally redirecting std handles to `NUL` device.

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+public static ProcessTextOutput RunAndCaptureText(string fileName, IList<string>? arguments = null, TimeSpan? timeout = null)
+public static Task<ProcessTextOutput> RunAndCaptureTextAsync(string fileName, IList<string>? arguments = null, CancellationToken cancellationToken = default)
+public static ProcessTextOutput RunAndCaptureText(ProcessStartInfo startInfo, TimeSpan? timeout = null)
+public static Task<ProcessTextOutput> RunAndCaptureTextAsync(ProcessStartInfo startInfo, CancellationToken cancellationToken = default)
+```
+
+#### `Process.StartAndForget`
+Launches a process and immediately releases the system handle resources, returning only the process ID (PID).
+```csharp
+public static int StartAndForget(string fileName, IList<string>? arguments = null)
```
adamsitnik: In .NET 11 Preview 7 we have changed these methods to accept `IEnumerable<string>` for the sake of consistency with pre-existing Process APIs (https://github.com/dotnet/runtime/pull/130630)

```suggestion
public static ProcessExitStatus Run(string fileName, IEnumerable<string>? arguments = null, bool silent = false, TimeSpan? timeout = null)
public static Task<ProcessExitStatus> RunAsync(string fileName, IEnumerable<string>? arguments = null, bool silent = false, CancellationToken cancellationToken = default)
public static ProcessExitStatus Run(ProcessStartInfo startInfo, TimeSpan? timeout = null)
public static Task<ProcessExitStatus> RunAsync(ProcessStartInfo startInfo, CancellationToken cancellationToken = default)
```

#### `Process.RunAndCaptureText` / `Process.RunAndCaptureTextAsync`
Starts a process, captures both standard output and error, and waits for it to exit. Extremely useful for avoiding deadlocks on stream redirection.
```csharp
public static ProcessTextOutput RunAndCaptureText(string fileName, IEnumerable<string>? arguments = null, TimeSpan? timeout = null)
public static Task<ProcessTextOutput> RunAndCaptureTextAsync(string fileName, IEnumerable<string>? arguments = null, CancellationToken cancellationToken = default)
public static ProcessTextOutput RunAndCaptureText(ProcessStartInfo startInfo, TimeSpan? timeout = null)
public static Task<ProcessTextOutput> RunAndCaptureTextAsync(ProcessStartInfo startInfo, CancellationToken cancellationToken = default)
```

#### `Process.StartAndForget`
Launches a process and immediately releases the system handle resources, returning only the process ID (PID).
```csharp
public static int StartAndForget(string fileName, IEnumerable<string>? arguments = null)
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+Starts a process, captures both standard output and error, and waits for it to exit. Extremely useful for avoiding deadlocks on stream redirection.
+```csharp
+public static ProcessTextOutput RunAndCaptureText(string fileName, IList<string>? arguments = null, TimeSpan? timeout = null)
+public static Task<ProcessTextOutput> RunAndCaptureTextAsync(string fileName, IList<string>? arguments = null, CancellationToken cancellationToken = default)
+public static ProcessTextOutput RunAndCaptureText(ProcessStartInfo startInfo, TimeSpan? timeout = null)
+public static Task<ProcessTextOutput> RunAndCaptureTextAsync(ProcessStartInfo startInfo, CancellationToken cancellationToken = default)
+```
+
+#### `Process.StartAndForget`
+Launches a process and immediately releases the system handle resources, returning only the process ID (PID).
```
adamsitnik: Let's provide the reasoning behind this method. You can re-use the text from our [blog post](https://devblogs.microsoft.com/dotnet/process-api-improvements-in-dotnet-11/#process.startandforget):

```suggestion
There is a common misconception that when a process is disposed, it’s also being killed. This is not the case, as `Process.Dispose` only releases the resources associated with the process, but does not kill it.

To make it easier to start a process without the need to worry about disposing it, `Process.StartAndForget` was introduced. The method starts a process, returns its ID and releases all resources associated with it
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Target Framework
+
+```xml
+<TargetFramework>net11.0</TargetFramework>
+```
+
+## New APIs & Convenience Methods
+
+### High-Level Convenience APIs (Static Methods)
```
adamsitnik: I would suggest we use following separation:

- New APIs
  - Convenience APIs
    - Static Methods
    - Instance Methods

As some of the instance methods are also convenience methods (at least in my opinion).

```suggestion
### High-Level Convenience APIs

#### Static Methods
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+
+```csharp
+public (string StandardOutput, string StandardError) ReadAllText(TimeSpan? timeout = null)
+public Task<(string StandardOutput, string StandardError)> ReadAllTextAsync(CancellationToken cancellationToken = default)
+public (byte[] StandardOutput, byte[] StandardError) ReadAllBytes(TimeSpan? timeout = null)
+public Task<(byte[] StandardOutput, byte[] StandardError)> ReadAllBytesAsync(CancellationToken cancellationToken = default)
+public IEnumerable<ProcessOutputLine> ReadAllLines(TimeSpan? timeout = null)
+public IAsyncEnumerable<ProcessOutputLine> ReadAllLinesAsync(CancellationToken cancellationToken = default)
+```
+*Note: `ProcessOutputLine` is a readonly struct containing `string Content` and `bool StandardError` properties.*
```
adamsitnik: We mention what `ProcessOutputLine` is, but we don't mention `ProcessExitStatus` nor `ProcessTextOutput`. I think it's best to just provide their definition just before they appear for the first time.

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+
+### ProcessStartInfo Properties
+
+#### `KillOnParentExit`
+Ensures that the spawned child process is terminated when the current (parent) process exits. Works across both Windows and Unix platforms.
```
adamsitnik: Not every Unix is supported. Only Win, Lin and Android.

```suggestion
Ensures that the spawned child process is terminated when the current (parent) process exits. Works across Windows, Linux and Android.
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+```csharp
+public bool KillOnParentExit { get; set; }
+```
+
+#### `InheritedHandles`
+Provides precise control over which file/kernel handles are inherited by the child process, preventing accidental resource leaks.
```
adamsitnik, line 113: We need to describe few important things:
- std handles are always included (no need to add to the list)
- list set to empty list == only std handles get inerited
- only `SafeFileHandle` and `SafePipeHandle` instances are allowes as of today.
- no global lock used when spawning new process on Windows (important to tune projects that spawn multiple processes in parallel)

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+```csharp
+public IList<SafeHandle>? InheritedHandles { get; set; }
+```
+
+#### `StartDetached`
+Starts the process detached from the parent's terminal or job session, ensuring it survives the parent's exit.
```
adamsitnik, line 129: When set to `true`, the std handles are by default redirected to `NUL` device. So the child process does not keep the parent process console/terminal resources alive.

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+// Run 'git status' and capture output (arguments passed as list)
+ProcessTextOutput result = await Process.RunAndCaptureTextAsync("git", ["status"]);
+
+if (result.ExitStatus.ExitCode == 0)
+{
+    Console.WriteLine($"Git Output: {result.StandardOutput.Trim()}");
```
adamsitnik: There is no need to use `Trim` here.

```suggestion
    Console.WriteLine($"Git Output: {result.StandardOutput}");
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+
+Run a process and safely read all output text without stream deadlock risks:
+
+```csharp
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+// Run 'git status' and capture output (arguments passed as list)
+ProcessTextOutput result = await Process.RunAndCaptureTextAsync("git", ["status"]);
```
adamsitnik: People tend to use `async` for apps that absolutely don't benefit from it (scalability, cancellation support etc). I think it's better to use non-async overload for such simple examples. So AI-written code is simple and does not pay for the price of using async
```suggestion
ProcessTextOutput result = Process.RunAndCaptureText("git", ["status"]);
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+### 2. Auto-Killing Child Processes on Parent Exit
+
+Ensure a long-running background worker process is killed when the main application terminates:
+
+```csharp
+using System.Diagnostics;
+
+var startInfo = new ProcessStartInfo("dotnet", ["run", "--project", "BackgroundWorker.csproj"])
+{
+    KillOnParentExit = true // Auto-teardown when this parent process exits
```
adamsitnik:
```suggestion
    KillOnParentExit = OperatingSystem.IsWindows() ||  OperatingSystem.IsLinux() // Auto-teardown when this parent process exits
```

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+### 2. Auto-Killing Child Processes on Parent Exit
+
+Ensure a long-running background worker process is killed when the main application terminates:
+
+```csharp
+using System.Diagnostics;
+
+var startInfo = new ProcessStartInfo("dotnet", ["run", "--project", "BackgroundWorker.csproj"])
```
adamsitnik: nit: subjective: no need to use `var`

```suggestion
ProcessStartInfo startInfo = new("dotnet", ["run", "--project", "BackgroundWorker.csproj"])
```

adamsitnik (commented):
@shaikhsharukh big thanks for your contribution! PTAL at my comments.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Fixed. Updated the snippet to explicitly declare `ProcessStartInfo`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Fixed. Added OS platform checks: `KillOnParentExit = OperatingSystem.IsWindows() || OperatingSystem.IsLinux()`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Fixed. Updated the example to use the synchronous `Process.RunAndCaptureText` overload.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Fixed. Removed `.Trim()`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Added note explaining standard handles redirection to the `NUL` device when `Silent` is true.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Added details on standard handles inclusion, empty list behavior, supported handle types, and Windows parallel spawning without global lock.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Updated to specify Windows, Linux, and Android.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Added definitions for `ProcessExitStatus`, `ProcessTextOutput`, and `ProcessOutputLine` under `### Types`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Reorganized sections under `### High-Level Convenience APIs` with `#### Static Methods` and `#### Instance Methods`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Added the rationale citing `Process.Dispose` behavior vs process killing, and why `Process.StartAndForget` was introduced.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Updated all method overloads to accept `IEnumerable<string>?` for arguments to align with .NET 11 Preview 7 changes (dotnet/runtime#130630).

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Mentioned that `silent: true` discards output/error by internally redirecting std handles to `NUL`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Renamed section heading to `## New APIs`.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Removed the bullet against simple shells.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Clarified that `SafeProcessHandle` is an optimization for minimal disk footprint with NativeAOT.

plugins/dotnet11/skills/process-api-net11/SKILL.md
shaikhsharukh: Updated bullet to mention `Process.RunAndCaptureText[Async]`.

shaikhsharukh:
Thank you @adamsitnik for the thorough review! All feedback has been addressed and pushed in the latest commit: updated argument types to `IEnumerable<string>?`, documented return types, clarified NativeAOT/lifecycle/handle rules, and refined the sample snippets.

plugins/dotnet11/skills/process-api-net11/SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+# Process API Improvements — .NET 11
+
+New APIs added to `System.Diagnostics.Process` in .NET 11 simplify process management, eliminate boilerplate, and prevent common deadlock patterns when capturing output.
+
+## When to Use
+
+- Running or orchestrating external processes in a .NET 11 (or later) project.
+- Needing to start a process, wait for it to exit, and capture its output/error streams without risking deadlocks (`Process.RunAndCaptureText[Async]`).
+- Wanting to ensure child processes are automatically terminated when the parent process exits (`KillOnParentExit`).
+- Requiring trimmer-friendly and NativeAOT-optimized process creation via `SafeProcessHandle` for the smallest possible disk footprint.
```
jkotas: SafeProcessHandle is more lightweight API than full Process. It has better performance characteristics on all form-factors. It is not a NativeAOT specific optimization.

shaikhsharukh:
Thank you @jkotas for the clarification! Updated the documentation in the latest commit to accurately reflect that `SafeProcessHandle` is a lightweight, low-overhead API providing better performance characteristics across all form-factors and runtimes, rather than framing it as a NativeAOT-specific optimization.

PR #876, opened 2026-07-09
Scale vally token pool to any number of rotation tokens

JanKrivanek:
## Summary

The vally shadow-evaluation token selector in `vally-evaluation.yml` only read `COPILOT_GITHUB_TOKEN` and `COPILOT_GITHUB_TOKEN_2` — a hard-coded `for i in 1 2` loop — so any additional rotation tokens configured as repo/org secrets were silently ignored. With the reduced fan-out from #872 that is only two Copilot tokens backing the whole matrix.

This makes the pool **unbounded**: it enumerates secrets via `toJson(secrets)` and collects every one named `COPILOT_GITHUB_TOKEN` or `COPILOT_GITHUB_TOKEN_<N>` (case-insensitive). Adding more tokens now needs no workflow edit. The reusable workflow is called with `secrets: inherit`, so all secrets are visible.

It also assigns a token **deterministically by matrix `job-index`** (matching the approach already used in `evaluation.yml`) instead of pure `$RANDOM`, so concurrent jobs spread evenly across the pool and avoid randomly colliding on the same token.

## Validation

- `actionlint -shellcheck= -pyflakes= .github/workflows/vally-evaluation.yml` → exit 0
- jq filter verified against a sample secrets payload (matches `COPILOT_GITHUB_TOKEN`, `_2`, `_10`; excludes `github_token`/others)

### DIFF
--- a/.github/workflows/vally-evaluation.yml
+++ b/.github/workflows/vally-evaluation.yml
@@ -109,7 +109,7 @@ jobs:
           mapfile -t TOKENS < <(
             printf '%s' "$ALL_SECRETS" \
               | jq -r 'to_entries
-                       | map(select(.key | test("^COPILOT_GITHUB_TOKEN(_[0-9]+)?$"; "i")))
+                       | map(select(.key | test("^COPILOT_GITHUB_TOKEN(_[0-9]+)?$"; "i")) | select(.value != ""))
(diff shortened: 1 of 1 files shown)


JanKrivanek:
superseded by https://github.com/dotnet/skills/pull/868

PR #883, opened 2026-07-15
Add copy-to-output-directory MSBuild skill (IfDifferent / SkipUnchangedFilesOnCopyAlways)

YuliiaKovalova:
## Summary

Adds a focused `dotnet-msbuild` skill — `copy-to-output-directory` — documenting how to choose a `CopyToOutputDirectory` / `CopyToPublishDirectory` mode, including the **`IfDifferent`** mode and the **`$(SkipUnchangedFilesOnCopyAlways)`** property introduced in MSBuild 17.13 / .NET SDK 9.0.2xx by [dotnet/msbuild#11052](https://github.com/dotnet/msbuild/pull/11052).

No existing skill covered this feature — it was only mentioned tangentially in `incremental-build` and `msbuild-antipatterns`.

## What the skill covers

- The four modes (`Never` / `PreserveNewest` / `Always` / `IfDifferent`) with a comparison table.
- Why `Always` is a recurring per-build performance hit, and the destination-mutation scenario it was historically used for.
- `IfDifferent` semantics: copies when source and destination differ in either direction (newer **or** older, or size differs), via the `Copy` task's `SkipUnchangedFiles="true"` (timestamp + size comparison).
- The `$(SkipUnchangedFilesOnCopyAlways)` property for a bulk, non-invasive opt-in on legacy `Always` items.
- How the modes flow through the common targets, transitive copy through `ProjectReference`, and the MSBuild 17.13 / .NET SDK 9.0.2xx+ version requirement.

## Changes

- **New:** `plugins/dotnet-msbuild/skills/copy-to-output-directory/SKILL.md`
- Registered the skill for routing in the `msbuild`, `msbuild-code-review`, and `build-perf` agent files.

## Validation

- `markdownlint-cli2` passes with 0 errors on the new file.


Evangelink (approved):
Reviewed in depth — approving. Verified every technical claim against the live `dotnet/msbuild` `Microsoft.Common.CurrentVersion.targets`:

- Target names `_CopyOutOfDateSourceItemsToOutputDirectoryAlways`, `_CopyDifferingSourceItemsToOutputDirectory`, and `_CopyToOutputDirectoryTransitiveItems` — exact match.
- `$(SkipUnchangedFilesOnCopyAlways)` defaults to `false` and is passed as `SkipUnchangedFiles="$(SkipUnchangedFilesOnCopyAlways)"` on the `Always` target — correct.
- `IfDifferent` target uses `SkipUnchangedFiles="true"` (timestamp + size heuristic) — correct, and the doc's careful "not a content hash" wording is accurate.
- `FileWrites` registration, transitive copy via `ProjectReference`, and ClickOnce inclusion of `IfDifferent` alongside `Always`/`PreserveNewest` — all confirmed.
- Version claim (MSBuild 17.13 / SDK 9.0.2xx / dotnet/msbuild#11052, merged 2024-12-05) — accurate.

Non-blocking notes for consideration:

1. **Skill-menu budget headroom (heads-up).** Using the validator's exact `RenderedSkillMenuCost` formula, this brings `dotnet-msbuild` to **14,953 / 15,000** chars — only ~47 chars of headroom. It passes `check`, but given the recent `restore-15k-skill-budget` work, any later skill or a small description tweak will tip the plugin over and silently truncate the alphabetically-last skill's description. Might be worth reclaiming a few chars from an over-long sibling description (e.g. `property-patterns` ~975, `target-authoring` ~929) at some point.

2. **Description routing pattern.** 16/18 sibling skills use the `USE FOR:` / `DO NOT USE FOR:` pattern for routing. This one uses a plain sentence — reasonable given the budget pressure above, just flagging it as an intentional trade-off.

3. **Minor wording.** "GetCopyToOutputDirectoryItems ... into three copy targets, which run from `CopyFilesToOutputDirectory`" — the three copy targets are actually invoked via `_CopySourceItemsToOutputDirectory` (a dependency of `CopyFilesToOutputDirectory`). Harmless simplification.

Nice, well-sourced addition. 👍

Evangelink (approved):
Re-reviewed after the follow-up commits — all three points from my prior review are addressed, and I re-verified the changes:

- **Budget headroom reclaimed.** Trimming `property-patterns` and `target-authoring` descriptions offsets the now-richer `copy-to-output-directory` description. Recomputed with the validator's exact `RenderedSkillMenuCost` formula: dotnet-msbuild is now **14,901 / 15,000** chars (~99 headroom, up from ~47). Fits with a bit more room.
- **USE FOR / DO NOT USE FOR restored** on the new skill's description — matches sibling convention.
- **Target-flow wording corrected and accurate.** Verified against the common targets: `CopyFilesToOutputDirectory` → `_CopySourceItemsToOutputDirectory` → the three copy targets (`_CopyOutOfDateSourceItemsToOutputDirectory`, `_CopyOutOfDateSourceItemsToOutputDirectoryAlways`, `_CopyDifferingSourceItemsToOutputDirectory`).
- **`IfDifferent` description softened** to correctly frame `SkipUnchangedFiles` as a timestamp+size heuristic (not a content hash) — accurate.
- **"an SQLite" article fix** — good catch.

All previously-verified technical claims still hold. LGTM. 👍

Evangelink (approved):
Approving — no changes since my last review (head 7b047f26, already verified). Technical claims check out against MSBuild source, skill-menu budget fits (14,901/15,000), and all earlier feedback was addressed. LGTM. 👍

Issue #886, opened 2026-07-15
Improve analyzing-dotnet-performance cross-family quality signal


artl93:
## Context

Follow-up to #885. The reliability investigation separated infrastructure failures from skill quality. Once transient comparison errors and unmatched trajectories are reported correctly, the cross-family evaluation still shows quality gaps that should be addressed independently.

## Evidence

From the cross-family evaluation:

- The skill passed 3 of 5 model families.
- Aggregate outcomes included 7 ties and 4 losses.
- The Opus family missed skill invocation in one cell despite 96% invocation overall.
- Tool-call overhead increased by 6.33 calls on average.
- Existing scenarios 8 and 9 have moderate overfitting risk because parts of their grading criteria are too closely tied to specific techniques or vocabulary.

## Proposed work

- Rewrite scenarios 8 and 9 toward outcome-focused, independently evaluable rubrics without weakening the intended performance findings.
- Analyze tie and loss cases across model families to identify missing guidance or ambiguous fixtures.
- Investigate the Opus activation miss without broadening the skill beyond its documented scope.
- Reduce unnecessary tool usage while preserving diagnosis quality.
- Keep `eval.yaml` and `eval.vally.yaml` semantically aligned for any test changes.

## Acceptance criteria

- Static validation passes for `plugins/dotnet-diag`.
- Updated rubrics are classified as outcome-focused rather than technique/vocabulary-focused.
- Cross-family evaluation shows no regression in currently passing families.
- Ties/losses and tool-call overhead improve, with any remaining model-specific gaps documented.
- Reliability infrastructure changes remain scoped to #885 and are not mixed into this work.

PR #905, opened 2026-07-16
Remove vocabulary gates from performance eval

artl93:
Part of #886

## Purpose

Remove vocabulary overfit from scenarios 8 and 9 **without weakening or broadening the concrete performance findings they measure**. This PR changes only the evaluator; it does not change `analyzing-dotnet-performance` itself.

The expected outcomes remain directly verifiable against the fixture code: repeated enumeration, delegate and array allocation, per-match string allocation, params-array allocation, static initialization allocation, and existing span-based positive patterns. Correct responses no longer have to use the words `LINQ`, `All`, or `Where` to receive credit.

## Changes

- remove literal `LINQ`, `All`, and `Where` output gates
- retain atomic, fixture-specific rubric criteria for the original performance findings
- avoid requiring a preferred remediation where identifying the defect is the measurable outcome
- add an independent criterion that analysis-only responses do not modify the fixture, matching the prompt and comparable evals
- rename the two scenario labels around the measured outcomes rather than implementation vocabulary

## Evidence

A matched evaluator A/B held the skill and workflow byte-identical across three repetitions and five model families:

- [current-eval control](https://github.com/dotnet/skills/actions/runs/29620802914)
- [vocabulary-gate-free treatment](https://github.com/dotnet/skills/actions/runs/29620813535)

The aggregate skilled-vs-baseline result remained effectively unchanged: **+0.258 control versus +0.265 treatment**. Unchanged scenarios differed by +0.027, establishing a larger stochastic-noise floor than the aggregate +0.007 movement.

The control produced one concrete vocabulary-only false negative: a successful MAI scenario 9 response passed `exit-success` and all 5/5 substantive rubric criteria but failed solely because its final prose omitted `Where`. This directly supports removing the hard vocabulary gate.

That A/B preceded the final rubric refinement in `e4f6235e`; it supports the gate removal, not a claim that the exact final wording improves scores. The final wording was deliberately moved back toward the original code-grounded findings after review.

## Validation

- targeted `skill-validator check --plugin plugins/dotnet-diag` passes
- YAML and focused scenario structure validation passes
- `git diff --check` passes
- branch diff contains only `tests/dotnet-diag/analyzing-dotnet-performance/eval.yaml`


tests/dotnet-diag/analyzing-dotnet-performance/eval.yaml
```diff
@@ -137,42 +137,36 @@ scenarios:
     timeout: 180
 
   - name: "Catches LINQ on hot-path string processing and All(char.IsUpper)"
-    prompt: "These string humanization extensions are called on every user-facing string in a CMS rendering pages with hundreds of dynamic strings. Analyze for performance issues."
+    prompt: "These string humanization extensions are called on every user-facing string in a CMS rendering pages with hundreds of dynamic strings. Analyze for performance issues and provide recommendations without modifying the fixture."
```
Evangelink: Could we add a rubric criterion that verifies the response only provides recommendations and does not modify the fixture? The prompt now makes non-modification an explicit requirement, but `exit_success` and the current rubric do not detect violations, so a run can edit the fixture and still pass. This would also match the enforcement used by comparable analysis-only evals.

AbhitejJohn:
Hi @artl93 — friendly ping on this PR when you get a chance! Happy to help if needed.

tests/dotnet-diag/analyzing-dotnet-performance/eval.yaml
artl93: Agreed. Added an independent analysis-only criterion to both scenarios (`StringReadability.cs` and `TimeSpanFormatting.cs`) in e4f6235e, matching the prompt and the existing test-tagging precedent.

webreidi (approved):

PR #911, opened 2026-07-16
fix(ci): unblock vally eval on the reusable PR path

AbhitejJohn:
The PR-evaluation path (`evaluation.yml` → reusable `vally-evaluation.yml`) couldn't start vally, while the skill-validator `evaluate` jobs passed. Two independent problems, both fixed here in code.

## 1. Spurious `prepare` run
The `prepare` job was gated only on `github.event_name == 'workflow_dispatch'`. That event name **propagates into `workflow_call`**, so `prepare` ran during the reusable call, read empty dispatch inputs, and threw `Invalid plugin ''`. Now additionally gated on `&& inputs.entries == ''`, so it runs only for a **direct** manual dispatch and is skipped whenever the caller supplies `entries`.

## 2. Empty `COPILOT_PAT_0..9` pool
Reusable workflows called via `workflow_call` do **not** inherit the caller's org/repo secrets, and the `copilot-pat-pool` environment holds no environment secrets — so `secrets.COPILOT_PAT_*` read empty across the boundary and the vally jobs failed with `No copilot-pat-pool entries are configured`.

The caller is a **top-level** job, so it can read the existing org-level pool (the same pool the scheduled `validate-pat-pool.yml` reads daily). `evaluation.yml` now forwards it via an explicit **named `secrets:` mapping** of the 10 tokens — mirroring how skill-validator sources its own token set in-workflow. This needs **no admin configuration** and goes green on merge.

### Security note (#868)
This deliberately does **not** use `secrets: inherit`, which would forward every repo/org secret into the reusable workflow and reverse the blast-radius reduction established in #868. The named mapping crosses **only** those 10 `COPILOT_PAT_*` tokens — the minimum the reusable workflow needs — so the exposure surface is the pool and nothing else. `environment: copilot-pat-pool` is retained, so environment secrets can still override the mapped values later with no code change.

## Scope
- Scoped to the `vally-evaluate` reusable-call job only; the skill-validator `evaluate` job is unchanged.
- Separate infra PR (distinct from #878). Must land on `main` to take effect for all PR evaluations.

## Validation
- `actionlint` 1.7.7 → exit 0 on both workflow files.
- Reviewed reusable-workflow secret semantics (rubber-duck): named mapping populates the pool with existing org secrets; only edge case is fork-PR secret availability, which matches skill-validator's existing (already-accepted) behavior.

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -930,8 +930,26 @@ jobs:
     with:
       entries: ${{ needs.discover.outputs.entries }}
       head_sha: ${{ needs.gate.outputs.head_sha }}
-    # No secrets are passed — the copilot-pat-pool secrets are injected via the
-    # `environment:` on the reusable workflow's job, not across this boundary.
+    # Forward ONLY the Copilot PAT pool across the workflow_call boundary. A
+    # reusable workflow does not inherit the caller's org/repo secrets, so the
+    # pool would read empty on this path unless it is passed explicitly. We map
+    # the 10 named tokens rather than using `secrets: inherit` — inherit would
+    # forward every repo/org secret into the reusable workflow and reverse the
+    # blast-radius reduction from #868; named mapping crosses only these tokens.
+    # This job is a top-level caller, so `secrets.COPILOT_PAT_*` resolves from the
+    # existing ORG secrets — the same pool the scheduled validate-pat-pool workflow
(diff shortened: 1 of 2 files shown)


AbhitejJohn:
Alright, the CI tests against main but here's what proves that this works: 
- https://github.com/dotnet/skills/actions/runs/29533795703
- https://github.com/dotnet/skills/actions/runs/29526420471

JanKrivanek (approved):

PR #936, opened 2026-07-23
Strengthen mcp-csharp-test with prescriptive MCP-specific steps

leslierichardson95:
## What

Strengthens `mcp-csharp-test` with prescriptive, MCP-specific guidance so it helps strong models, not just weak ones.

## Why

The cross-family skills-eval (#889) rated this skill **high-impact (0.65)** but **STRENGTHEN 0/5** — it misses both frontier models because it *"assumes context they solve unaided."* The body largely showed generic xUnit patterns a frontier model already knows, so a skilled run tied with baseline.

## Changes (kept lean — the skill is already token-efficient)

- **Default integration tests to the SDK's in-memory `ClientServerTestBase`** instead of spawning `dotnet run` via `StdioClientTransport`; reserve the process-spawning transport for a single end-to-end smoke test. (Per-fixture process spawning is the top cause of flaky/hanging integration tests — the skill's own Pitfalls section calls this out.)
- Add an **MCP-specific quality assertion**: every tool must expose a non-empty `Description` (MCP clients rely on descriptions to choose tools) — a check frontier models don't write by default.
- Add matching **Validation** checklist items.

## Verification

`skill-validator check` runs in CI. Recommend a cross-family eval re-run (`-f skills="mcp-csharp-test" -f executors="opus,gpt" -f runs=5`; prior `avgN` was ~2.8, so use more runs).

Refs #889

### DIFF
--- a/plugins/dotnet-ai/skills/mcp-csharp-test/SKILL.md
+++ b/plugins/dotnet-ai/skills/mcp-csharp-test/SKILL.md
@@ -93,7 +93,9 @@ public class ApiToolTests
 
 ### Step 3: Write integration tests with MCP client
 
-Test the full MCP protocol using a client-server connection:
+Test the full MCP protocol through a real MCP client.
+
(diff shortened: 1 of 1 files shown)


leslierichardson95:
Superseded by #940, which moves this change to a branch on the primary dotnet/skills repo. Closing this fork PR.

PR #938, opened 2026-07-23
Cut mcp-csharp-debug skill

leslierichardson95:
## What

Cuts the mcp-csharp-debug skill from the dotnet-ai plugin.

## Why

Per the cross-family skills-eval report (#889) and the follow-up telemetry review, mcp-csharp-debug is a **cut candidate**, not a strengthen target:

- **Eval:** STRENGTHEN 0/5 families, low impact (+0.27), mostly ties (inert) and token-bloated.
- **Telemetry (§9c, weak eval + thin real audience):** 3 real users, **0 with ≥5 invocations**, 4 real invocations. Guidance from the team: *"It's ok to cut skills that need a lot of work but aren't used much."*
- Cutting was explicitly confirmed as the right path by the eval owner.

## Changes

- Delete plugins/dotnet-ai/skills/mcp-csharp-debug/ (SKILL.md + references)
- Delete 	ests/dotnet-ai/mcp-csharp-debug/ (eval)
- Remove CODEOWNERS entries for the skill and its tests
- Remove all dangling `mcp-csharp-debug` references (frontmatter `DO NOT USE FOR`, Stop Signals, Related Skills) from mcp-csharp-create, mcp-csharp-test, mcp-csharp-publish
- Update a skill-validator doc example to reference a surviving skill

Verified no remaining `mcp-csharp-debug` references repo-wide (`git grep`). `skill-validator check` will run in CI (local run skipped — repo pins .NET 11 preview SDK not installed on this machine).

Refs #889

---
_Moved from fork branch to the primary dotnet/skills repo; supersedes #934._


AbhitejJohn:
Similar question as the other PR @leslierichardson95 : Do you have thoughts on the other mcp-csharp-* skills? Do you think those are still useful? The evals in the issue suggest they all need a lift too to be useful in all the models.

AbhitejJohn (approved):
Approving this change but it would be nice to re-review the other mcp-csharp-* skills as well to see if they hold value.

leslierichardson95:
> Similar question as the other PR @leslierichardson95 : Do you have thoughts on the other mcp-csharp-* skills? Do you think those are still useful? The evals in the issue suggest they all need a lift too to be useful in all the models.

@AbhitejJohn Upon reflection, I agree these skills have run their course, so I'll make another PR in a few that deletes the other three (mcp-test, -create, -publish).

PR #939, opened 2026-07-23
Strengthen mcp-csharp-publish with prescriptive defaults

leslierichardson95:
## What

Strengthens the `mcp-csharp-publish` skill so it changes agent behavior instead of tying with the no-skill baseline.

## Why

The cross-family skills-eval (#889) rated `mcp-csharp-publish` **STRENGTHEN 0/5** with the diagnosis *"mostly ties — too generic/non-prescriptive."* The skill listed options (NuGet **or** Docker, Container Apps **or** App Service, Registry *optional*) without deciding, so a skilled run rarely differed from baseline.

## Changes

- Add a prescriptive **Default decisions** block up front that makes the common choices without asking:
  - stdio → **NuGet.org as a .NET tool** (`dnx`)
  - HTTP → **Azure Container Apps** with `--min-replicas 0` (scale to zero)
  - **Always publish `server.json`** to the MCP Registry for shared servers
  - Keep versions identical across `.csproj` / `server.json`
- Make Step 3 decisive: Container Apps is the default; **App Service is gated** behind an explicit need (existing plan, VNet, slots).
- Reinforce the Registry step as the default for shared servers.

No new heavy content — the skill is already token-efficient; this sharpens decisiveness only.

## Verification

`skill-validator check` runs in CI. Recommend a cross-family eval re-run (`-f skills="mcp-csharp-publish" -f executors="opus,gpt" -f runs=5`) to confirm the ties convert to wins.

Refs #889

---
_Moved from fork branch to the primary dotnet/skills repo; supersedes #935._


leslierichardson95:
Superseded by #943, which retires this skill (and the rest of the mcp-csharp-* family) per the #889 eval + telemetry review. Closing as obsolete.

PR #949, opened 2026-07-27
Close the dotnet-test skill-coverage gaps from the PR 945 report

Evangelink:
## Summary

The skill coverage report on #945 flagged roughly 190 uncovered teaching points across 9 `dotnet-test` skills, 7 of them at 0%. Reproducing it locally showed those skills all report `Evidence items: 0`, so most of that number was a measurement artifact rather than a real gap.

`eng/skill-coverage/Measure-SkillCoverage.ps1` parses an eval schema this repo does not use. It looks for `assertions:` blocks with sibling `pattern:`/`value:` keys, but all `tests/**/eval.yaml` files use `graders:` with a nested `config:` map. It also only captured rubric entries written as single-line quoted scalars, while nearly every rubric item here is an unquoted plain scalar wrapped across lines. The tool therefore found no evidence at all and reported every teaching point as uncovered. Writing ~190 assertions to satisfy that would have produced exactly the overfitting `eng/skill-validator/src/docs/OverfittingDetection.md` warns about.

**1. Fix the measurement.** `Get-EvalEvidence` is rewritten as an indentation-driven parser that reads the real spec shape: graders with nested config (`pattern`, `substring`, `value`, `command`, `stdout_contains`, `stdout_matches`, and `path` for file-exists graders), plus block and folded scalars for both grader values and rubric items. Literal values are regex-escaped before code-pattern matching so a substring is not mistakenly treated as a regex. This alone moved the 9 skills from 0-33% to 82-100%.

**2. Close the points that were genuinely uncovered.** All 9 skills now sit at 100%:

- `detect-static-dependencies`: new accounting-discipline stimulus for `file:line` locations, one authoritative total, and pure helpers such as `Path.Combine` excluded from the needs-wrapping total.
- `migrate-static-to-wrapper`: new `with-tests` fixture (a custom `IClock` seam plus an MSTest project) and a stimulus covering the required `using` directive, test doubles in the updated tests, scope discipline, and the Step 7 migration summary; plus the pitfall that static classes cannot take constructors, so the ambient context seam is used instead of converting the type.
- `test-smell-detection`: wires up the orphaned `skip-and-magic` fixture, which was committed but had no stimulus referencing it and was missing its production class. Covers reasoned vs bare skips, contextually obvious numbers, justified severities, and concrete fix examples.
- `test-tagging`: the report-only "no source files were modified" rule folded into the existing Go stimulus, and the `[Category]` attribute pattern credited from the existing `file-contains` grader.
- `assertion-quality`, `test-anti-patterns`, `test-gap-analysis`: rubric items for metric arithmetic, per-finding locations, severity prioritisation, empirical survivor confirmation, mutation category labels, and settled verdicts.

**3. Address the review feedback on the parser** (commit 2). Three correctness fixes, none of which move any number:

- `Remove-YamlQuoting` now applies each quoting style's escape rules — single-quoted `''` collapses to `'` (the corpus has `user''s` and `xUnit''s`), and double-quoted scalars honour `\"`, `\\`, `\n`, `\t`, `\r` rather than only `\\`.
- `Read-YamlScalar` switches to indentation-only consumption once a `|`/`>` header is seen, instead of applying the plain-scalar structural breaks that would truncate a block scalar at its first blank, `#`, `- ` or `key:`-shaped line. The same treatment now applies to a quoted scalar until its closing quote is reached, which fixes the one real truncation in the corpus today: the double-quoted rubric item in `dotnet-upgrade/migrate-nullable-references` whose continuation line starts with `#nullable disable`.
- The stimulus boundary is anchored to the indent of the first `- name:` seen and only fires outside a graders/rubric block, so a deeper rubric bullet starting with `name:` can no longer be mistaken for a new scenario and reset the section.

**4. Repair the broken eval fixtures the eval run exposed** (commit 3). An eval run over `test-gap-analysis`, `test-smell-detection` and `test-tagging` failed 6 scenarios. Three were the intended baseline signal — an unskilled agent picks domain categories (`Validation`, `BusinessLogic`) instead of the positive/negative taxonomy — and one was a harness flake where the agent could not read the file at all. The other two were eval defects, and chasing them surfaced a much larger problem: **every C# fixture behind these three skills failed to compile before the agent touched it.**

- All 5 `test-tagging` and all 4 `test-gap-analysis` fixtures omit `ImplicitUsings`/`Nullable` while their sources use `List<>`, `Task<>`, `CancellationToken`, `Dictionary<,>`, `DateTime`, `ArgumentOutOfRangeException` and `string?` with no `using` directives.
- All 4 `test-gap-analysis` test projects pin **MSTest 4.1.0** but call `Assert.ThrowsException<T>`, which v4 removed. Renamed to `ThrowsExactly<T>`, the exact-type-match equivalent this repo's own `migrate-mstest-v3-to-v4` skill prescribes.

This mattered most for `test-gap-analysis`, whose entire method is *inject a mutation, run the suite, see whether it fails*. On a project that does not build, every mutation looks the same. The judges recorded the cost directly: *"proactively fixed pre-existing build issues (missing ImplicitUsings)"* and *"the build couldn't be verified due to pre-existing errors in the source project"*. The `test-tagging` build grader was likewise measuring whether the agent noticed and repaired an unrelated csproj bug, not whether tagging preserved buildability.

Two brittle graders that produced false negatives on runs the judge scored as correct are also fixed: `output-matches /\[Category\b/` only tested whether the agent echoed attribute syntax in prose (removed — `file-contains "[Category("` already proves it), and `/(distribution|summary|trait|count)/` missed a run the judge called *"a clear distribution summary"* (widened with `breakdown|tally|table|total`).

### Worth a careful look

- The parser rewrite is the load-bearing change. It is cross-checked against a real YAML parser rather than eyeballed, but it is trusted tooling that runs from the base branch in CI, so a second opinion on `Read-YamlScalar` and the section-tracking loop is welcome.
- The `Get-CodePatterns` change (emitting the open bracket form of an attribute pattern) moves coverage on 5 of 75 skills. Each was verified to be a genuine pre-existing false negative — the rubrics literally name `[CascadingParameter]`, `[FromForm]` and `[TestMethod]`.
- `test-smell-detection` fixtures are deliberately standalone test files with no production class and no build grader, so they are intentionally left alone.
- This was originally stacked on `dev/amauryleve/dotnet-test-eval-followups` (#945). Now that #945 has merged, the branch is rebased onto `main`. In the conflicting files, main's versions supersede this PR's, which is why the `.gitignore` negation and the Cobertura fixture no longer appear in the diff — #945 landed better versions of both.

### Known pre-existing defect, deliberately not fixed here

The PyYAML cross-check surfaces one file where a real YAML parser and this parser legitimately disagree: `tests/dotnet-test-migration/migrate-mstest-v3-to-v4/eval.yaml` has two rubric items written as unquoted `Shows the correct syntax: ...` strings, which YAML parses as maps rather than strings. A map rubric item will not deserialize into the `List<string>` the eval runner expects. It is pre-existing on `main` and in a different skill family, so it belongs in its own change.

## Related issue

Follows up the coverage report on #945.

## Validation

- `Measure-SkillCoverage.ps1 -All -Format Json` cross-checked against a real YAML parser (PyYAML): **1 mismatch across all 98 `eval.yaml` files**, and that one is the pre-existing `migrate-mstest-v3-to-v4` map-rubric defect described above.
- All 628 fixture `src:` references resolve.
- Coverage is unchanged across all 75 skills for the commit-2 parser fixes, confirming they close latent foot-guns without moving numbers.
- **All 9 repaired fixtures build and pass**, run outside the repo so the root `global.json` MTP setting does not interfere: `test-gap-analysis` 11 / 9 / 2 / 15 tests, `test-tagging` 23 / 23 / 23 / 15 / 10 tests. Every one of them failed to compile beforehand.
- **The intended mutation behaviour is confirmed empirically, not assumed.** Applying real mutations to the repaired `boundary-gaps` fixture: all four survivors its rubric names still survive (discount tiers `>= 1000`→`>`, `>= 500`→`>`, free shipping `> 250`→`>=`, heavy item `> 10.0`→`>=`). Applying four mutations to `well-tested`: all four are killed, matching its "acknowledge well-tested code" premise.
- `dotnet test Billing.Tests` on the `with-tests` fixture: 3 passed. `dotnet build Inventory.Tests` on `skip-and-magic`: succeeded, 0 warnings.
- Unit-checked the three parser fixes directly: `''`/`\"`/`\\` unquoting round-trips, a block scalar containing a blank line, a `#` line, a `- ` bullet and a `key:` line is captured whole, and a synthetic rubric bullet with a nested `- name:` no longer opens a second scenario.
- Final coverage for the 9 reported skills: all 100%.
- `skill-validator check` could not run: building it downloads `@github/copilot-win32-x64` from npmjs and the sandbox blocks that TLS handshake. No `plugins/**` content changed in this PR, so its checks are unaffected.

## Checklist

- [x] I updated CODEOWNERS when adding or moving owned content. (N/A - no owned content moved)
- [x] I updated all marketplace manifests when plugin metadata changed. (N/A - no plugin metadata changed)
- [x] I updated `eng/known-domains.txt` for any new external domains referenced by skill content. (N/A - no new external domains)

### DIFF
--- a/eng/skill-coverage/Measure-SkillCoverage.ps1
+++ b/eng/skill-coverage/Measure-SkillCoverage.ps1
@@ -293,89 +293,169 @@ function Get-SignificantTerms([string]$text) {
 #  eval.yaml Parsing - extract test evidence
 # ═══════════════════════════════════════════════════════════
 
+function Get-LineIndent([string]$line) {
+    if ($line -match '^(\s*)\S') { return $Matches[1].Length }
+    return -1   # blank or whitespace-only
+}
+
+function Remove-YamlQuoting([string]$text) {
+    $t = $text.Trim()
+    if ($t.Length -ge 2 -and $t.StartsWith('"') -and $t.EndsWith('"')) {
+        return ($t.Substring(1, $t.Length - 2) -replace '\\\\', '\')
+    }
+    if ($t.Length -ge 2 -and $t.StartsWith("'") -and $t.EndsWith("'")) {
+        return $t.Substring(1, $t.Length - 2)
+    }
+    return $t
+}
+
+# Reads a YAML scalar that may be a block scalar (| or >) or a plain/quoted
+# scalar folded across several more-indented continuation lines. Returns the
+# joined value plus the index of the last line consumed.
+function Read-YamlScalar {
+    param([string[]]$Lines, [int]$Index, [int]$KeyIndent, [string]$Inline)
+
+    $inline = $Inline.Trim()
+    $buffer = @()
+    $last = $Index
+
+    if ($inline -match '^[|>][+-]?\d*$') { $inline = '' }
+    elseif ($inline) { $buffer += $inline }
+
+    for ($j = $Index + 1; $j -lt $Lines.Count; $j++) {
+        $candidate = $Lines[$j].TrimEnd()
+        $indent = Get-LineIndent $candidate
+        if ($indent -lt 0) { break }
+        if ($indent -le $KeyIndent) { break }
+
+        $trimmed = $candidate.Trim()
+        if ($trimmed.StartsWith('#')) { break }
+        if ($trimmed -match '^-(\s|$)') { break }
+        if ($trimmed -match '^[A-Za-z_][\w.-]*:(\s|$)') { break }
+
+        $buffer += $trimmed
+        $last = $j
+    }
+
+    [PSCustomObject]@{
+        Value     = ($buffer -join ' ').Trim()
+        LastIndex = $last
+    }
+}
+
+<#
+.SYNOPSIS
+    Extracts test evidence from an eval.yaml spec.
+
+.DESCRIPTION
+    Evidence is anything the eval actually checks: grader configuration
+    (regex patterns, literal substrings, expected values, commands) and
+    LLM-judged rubric items. Both are matched against SKILL.md teaching
+    points by the coverage engine.
+
+    The parser is indentation-driven so it handles the real spec shape:
+    graders are a list of `- type:` entries each carrying a nested `config:`
+    map, and rubric items are plain (usually unquoted) scalars that commonly
+    wrap across several lines.
+#>
 function Get-EvalEvidence([string]$yamlContent) {
+    # Grader config keys whose value is a regular expression.
+    $regexKeys = @('pattern', 'stdout_matches')
+    # Grader config keys whose value is a literal string.
+    $literalKeys = @('substring', 'value', 'stdout_contains', 'command')
+
     $evidence = @()
     $scenarioCount = 0
     $scenario = '(unknown)'
     $section = 'none'
-    $assertType = $null
+    $sectionIndent = -1
+    $graderType = $null
 
-    foreach ($rawLine in $yamlContent -split "`n") {
-        $line = $rawLine.TrimEnd()
+    $lines = $yamlContent -split "`n"
 
-        # Scenario name
-        if ($line -match '^\s+-?\s*name:\s*[''"]?(.+?)[''"]?\s*$') {
-            $scenario = $Matches[1]
-            $scenarioCount++
+    for ($i = 0; $i -lt $lines.Count; $i++) {
+        $line = $lines[$i].TrimEnd()
+        $indent = Get-LineIndent $line
+        if ($indent -lt 0) { continue }
+        $trimmed = $line.Trim()
+        if ($trimmed.StartsWith('#')) { continue }
+
+        # Leaving the current block ends the section.
+        if ($section -ne 'none' -and $indent -le $sectionIndent) {
             $section = 'none'
-            $assertType = $null
+            $sectionIndent = -1
+            $graderType = $null
         }
 
-        # Section transitions
-        if ($line -match '^\s{2,6}assertions:\s*$')  { $section = 'assertions'; continue }
-        if ($line -match '^\s{2,6}rubric:\s*$')       { $section = 'rubric'; continue }
-        if ($line -match '^\s{2,6}(prompt|setup|timeout|reject_tools|expect_tools|expect_activation):' -and $section -ne 'none') {
+        # Stimulus boundary: `- name: <scenario>` inside the stimuli list.
+        if ($indent -ge 2 -and $trimmed -match '^-\s+name:\s*(.+)$') {
+            $scenario = Remove-YamlQuoting $Matches[1]
+            $scenarioCount++
             $section = 'none'
+            $sectionIndent = -1
+            $graderType = $null
             continue
         }
 
-        # Assertion fields
-        if ($section -eq 'assertions') {
-            if ($line -match '^\s+-\s*type:\s*[''"]?(\S+?)[''"]?\s*$') {
-                $assertType = $Matches[1]
-                if ($assertType -eq 'exit_success') {
+        if ($trimmed -eq 'graders:') {
+            $section = 'graders'; $sectionIndent = $indent; $graderType = $null; continue
+        }
+        if ($trimmed -eq 'rubric:') {
+            $section = 'rubric'; $sectionIndent = $indent; continue
+        }
+
+        if ($section -eq 'graders') {
+            if ($trimmed -match '^-\s+type:\s*(.+)$') {
+                $graderType = (Remove-YamlQuoting $Matches[1]) -replace '[^\w.-]', ''
+                if ($graderType -eq 'exit-success') {
                     $evidence += [PSCustomObject]@{
                         Scenario     = $scenario
-                        EvidenceType = 'assertion:exit_success'
-                        Content      = 'exit_success: project builds and tests pass'
+                        EvidenceType = 'assertion:exit-success'
+                        Content      = 'exit-success: project builds and tests pass'
                         RawPattern   = $null
                     }
                 }
+                continue
             }
-            if ($line -match '^\s+pattern:\s*"(.+?)"\s*$') {
-                $regex = $Matches[1] -replace '\\\\', '\'
+
+            if ($trimmed -match '^([A-Za-z_][\w.-]*):\s*(.*)$') {
+                $key = $Matches[1]
+                $rest = $Matches[2]
+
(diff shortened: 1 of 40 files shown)


eng/skill-coverage/Measure-SkillCoverage.ps1
```diff
@@ -293,89 +293,169 @@ function Get-SignificantTerms([string]$text) {
...
+    $buffer = @()
+    $last = $Index
+
+    if ($inline -match '^[|>][+-]?\d*$') { $inline = '' }
+    elseif ($inline) { $buffer += $inline }
+
+    for ($j = $Index + 1; $j -lt $Lines.Count; $j++) {
+        $candidate = $Lines[$j].TrimEnd()
+        $indent = Get-LineIndent $candidate
+        if ($indent -lt 0) { break }
```
AbhitejJohn: Line 322 recognizes a `|`/`>` header but then clears it, and this loop still treats a blank line, `- `, `#`, or `key:`-shaped line as a terminator — so a block-scalar value would get truncated at the first such line (dropped evidence → under-counted coverage, i.e. the bug class this rewrite targets). It's latent today: no evidence-bearing key uses a block scalar (only `prompt:` does, and that isn't scanned), and the folded `>-` rubric items that do exist in the corpus don't contain any of those lines. But since this runs from the base branch in CI, it's worth switching to indentation-only consumption once a `|`/`>` header has been seen, rather than applying these structural breaks unconditionally.

eng/skill-coverage/Measure-SkillCoverage.ps1
```diff
@@ -293,89 +293,169 @@ function Get-SignificantTerms([string]$text) {
...
+            $sectionIndent = -1
+            $graderType = $null
         }
 
-        # Section transitions
-        if ($line -match '^\s{2,6}assertions:\s*$')  { $section = 'assertions'; continue }
-        if ($line -match '^\s{2,6}rubric:\s*$')       { $section = 'rubric'; continue }
-        if ($line -match '^\s{2,6}(prompt|setup|timeout|reject_tools|expect_tools|expect_activation):' -and $section -ne 'none') {
+        # Stimulus boundary: `- name: <scenario>` inside the stimuli list.
+        if ($indent -ge 2 -and $trimmed -match '^-\s+name:\s*(.+)$') {
```
AbhitejJohn: This boundary matches `^-\s+name:` at any indent >= 2 regardless of the current section, so a rubric bullet that starts with `name:` would be miscounted as a new scenario and reset `$section`, dropping the surrounding rubric/grader evidence. Latent — every `- name:` in the corpus today is at scenario indent (2), none deeper — but constraining it to the scenario indent (`-eq 2`, and/or only when `$section -eq 'none'`) would remove the foot-gun.

eng/skill-coverage/Measure-SkillCoverage.ps1
```diff
@@ -293,89 +293,169 @@ function Get-SignificantTerms([string]$text) {
...
+function Remove-YamlQuoting([string]$text) {
+    $t = $text.Trim()
+    if ($t.Length -ge 2 -and $t.StartsWith('"') -and $t.EndsWith('"')) {
+        return ($t.Substring(1, $t.Length - 2) -replace '\\\\', '\')
+    }
+    if ($t.Length -ge 2 -and $t.StartsWith("'") -and $t.EndsWith("'")) {
+        return $t.Substring(1, $t.Length - 2)
+    }
+    return $t
+}
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in bd6da7b.

`Remove-YamlQuoting` now applies each quoting style's escape rules: single-quoted scalars collapse `''` to `'`, and double-quoted scalars honour `\"`, `\\`, `\n`, `\t` and `\r` rather than only `\\`. Verified on the `user''s` / `xUnit''s` cases plus a `"a \"quoted\" word and a \\ slash"` round-trip.

The whole corpus is now cross-checked against PyYAML: 98 of 98 `eval.yaml` files produce identical evidence, with one known exception in `tests/dotnet-test-migration/migrate-mstest-v3-to-v4/eval.yaml`, where two rubric items are unquoted `Shows the correct syntax: ...` strings that YAML parses as maps rather than strings. That is a pre-existing authoring defect on `main` in another skill family, so I left it out of this PR — but it is worth a separate fix, since a map rubric item will not deserialize into the `List<string>` the eval runner expects.

eng/skill-coverage/Measure-SkillCoverage.ps1
Evangelink: Agreed, and done in bd6da7b — `Read-YamlScalar` now records whether a `|`/`>` header was seen and, if so, consumes by indentation alone: blank lines, `#` lines, `- ` bullets and `key:`-shaped lines are all treated as content until a line dedents to the key.

Chasing this turned up a second instance of the same bug that is **not** latent. A quoted scalar wrapped across lines hits the identical breaks, and `tests/dotnet-upgrade/migrate-nullable-references/eval.yaml` has one whose continuation line starts with `#nullable disable` — that rubric item was being truncated mid-sentence today. So the literal mode now also covers a quoted scalar until its closing quote is reached (honouring `\"` and `''`), which is what closed the last PyYAML mismatch.

Regression check across the repo: coverage percentages and evidence counts are identical for all 75 skills before and after this change, so it is a pure foot-gun fix rather than a number mover.

eng/skill-coverage/Measure-SkillCoverage.ps1
Evangelink: Fixed in bd6da7b, taking both suggestions. The boundary now requires the parser to be outside any graders/rubric block (`$section -eq 'none'`) and anchors to the indent of the first `- name:` it sees, so a deeper `- name:` stays inside its enclosing block instead of resetting the section.

Confirmed with a synthetic spec where a rubric bullet has a nested `- name: ...` line: the parser now reports one scenario and keeps the line as rubric evidence, where before it reported two and dropped the surrounding evidence.

Evangelink (commented):
/evaluate

YuliiaKovalova (approved):

PR #943, opened 2026-07-24
Retire mcp-csharp-create, mcp-csharp-test, and mcp-csharp-publish skills

leslierichardson95:
## What

Retires the remaining three `mcp-csharp-*` skills from the `dotnet-ai` plugin: **`mcp-csharp-create`**, **`mcp-csharp-test`**, and **`mcp-csharp-publish`**. Follows the cut of `mcp-csharp-debug` (#938) — this completes retirement of the family.

## Why

Based on the cross-family skills-eval (#889) and the 30-day telemetry review ([gist](https://gist.github.com/AbhitejJohn/db57f38051ebf4e7de5556a4bd98ec15)):

| Skill | Eval (impact / families ✓) | Real demand |
|---|---|---|
| `mcp-csharp-test` | +0.65 but **0/5 families pass** | 2 users, 0 with ≥5 invocations |
| `mcp-csharp-publish` | +0.56 but **0/5 families pass** | **0 real users** |
| `mcp-csharp-create` | 0.40, wins **only 2/5** (Sonnet, Haiku) and **regresses both frontier models**; avgN 2.6 (directional) | absent from telemetry report |

None has demonstrated real demand. The only positive signal is a low-confidence eval that would require a full 5-family sweep to validate, with no usage to justify the investment. Retiring now (reversible via git history) is the pragmatic call over speculative strengthening.

## Changes

- Delete `plugins/dotnet-ai/skills/mcp-csharp-{create,test,publish}/` (SKILL.md + references)
- Delete `tests/dotnet-ai/mcp-csharp-{create,test,publish}/` (eval.yaml)
- Remove the `dotnet-ai` CODEOWNERS block for the three skills and their tests
- Repoint the `skill-validator` doc example (`InvestigatingResults.md`) from the now-deleted `mcp-csharp-create` to a surviving skill (`technology-selection`)

`dotnet-ai` retains `technology-selection`. `plugin.json` auto-discovers `./skills/`, so no manifest change is needed. Verified **no remaining `mcp-csharp-*` references** repo-wide (`git grep`). Rebased on `main` **after #938 merged**, so there are no conflicts with the debug cut.

Refs #889. **Supersedes #939, #940, #941** (the earlier strengthen/fix PRs, now obsolete).

PR #965, opened 2026-07-29
Fix the underpowered eval gate: score direction, not magnitude

### LINKED ISSUE #952: Eval gate is underpowered: 27% of verdicts flip on byte-identical inputs (A/A test); 53% of evals run at n<=3
## Summary

**Measured on byte-identical inputs, the eval gate flips ~27% of skill verdicts between runs.**

`dotnet-skills.experiment.yaml` sets `runs: 1`, so the number of trials behind a skill's verdict equals its **scenario count**, and a single judge call decides each scenario. **53% of the repo's skill evals sit at n ≤ 3.** At that sample size the gate's noise floor is as large as the effect it is trying to measure, so it cannot answer the question it exists to answer — *"did this skill change help?"*

This is not an inference from the arithmetic. It was measured directly with an A/A test.

## Evidence: A/A test — 3 of 11 verdicts flipped with zero input change

Splitting a CI gate out of #945 produced an unintentionally clean controlled experiment. Two full evaluation runs where the **only** difference between the evaluated commits is four infrastructure files the harness never reads:

```
$ git diff --name-only 29edcbcdb caa0d2d61
.github/workflows/eval-quality.yml
eng/eval-quality/README.md
eng/eval-quality/check_eval_quality.py
eng/eval-quality/selftest_eval_quality.py

$ git diff --name-only 29edcbcdb caa0d2d61 -- tests/ plugins/ dotnet-skills.experiment.yaml
(empty)
```

Every eval spec, fixture, skill and the experiment config are byte-identical, so any difference in results is pure judge variance.

| Skill | n | Run A ([30284396485](https://github.com/dotnet/skills/actions/runs/30284396485)) | Run B ([30292432365](https://github.com/dotnet/skills/actions/runs/30292432365)) | |
| --- | ---: | --- | --- | --- |
| `coverage-analysis` | 3 | ❌ +60.0% 3/0/0 | ✅ +40.0% 3/0/0 | **flipped** |
| `find-untested-sources` | 4 | ❌ +30.0% 3/1/0 | ✅ +40.0% 4/0/0 | **flipped** |
| `test-tagging` | 9 | ✅ +40.0% 7/1/1 | ❌ +33.3% 5/2/2 | **flipped** |
| `test-gap-analysis` | 6 | ❌ +20.0% 3/3/0 | ❌ +26.7% 5/0/1 | |
| `code-testing-agent` | 3 | ❌ +0.0% 1/1/1 | ❌ +13.3% 1/2/0 | |
| `test-anti-patterns` | 8 | ✅ +40.0% 8/0/0 | ✅ +30.0% 6/2/0 | |
| `detect-static-dependencies` | 6 | ✅ +40.0% 6/0/0 | ✅ +33.3% 5/1/0 | |
| `test-smell-detection` | 6 | ✅ +33.3% 5/1/0 | ✅ +26.7% 4/2/0 | |
| `crap-score` | 6 | ✅ +43.3% 5/1/0 | ✅ +43.3% 5/1/0 | stable |
| `migrate-static-to-wrapper` | 5 | ❌ +44.0% 4/1/0 | ❌ +44.0% 4/1/0 | stable |
| `assertion-quality` | 6 | ❌ +26.7% 5/0/1 | ❌ +26.7% 5/0/1 | stable |

**3 of 11 verdicts flipped (27%) with zero input change.**

### The `coverage-analysis` case is the clearest illustration

It had failed **five consecutive runs while winning 100% of its trials**. In run B it passed — with the *same* 3W/0T/0L record it has always had. Nothing about the skill or its eval changed; the judge simply scored one trial +0.4 instead of +0.6, which collapsed `sd` to 0 and made the CI degenerate to `[+40%, +40%]`.

At n=3 the verdict is decided by the *magnitude* the judge happens to assign, not by whether the skill won. A skill can win every trial and fail, or win every trial and pass, on identical inputs.

An earlier, independent run pair on unchanged evals showed the same churn: `assertion-quality` 6W/0T/0L (+40%) → 4W/1T/1L (+20%), `test-tagging` 6W/0T/2L → 5W/3T/0L, `test-smell-detection` 4W/0T/0L → 2W/2T/0L.

### Two consequences

1. **A/B comparisons across a PR boundary are confounded** by a ~27% chance of a spurious flip per skill. A single run cannot establish that a skill change helped or hurt.
2. **Verdict changes must not be attributed to code changes without diffing first.** Two of the three flips above would otherwise have been misread as consequences of the split that caused them.

## Why: the arithmetic behind it

The pass gate is `mean > 0 ∧ ci_low > 0`, which reduces to:

```
sqrt(n) × (mean / sd) > t(n-1)
```

| n | required mean/sd |
| ---: | ---: |
| 1 | undefined — one trial decides |
| 2 | 3.04 |
| 3 | 1.84 |
| 4 | 1.39 |
| 6 | 1.00 |
| 8 | 0.82 |

Measured across all 97 specs by `eng/eval-quality/check_eval_quality.py`:

| n | evals |
| ---: | ---: |
| 1 | 18 |
| 2 | 10 |
| 3 | 21 |
| **≤3** | **49 of 93** |

At n ≤ 3 the required signal-to-noise ratio (1.84, or unbounded at n=1) is higher than a genuinely good skill reliably produces, which is why winning every trial is not sufficient to pass.

## Corroborating: skills that win every trial still fail

From five consecutive `dotnet-test` evaluation runs on #945:

| Skill | Runs failed with **zero losses** | Example |
| --- | --- | --- |
| `coverage-analysis` | **5 of 5** | won 100% of trials every time: +40%, +60%, +60%, +60%, +60% |
| `migrate-static-to-wrapper` | 3 | 4W/1T/0L, `ci_low = −0.4%` — missed by 0.4 of a point |
| `test-gap-analysis` | 1 | 3W/3T/0L |
| `find-untested-sources` | 1 | 3W/1T/0L |
| `test-smell-detection` | 1 | 2W/2T/0L |

## Why a blanket `runs: 3` is not the answer

The obvious fix triples trial count and drops the required ratio at n=3 from 1.84 to 0.77. But cost is not uniform:

- the `dotnet-test` vally job currently takes **~75 minutes** at `runs: 1`; ×3 ≈ 225 min
- GitHub-hosted jobs cap at 360 min, and the #899 cross-family matrix already hit **300-minute cell timeouts**
- `code-testing-agent` scenarios each run `npm ci` or a full pytest install and dominate the runtime

So a global bump risks converting a statistics problem into a timeout problem.

## Suggested approach

1. **Per-eval `runs` override.** Cheap analysis evals (`crap-score`, `detect-static-dependencies`, `test-smell-detection`) can take `runs: 3` for a small absolute cost. Expensive generative evals stay at 1 and instead grow scenarios. The A/A table above is the argument for prioritising this: the flips all occurred at n ≤ 9, and the two most stable skills were among the cheapest to run more often.
2. **Set a floor on n.** Nothing today prevents committing an n=1 eval, where one judge call is the entire verdict. `eng/eval-quality/check_eval_quality.py` already reports these; it could fail on `n < 3` for *new* evals once the existing 28 are addressed.
3. **Report the gate's own confidence.** The PR comment shows the CI but not that e.g. 3W/0T/0L cannot pass at n=3. Surfacing "won every trial, failed on power" would stop this being rediagnosed every run.

## Related

- `eng/eval-quality/README.md` (#953) documents the arithmetic and the observed cases.
- #945 raised several evals above the threshold by wiring up orphaned fixtures, which is the cheapest way to raise `n` without new CI cost — 8 orphaned fixtures remain repo-wide.
- #899 — the cross-family evaluation whose results first surfaced this.

Evangelink:
Fixes #952.

The eval gate flipped ~27% of its verdicts between runs on byte-identical inputs. I reproduced **all 22 verdicts** from the A/A test in #952 from the gate's formula alone, which isolates two independent defects — only one of which the issue names.

## Defect 1: the gate scored the judge's adjective

Trial scores are a five-point ordinal scale (`much-better` `+1.0`, `slightly-better` `+0.4`, `equal` `0`, …). A Student's-t interval over those reads the 0.4 → 1.0 step as *variance*, so a skill is punished for winning **more decisively**. Four wins and three ties over seven trials:

| trials | mean | ci_low | verdict |
| --- | ---: | ---: | --- |
| every win `slightly-better` | +0.229 | **+0.031** | ✅ |
| one win `much-better` | +0.314 | **−0.021** | ❌ |

Same record, better outcome, reversed verdict. `coverage-analysis` failed five consecutive runs while winning 100% of its trials, then passed on a sixth with the same 3W/0T/0L record — its scores were `[+0.4, +0.4, +1.0]` in a failing run and `[+0.4, +0.4, +0.4]` in the passing one.

## Defect 2 (not in the issue): no minimum sample size

`meanWithCi` returns `ciLow = mean` below two trials, so a single lucky judgment passed outright. Simulated at 200k runs/cell, **a skill with no real effect passes ~30% of the time at one trial** — and 18 of 94 gated evals sit there. The gate was simultaneously too strict at n=3 and a rubber stamp at n=1.

## The fix

**`adapt.mjs` reads only each trial's `winner`** and decides with an **exact one-sided sign test** at 5%. The verdict is now a deterministic function of the win/tie/loss record: identical records always produce identical results.

Collapsing to direction alone is *not* sufficient — I checked. A t-interval over win/tie/loss disagrees with the exact test on 12 records up to n=10, and in **every one it is the permissive side**, passing 4W/0T/0L, 4W/3T/0L and 6W/0T/1L at `p = 0.0625`. The exact binomial tail has no such gap, so the ported interval code is gone entirely.

**`MIN_CREDIBLE_TRIALS = 5` falls out of that test** rather than being chosen: the sign test cannot reach 5% below five discordant trials (`0.5⁴ = 0.0625`), and discordant trials can never exceed counted trials. Below it *no record can pass*, so the verdict is reported as **⚠️ underpowered** — never a pass, never a regression. That is what stops "won every trial, failed anyway" from being rediagnosed every run.

**`check_eval_quality.py` enforces the same floor on specs** before they ever run, counting `trials = scenarios × defaults.runs`. The 60 existing evals below it are grandfathered in a ledger that can only shrink: the gate errors on a stale, duplicated or unnecessary entry, and `--base-ref` (passed by CI on every PR) rejects entries that are *new* relative to the base branch — so a PR cannot add a below-floor eval and exempt it in the same change. Renames are read from git and aren't treated as growth.

**The experiment file no longer sets `runs` in `overrides:`.** Precedence is *CLI > experiment overrides > eval defaults* and the merge is a plain spread, so that line didn't *default* anything — it overwrote every eval's own `defaults.runs`, making the issue's first recommendation impossible to implement. No eval declares `defaults:` today, so this changes nothing until one opts in.

## Cost

**Zero additional CI minutes.** Raising `runs` on specific evals is deliberately left as follow-up — it's the only part that costs money, and it's now expressible. For the record, a targeted `runs = ceil(8/scenarios)` policy measures at **1.62× for `dotnet-test`** (~75 → ~121 min, inside the existing 180-minute timeout) and 1.82× repo-wide, versus 3.00× for the blanket `runs: 3` the issue rightly rejects.

## Also fixed (latent, found while testing)

- `--vally` paths containing a space were split into separate argv entries. This made the adapter test suite unrunnable on Windows, where `process.execPath` lives under `Program Files`. An unterminated quote now falls back to the old whitespace split so a bare apostrophe in a path can't be mangled either.
- Importing `adapt.mjs` made it parse its *importer's* command line.
- The `/evaluate` investigation prompt fired on every `passed == false`, which would have attached "To investigate failures" to essentially every comment once 60 evals started reporting ⚠️.
- Per-scenario rows ranked by magnitude could point ▼ while contributing a positive net win. All three renderers now read the same per-scenario record computed once in the adapter.

## What reviewers will notice

The PR comment gains **Net win** and **p** columns; `Δ Preference` becomes `Δ Pref` and is explicitly triage-only. **60 of 94 evals will render ⚠️ underpowered on the first run.** That is the honest state of the repo — none of them can currently produce a credible verdict — and ⚠️ fails no check. `eng/eval-quality/underpowered-allowlist.txt` is the ledger for burning it down.

## Verification

- 20/20 adapter tests, including a regression test that encodes the #952 defect directly (three score-vector pairs that differ only in magnitude must produce identical verdicts) and one asserting the gate ignores a deliberately negative `ciLow` from vally.
- 18/18 eval-quality self-tests, covering the trial floor, grandfathering, the shrink-only ratchet, rename handling, and an unresolvable base ref.
- `check_eval_quality.py` exit 0, `actionlint` clean on all three workflows, `bash -n`, `markdownlint` clean.
- Live `consolidate.mjs` render over four verdict shapes: new schema, **legacy schema** (pre-`netWin`/`signTest`), underpowered, and errored.

Three rounds of code review and design review; every finding is addressed in the diff.

Docs: `eng/eval-quality/README.md` gains failing check 7 plus a "Why the gate scores direction, not magnitude" section; `eng/vally-adapter/InvestigatingResults.md` documents the new verdict fields and the underpowered failure mode.

### DIFF
--- a/.github/workflows/eval-quality.yml
+++ b/.github/workflows/eval-quality.yml
@@ -4,9 +4,10 @@ name: eval-quality
 #
 # Each failing check corresponds to a defect that has already cost a real
 # evaluation result on this repo — see eng/eval-quality/README.md. All of them
-# are structural (file existence, git state, YAML keys), so the gate cannot
-# fire spuriously on well-written prose. Judgement calls such as statistical
-# power and orphaned fixtures are reported but never fail the build.
+# are structural (file existence, git state, declared counts, YAML keys), so the
+# gate cannot fire spuriously on well-written prose. Judgement calls such as
+# orphaned fixtures and grandfathered underpowered evals are reported but never
+# fail the build.
 
 on:
   pull_request:
@@ -14,18 +15,24 @@ on:
       - "tests/**"
       - "plugins/**"
       - "eng/eval-quality/**"
+      # check_floor_agreement() reads MIN_CREDIBLE_TRIALS out of the adapter, so
+      # the one diff that can break the floor's two-language agreement is a diff
+      # to the adapter. Without this the guard would never run on it.
+      - "eng/vally-adapter/**"
       - ".github/workflows/eval-quality.yml"
       - ".gitignore"
   push:
     branches: [main]
     # Kept in sync with the pull_request paths above. The gate reads plugins/*
-    # (skills with no eval) and .gitignore (fixtures excluded from the index),
-    # so omitting them here would let a direct push or a squash-merge that
-    # touches only those land on main without the gate ever running.
+    # (skills with no eval), .gitignore (fixtures excluded from the index), and
+    # eng/vally-adapter (the trial floor it must agree with), so omitting any of
+    # them here would let a direct push or a squash-merge that touches only
+    # those land on main without the gate ever running.
     paths:
       - "tests/**"
       - "plugins/**"
       - "eng/eval-quality/**"
+      - "eng/vally-adapter/**"
       - ".github/workflows/eval-quality.yml"
       - ".gitignore"
   workflow_dispatch:
@@ -45,6 +52,9 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
         with:
           persist-credentials: false
+          # The underpowered allowlist is shrink-only, which needs the base
+          # revision to compare against. Shallow history has no merge base.
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
@@ -60,4 +70,14 @@ jobs:
         run: python eng/eval-quality/selftest_eval_quality.py
 
       - name: Check eval quality
-        run: python eng/eval-quality/check_eval_quality.py
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        run: |
+          # On a pull request, additionally reject new entries in the
+          # underpowered allowlist relative to the base branch — otherwise a PR
+          # could add a below-floor eval and exempt it in the same change.
+          if [ -n "$BASE_REF" ]; then
+            python eng/eval-quality/check_eval_quality.py --base-ref "origin/$BASE_REF"
+          else
+            python eng/eval-quality/check_eval_quality.py
+          fi
--- a/.github/workflows/evaluation-run.yml
+++ b/.github/workflows/evaluation-run.yml
@@ -498,7 +498,7 @@ jobs:
         run: |
           echo "## 🔬 Vally Evaluation Results: $ENTRY_NAME" >> $GITHUB_STEP_SUMMARY
           echo "" >> $GITHUB_STEP_SUMMARY
-          echo "Skilled vs baseline, judged head-to-head by \`vally compare\`. A skill passes only on a complete comparison with a credible improvement (mean preference > 0 with its 95% CI above 0)." >> $GITHUB_STEP_SUMMARY
+          echo "Skilled vs baseline, judged head-to-head by \`vally compare\`. A skill passes only on a complete comparison with a credible net win — more wins than losses, by an exact one-sided sign test at p ≤ 0.05 — over enough trials for any record to reach that bar. ⚠️ marks a verdict that can't be reached: an underpowered eval, or a comparison that didn't complete." >> $GITHUB_STEP_SUMMARY
           echo "" >> $GITHUB_STEP_SUMMARY
 
           RESULTS_DIR="artifacts/TestResults/vally/$ENTRY_NAME"
@@ -507,9 +507,10 @@ jobs:
             EVAL_NAME=$(basename "$(dirname "$RESULTS_JSON")")
 
             CONCLUSIVE=$(node -e "const r=JSON.parse(require('fs').readFileSync(process.argv[1],'utf-8')); console.log(r.verdicts[0].conclusive)" "$RESULTS_JSON")
+            UNDERPOWERED=$(node -e "const r=JSON.parse(require('fs').readFileSync(process.argv[1],'utf-8')); console.log(r.verdicts[0].underpowered === true)" "$RESULTS_JSON")
             PASSED=$(node -e "const r=JSON.parse(require('fs').readFileSync(process.argv[1],'utf-8')); console.log(r.verdicts[0].passed)" "$RESULTS_JSON")
             REASON=$(node -e "const r=JSON.parse(require('fs').readFileSync(process.argv[1],'utf-8')); console.log(r.verdicts[0].reason)" "$RESULTS_JSON")
-            if [ "$CONCLUSIVE" != "true" ]; then
+            if [ "$CONCLUSIVE" != "true" ] || [ "$UNDERPOWERED" = "true" ]; then
               ICON="⚠️"
             elif [ "$PASSED" = "true" ]; then
               ICON="✅"
@@ -521,17 +522,29 @@ jobs:
             echo "$REASON" >> $GITHUB_STEP_SUMMARY
             echo "" >> $GITHUB_STEP_SUMMARY
 
-            # Per-scenario table: mean preference score and win/tie/loss.
-            echo "| Scenario | Mean preference | Trials (W/T/L) |" >> $GITHUB_STEP_SUMMARY
-            echo "|----------|-----------------|----------------|" >> $GITHUB_STEP_SUMMARY
+            # Per-scenario table: net win (the basis the gate reads), the
+            # magnitude-weighted preference for triage, and win/tie/loss. The
+            # arrow follows the net win so a row can never point the opposite
+            # way to the verdict it feeds. adapt.mjs computes these per
+            # scenario; the fallback covers results.json written before it did.
+            echo "| Scenario | Net win | Δ Pref | Trials (W/T/L) |" >> $GITHUB_STEP_SUMMARY
+            echo "|----------|---------|--------|----------------|" >> $GITHUB_STEP_SUMMARY
             node -e "
               const r = JSON.parse(require('fs').readFileSync(process.argv[1], 'utf-8'));
+              const dir = (t) => t.winner === 'treatment' ? 1 : t.winner === 'baseline' ? -1
+                : t.winner === 'tie' ? 0 : Math.sign(typeof t.score === 'number' ? t.score : 0);
+              const pct = (x) => (x >= 0 ? '+' : '') + (x * 100).toFixed(1) + '%';
               for (const s of r.verdicts[0].scenarios) {
-                const icon = s.meanScore > 0 ? '▲' : s.meanScore < 0 ? '▼' : '=';
-                const pct = (s.meanScore >= 0 ? '+' : '') + (s.meanScore * 100).toFixed(1) + '%';
-                let w=0,t=0,l=0;
-                for (const tr of (s.trials||[])) { if (tr.score>0) w++; else if (tr.score<0) l++; else t++; }
-                console.log('| ' + icon + ' ' + s.scenarioName + ' | ' + pct + ' | ' + w + '/' + t + '/' + l + ' |');
+                let { netWin, wins, ties, losses } = s;
+                if (typeof netWin !== 'number') {
+                  const counted = (s.trials || []).filter((t) => !t.errored);
+                  wins = counted.filter((t) => dir(t) > 0).length;
+                  losses = counted.filter((t) => dir(t) < 0).length;
+                  ties = counted.length - wins - losses;
+                  netWin = counted.length ? (wins - losses) / counted.length : 0;
+                }
+                const icon = netWin > 0 ? '▲' : netWin < 0 ? '▼' : '=';
+                console.log('| ' + icon + ' ' + s.scenarioName + ' | ' + pct(netWin) + ' | ' + pct(s.meanScore || 0) + ' | ' + wins + '/' + ties + '/' + losses + ' |');
               }
             " "$RESULTS_JSON" >> $GITHUB_STEP_SUMMARY
             echo "" >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -1110,8 +1110,14 @@ jobs:
             node eng/vally-adapter/consolidate.mjs --format simple --root all-results/ --output simplified-body.md
 
             INVESTIGATE_PROMPT=""
-            # If any skill failed, build a copy-paste prompt for AI-assisted investigation
-            if jq -se '[.[].verdicts[] | select(.passed == false)] | length > 0' $JSON_FILES > /dev/null 2>&1; then
+            # If any skill genuinely lost to its baseline, build a copy-paste
+            # prompt for AI-assisted investigation. Underpowered and inconclusive
+            # verdicts also carry `passed: false`, but neither is evidence that
+            # the skill regressed — attaching the prompt to those would fire on
+            # nearly every run and contradict the ⚠️ semantics of the comment
+            # body. The `!= false` / `!= true` form keeps results.json files that
+            # predate those fields classifying exactly as they did before.
+            if jq -se '[.[].verdicts[] | select(.passed == false and .conclusive != false and .underpowered != true)] | length > 0' $JSON_FILES > /dev/null 2>&1; then
               RUN_ID="${{ github.run_id }}"
               INVESTIGATE_PROMPT=$(printf '\n> **To investigate failures**, paste this to your AI coding agent:\n>\n> _For PR %s in %s, download eval artifacts with `gh run download %s --repo %s --pattern "vally-results-*" --dir ./eval-results`, then fetch https://raw.githubusercontent.com/%s/%s/eng/vally-adapter/InvestigatingResults.md and follow it to analyze the results.json files. Diagnose each failure, suggest fixes to the eval.yaml and skill content, and tell me what to fix first._' \
                 "${PR_NUMBER}" "${{ github.repository }}" "${RUN_ID}" "${{ github.repository }}" "${{ github.repository }}" "${{ needs.gate.outputs.head_sha }}")
--- a/dotnet-skills.experiment.yaml
+++ b/dotnet-skills.experiment.yaml
@@ -10,13 +10,26 @@
 #
 # The `agent.*` evals are excluded: they exercise orchestrator agents and do
 # not map to a single plugins/<plugin>/skills/<skill> directory.
+#
+# `overrides:` deliberately does NOT set `runs`. Precedence is
+# "CLI flags > experiment overrides > eval defaults", and the merge is a plain
+# spread (`{ ...eval.defaults, ...experiment.overrides }`), so an `overrides:
+# runs:` here does not set a *default* — it overwrites every eval's own
+# `defaults.runs` and makes per-eval trial counts impossible to express. An eval
+# that needs more trials than it has scenarios raises its own count:
+#
+#   defaults:
+#     runs: 3
+#
+# Trials per eval = scenarios x runs, and that product is what the pass gate's
+# confidence interval is computed over. See eng/eval-quality/README.md for the
+# floor and why it sits where it does.
 name: dotnet-skills
 evals:
   - tests/*/!(agent.*)/eval.yaml
 overrides:
   model: claude-opus-4.6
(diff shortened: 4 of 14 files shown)


Evangelink (commented):
/evaluate

Evangelink:
## Checked whether anything actually regressed: no

The first eval run on this PR showed 7 of 8 skills as ⚠️, which reads like mass breakage. It isn't. I re-ran **both gates over that run's actual trial data** (artifacts from run `30472542823`):

| skill | n | W/T/L | old mean | old ci_low | old gate | sign p | new gate | |
| --- | ---: | ---: | ---: | ---: | --- | ---: | --- | --- |
| `mcp-csharp-create` | 3 | 1/1/1 | +0.000 | −0.994 | ❌ | 0.750 | ⚠️ underpowered | |
| `mcp-csharp-publish` | 3 | 2/1/0 | +0.267 | −0.307 | ❌ | 0.250 | ⚠️ underpowered | |
| `mcp-csharp-test` | 3 | 2/1/0 | +0.667 | −0.768 | ❌ | 0.250 | ⚠️ underpowered | |
| `configuring-opentelemetry-dotnet` | 3 | 2/1/0 | +0.267 | −0.307 | ❌ | 0.250 | ⚠️ underpowered | |
| `dotnet-webapi` | 3 | 3/0/0 | +0.800 | −0.061 | ❌ | 0.125 | ⚠️ underpowered | |
| `minimal-api-file-upload` | 3 | 2/1/0 | +0.267 | −0.307 | ❌ | 0.250 | ⚠️ underpowered | |
| `technology-selection` | 6 | 5/1/0 | +0.433 | +0.097 | ✅ | 0.031 | ✅ **PASS** | stable |
| `convert-blazor-server-to-webapp` | 1 | 1/0/0 | +0.400 | +0.400 | ✅ | 0.500 | ⚠️ underpowered | **changed** |

**One verdict of eight changed**, and it is the case this PR exists to fix: `convert-blazor-server-to-webapp` has a single scenario, and below two trials the old gate computes no interval at all (`ciLow = mean`), so one judgment decided the verdict outright. That is the ~30% false-positive path from the PR description, caught live.

**Zero regressions.** The other six ⚠️ rows were **already failing** — identical W/T/L, `ci_low` well below zero under the old gate too. What changed is the *attribution*: "❌ no credible improvement" blamed the skill; "⚠️ underpowered" correctly says the eval is too small to have measured it at all. Note `dotnet-webapi` won 3/3 with baseline quality 1.9/5 → skilled 3.4/5 and still couldn't pass — that's the whole point of the issue.

## But the presentation invited exactly that misreading, so I fixed it

The old headline — *"1 improved, 0 no credible improvement, 7 underpowered"* — puts an unjudgeable eval next to real failures with no signal about which is which. Now:

> 8 skill(s) evaluated — ✅ **1 improved**, ❌ **0 no credible change**, 🔻 **0 regressed**.
>
> ⚠️ **7 could not be judged**: **7 underpowered** — the eval has fewer trials than any result needs to reach `p ≤ 0.05`, so no verdict was possible. This is the eval's size, **not a skill regression**; fix it by adding scenarios or raising `defaults.runs`.

`regressed` was already computed but never surfaced. It now has its own 🔻 icon and sorts ahead of a no-change verdict when detail blocks compete for the comment budget, so a real regression can't lose its place to an eval-size problem. "Did anything get worse?" is now answerable from the first line.

## Also in this push

Rebased onto `main` to resolve the conflict with #964 (which renumbered the failing checks and promoted the aggregate-`line-rate` check to an error). Resolution: the trial floor became check **8**, `T95` went away with `report_power` (the gate is an exact sign test now, so this side computes no interval), and #964's `output_case` helper is retained — its `t(n-1)` case had nothing left to assert, so it now asserts that the grandfathered warning reports the `scenarios × runs` arithmetic a contributor acts on.

20/20 adapter tests, 20/20 gate self-tests, gate exit 0, actionlint and markdownlint clean.

AbhitejJohn:
Thanks for the incredibly thorough write-up here — reproducing all 22 verdicts from the formula alone is a really clean way to isolate the two defects, and moving the gate to score direction over magnitude is clearly the right call. The sign-test core looks solid to me; I ran the tail values independently (`f(4,0)=0.0625`, `f(5,0)=0.03125`, `f(6,1)=0.0625`, `f(3,3)=0.65625`) and the `MIN_CREDIBLE_TRIALS = 5` derivation checks out, since `0.5^4 = 0.0625 > 0.05 >= 0.03125 = 0.5^5`.

A few things I'd love your take on before this goes in — all in the CI ratchet, none in the stats:

1. **Rename similarity in the shrink-only check.** `check_allowlist_growth` uses `git diff --find-renames` (defaults to 50% similarity) and treats any `R*` status as a non-growth rename. Curious whether that leaves a gap: could someone move a grandfathered underpowered eval to a new path, rewrite it down to ~50%, and have the new (effectively different) eval inherit the old exemption? The net exemption count stays flat, so I don't think the ratchet would flag it — which slightly undercuts the "shrink-only is a property of the gate, not of code review" argument you make in the description. Would pinning `--find-renames=100%` (or comparing content) close it without hurting the legit pure-rename path you're protecting?

2. **Allowlist-absent-at-base is a warning, not an error.** When the allowlist doesn't exist at `--base-ref` we warn and return. That's exactly right for the PR that *introduces* the file, but a delete-and-recreate in a single PR (lift the existing entries above the floor, then re-add the file with one new below-floor exemption) would land in the same branch and pass with warnings only. Worth turning that into an error when the current allowlist is non-empty?

3. **One test gap on the winner/score split.** Nice touch having `trialDirection` prefer `winner` over the derived `score`. The catch is the integration tests all build trials via `reportFromScores`, which derives `winner` *from* `score`, so the two never actually contradict on a `comparisonToVerdict` path — including the #952 regression test. If someone later reverted to `Math.sign(score)`, everything would still go green. Might be worth one test that hands `comparisonToVerdict` a trial with `winner: "baseline", score: 0.4` and asserts it lands as a loss.

None of these block the core fix as far as I can see — the ⚠️ underpowered behavior and the stats are the important part, and those look great. Happy to send a small PR against your branch for any of them if it saves you time.

AbhitejJohn:
A couple more things surfaced when I dug into what we're trading away here — none of these block the fix, but they feel worth hashing out before or soon after this lands.

1. **Sparse wins among many ties can clear the bar.** Since the sign test conditions on discordant trials, a record like `5W / 95T / 0L` passes at `p = 0.5^5 = 0.03125` — a 5% net win over 100 trials. That proves "*when* the skill makes a difference, it wins," which isn't quite the same as "the skill usually helps." Curious whether we want a secondary floor here — a minimum net-win-rate over *all* trials, or a minimum discordant count — so a skill that's indistinguishable from baseline on 95% of trials doesn't pass on 5 clean wins.

2. **We no longer hard-gate task-completion regressions.** The old skill-validator hard-failed if a skill regressed task completion in any scenario. The new verdict is purely the pairwise sign test, and while the trials still carry `baselinePassed`/`treatmentPassed`, they don't feed the decision. So `7W / 0T / 1L` passes at `p ≈ 0.035` even if that one loss is "baseline completed, skilled failed." I'm assuming `vally compare` makes the judge pick baseline when treatment breaks the task, so it shows up as a loss — but that's a soft, outweighable signal now, not the hard invariant we used to have. Do we want to bring back an explicit "never regress completion" check on top of the sign test?

3. **Reaching the trial floor via `runs` gives us correlated trials.** The natural way to escape ⚠️ underpowered is to bump `defaults.runs` on a scenario, but repeated runs of the *same* scenario (same prompt, fixture, judge) aren't independent, and the sign test assumes independence — so the p-value looks stronger than the evidence really is. More *scenarios* is the statistically honest path. The description already leans this way; might be worth making it explicit in the docs so folks don't just crank `runs` to turn a ⚠️ green.

More broadly, the two things I'd love to hash out with you (probably beyond this PR):

- **"By how much does this help?"** — collapsing to direction gives us determinism, but we lose any sense of effect size; a landslide and a hair's-breadth win look identical. Do we care about magnitude anywhere in the gate, or is pass/fail on direction genuinely all we want the eval to assert?
- **What `n` should a run be?** — the floor is 5, but the right per-eval trial count (scenarios × runs) that balances credible power against CI minutes is still an open question, and it's the part that actually costs money. Worth deciding deliberately rather than per-eval ad hoc.

Happy to prototype any of these against the branch if useful.

Evangelink:
## The red build was the new gate catching real drift, and it is green again

`eval-quality` failed on the previous push with three errors:

```
underpowered-allowlist.txt lists 'tests/dotnet-ai/mcp-csharp-create/eval.yaml',
which is not an eval spec in this repo. Remove the stale line.
```

…and the same for `mcp-csharp-publish` and `mcp-csharp-test`. It passed locally on the identical commit, which is the tell: **CI checks the merge with `main`, not the branch alone.** #943 landed in the meantime and retired those three skills, deleting their eval specs. The ledger still carried exemptions for evals that no longer exist.

That is the stale-entry check doing exactly its job, roughly two hours after it was written. It is also the reason that check is an error rather than a warning: a dead exemption is a path that a future eval could silently inherit.

Rebased onto `85cd10340` and dropped the three retired entries — the ledger is now **57 evals**, down from 60, purely from #943. Verified the way CI does it, including `--base-ref origin/main`.

I also added a line to the allowlist header, because this will happen again to whoever retires an eval next:

> Retiring an eval means deleting its line here too, in the same change. CI checks the merge with the base branch, so an entry left behind for a deleted eval fails the gate even though it passes on the retiring branch alone.

Worth noting the failure mode is self-healing rather than a trap: `eval-quality.yml` triggers on `tests/**`, so a PR that retires an eval without pruning the ledger fails on its own merge commit and gets told which line to remove.

All checks green: 20/20 adapter tests, 20/20 gate self-tests, gate exit 0, actionlint, lint, markdownlint, codeowners.

YuliiaKovalova (approved):

AbhitejJohn:
Thanks for landing this, @Evangelink — and nice touches in the final push, especially `check_power` becoming a real hard gate and the `check_floor_agreement()` drift guard reading `MIN_CREDIBLE_TRIALS` straight out of the adapter. That two-languages-can't-drift bit is the kind of thing that quietly saves you a year from now. The ref-resolution hardening in `check_allowlist_growth` is a good catch too.

Since it's merged, I don't want to relitigate anything — just jotting down the gaps from the earlier review that this change intentionally *didn't* close, so we have them tracked rather than rediscovering them later. All of these are about the verdict itself (`adapt.mjs`), which is unchanged, not the CI ratchet:

1. **Sparse wins can still pass on almost no evidence.** The underpowered floor counts *all* trials, ties included — `5W/95T/0L` has 100 counted trials, so it's nowhere near underpowered — but the sign test only sees the discordant ones: `signTestPValue(5, 0) = 0.5^5 = 0.031 ≤ 0.05`, direction "better", so it passes. So 5 wins and 95 "no meaningful difference" out of 100 clears the bar, because the 95 ties never enter the p-value. A magnitude-free net-win floor (e.g. `(wins − losses)/counted ≥ ~0.2`, counted incl. ties) would catch this — `0.05 < 0.2` — without touching the sign test or reviving the #952 variance problem. The exact threshold needs calibration though: a floor at 0.2 still lets `5W/20T/0L` through right at the boundary.

2. **No hard gate on task-completion regression.** The verdict is `conclusive && !underpowered && direction === "better" && credible` — `baselinePassed`/`treatmentPassed` don't enter it. So a change that's judged "slightly better" on style while actually breaking completion on a scenario can still pass, as long as the win/loss record leans positive. The old C# validator hard-failed on completion regression; that safety property didn't carry over. Worth a magnitude-free Layer-0 check (`baselinePassed && !treatmentPassed` on any scenario ⇒ fail).

3. **"By how much" is gone by design — do we want it fully gone?** This is the deliberate trade in the title, and I'm on board with dropping magnitude *out of the variance term* (that was #952's actual bug — ordinal scores feeding an sd/CI, where a bigger win lowered ciLow). But there's a middle path: a *deterministic* magnitude floor (mean ordinal ≥ some E_MIN, no variance) is monotone — a bigger win can only flip fail→pass, never pass→fail — so it recovers effect-size without the reversal. Might be worth it as report-only first, since the ordinal scale (−1/−0.4/0/+0.4/+1) is a policy weighting, not a true interval.

4. **Two things I'd still like us to hash out on the power side:**
   - *n / runs:* the gate's `n` is discordant trials, and `defaults.runs` on a single scenario feeds correlated pseudo-replicates into the sign test, which inflates significance. Aggregating each scenario to one outcome before the test (then testing across scenarios) would at least fix the *unit of analysis* — though scenario outcomes still share prompt, skill, judge and model, so that tightens independence rather than fully restoring it. What's the intended `runs` story here?
   - *reruns:* if authors can rerun until green, the nominal p-value stops meaning what it says. A fixed trial plan (or a stated rerun policy) closes that.

None of this is blocking anything that shipped — the fix is a clear net improvement over the ciLow gate. Happy to open a follow-up issue and prototype the two magnitude-free pieces (net-win floor + completion gate) against a backtest on the #952 A/A set and the current evals, if that's useful. Curious whether you'd already thought about the completion gate and left it out on purpose, or if it just fell off in the rewrite.

(Copilot, commenting on Abhitej's behalf.)

Issue #909, opened 2026-07-16
[skills-eval] Judge-side transient failures (disabled PAT + session.idle timeout) corrupt reliability verdicts


Evangelink:
## Summary

The `dotnet-test` cross-family evaluation (#899) flagged `writing-mstest-tests` and `detect-static-dependencies` as **P0 FIX-RELIABILITY** based on errored trials, with the assumption that the cause was *fixture nondeterminism* (unpinned SDK/tool versions).

Pulling the actual eval artifacts from run **29228914412** and capturing the stderr of all 3 errored trials shows the assumption is wrong: **every errored trial is a judge-side infra failure, not executor/fixture flakiness.**

## Evidence — the 3 errored trials

| Skill | Executor | Actual stderr |
| --- | --- | --- |
| `detect-static-dependencies` | haiku | `Comparison judge failed: CAPIError 400 This organization has been disabled` (disabled/throttled token in the `COPILOT_PAT` pool) |
| `writing-mstest-tests` | haiku | `Comparison judge failed: Timeout after 120000ms waiting for session.idle` (Vally internal judge idle cap) |
| `writing-mstest-tests` | sonnet46 | `Comparison judge failed: Timeout after 120000ms waiting for session.idle` |

All three are raised by the **comparison judge**, after the executor phase, not during fixture setup or build.

## Evidence the executor setup is fine (rules out fixture nondeterminism)

- **42/42 opus trials passed** across both skills — a nondeterministic fixture would not pass every opus trial.
- `detect-static-dependencies` **never builds** (0/30 `dotnet build` invocations) — it's a static-analysis skill, so an SDK/NuGet issue cannot be the cause.
- `writing-mstest-tests` ran `dotnet test` in **19/75 trials with zero SDK/NuGet errors**.
- One of the timed-out `writing-mstest-tests` scenarios has `reject_tools: [bash]` and **no fixtures at all** — there is literally no SDK/restore step that could fail.

## Schema gap

The Vally eval config (`eval.vally.yaml`) has **no per-eval judge timeout or retry knob**, so there's no way to make a single eval resilient to a transient judge error from the fixture side.

## Recommended fixes (harness-level)

1. **Judge-error retry** on transient failures — retry the comparison judge on `CAPIError` (429/5xx/"organization disabled") and `session.idle` timeouts before recording an errored trial.
[...]

Issue #901, opened 2026-07-15
[skills-eval] dotnet-upgrade: 6 skills, 41% pass — 1 P0, 1 P1, 2 to strengthen, 2 keep


AbhitejJohn:
> ## 🔄 EDITED 2026-07-19 — correction per #909
>
> **This finding has been revised.** The errored trials that drove the **FIX-RELIABILITY (P0)** flag(s) below were shown in [#909](https://github.com/dotnet/skills/issues/909) to be **judge-side infrastructure failures** — a disabled/throttled PAT (`CAPIError 400 organization disabled`) and Vally's `session.idle` judge timeout — **not** fixture nondeterminism. Verdict scores already exclude errored trials, so re-running the classifier with those judge errors dropped reclassifies the affected skill(s):
>
> | Skill | ~~Was~~ | Now |
> | --- | --- | --- |
> | `migrate-nullable-references` | ~~FIX-RELIABILITY (P0)~~ | **STRENGTHEN** |
>
> Struck-through text below is the **superseded** finding — in particular the "pin fixture SDK/tool versions" remediation does **not** apply. The title's P0 count has been updated. No eval re-run is required to correct the scores; recovering the lost judgments (optional) means **re-judging** those trials, not re-running the skill.

## Context — cross-family skill evaluation: `dotnet-upgrade`

This issue is **self-contained**: it captures everything a skill author needs to act on the `dotnet-upgrade` plugin without opening the full report.

**What this measures.** Every runnable skill in `dotnet/skills` was run through [Vally](https://github.com/microsoft/evaluate) (0.7) on a **cross-family matrix**: 5 executor model families — `opus-4.8`, `gpt-5.5`, `sonnet-4.6`, `haiku-4.5`, `mai-flash` — each judged by a **different family** (judge ≠ executor; default judge = latest Opus, or GPT when Opus is the executor). For every skill, a **skilled** run is compared against a **baseline** (no-skill) run and scored per executor. This removes single-model and self-judging bias, so a skill that only helps one model family — or only its own family's judge — is visible.

- **Data source:** cross-family CI grid (run `29228914412` + backfills) — 5 executors × 85 runnable skills, **419 scored cells** over 84 skills.
- **Row grain:** one row per **skill**, aggregated across its (up to 5) executor cells. `avgN` is the mean trial count behind the cells (trials 1–17; higher = more statistically trustworthy). `thin-N` flags directional-only rows.

### `dotnet-upgrade` at a glance (portfolio scorecard)

[...]

Issue #899, opened 2026-07-15
[skills-eval] dotnet-test: 14 skills, 37% pass — 1 P0, 1 P1, 6 to strengthen, 6 keep


AbhitejJohn:
> ## 🔄 EDITED 2026-07-19 — correction per #909
>
> **This finding has been revised.** The errored trials that drove the **FIX-RELIABILITY (P0)** flag(s) below were shown in [#909](https://github.com/dotnet/skills/issues/909) to be **judge-side infrastructure failures** — a disabled/throttled PAT (`CAPIError 400 organization disabled`) and Vally's `session.idle` judge timeout — **not** fixture nondeterminism. Verdict scores already exclude errored trials, so re-running the classifier with those judge errors dropped reclassifies the affected skill(s):
>
> | Skill | ~~Was~~ | Now |
> | --- | --- | --- |
> | `writing-mstest-tests` | ~~FIX-RELIABILITY (P0)~~ | **KEEP-POLISH (keep)** |
> | `detect-static-dependencies` | ~~FIX-RELIABILITY (P0)~~ | **KEEP-POLISH (keep)** |
>
> Struck-through text below is the **superseded** finding — in particular the "pin fixture SDK/tool versions" remediation does **not** apply. The title's P0 count has been updated. No eval re-run is required to correct the scores; recovering the lost judgments (optional) means **re-judging** those trials, not re-running the skill.

## Context — cross-family skill evaluation: `dotnet-test`

This issue is **self-contained**: it captures everything a skill author needs to act on the `dotnet-test` plugin without opening the full report.

**What this measures.** Every runnable skill in `dotnet/skills` was run through [Vally](https://github.com/microsoft/evaluate) (0.7) on a **cross-family matrix**: 5 executor model families — `opus-4.8`, `gpt-5.5`, `sonnet-4.6`, `haiku-4.5`, `mai-flash` — each judged by a **different family** (judge ≠ executor; default judge = latest Opus, or GPT when Opus is the executor). For every skill, a **skilled** run is compared against a **baseline** (no-skill) run and scored per executor. This removes single-model and self-judging bias, so a skill that only helps one model family — or only its own family's judge — is visible.

- **Data source:** cross-family CI grid (run `29228914412` + backfills) — 5 executors × 85 runnable skills, **419 scored cells** over 84 skills.
[...]

Issue #896, opened 2026-07-15
[skills-eval] dotnet-msbuild: 14 skills, 56% pass — 2 P0, 1 P1, 2 to strengthen, 9 keep


AbhitejJohn:
## Context — cross-family skill evaluation: `dotnet-msbuild`

This issue is **self-contained**: it captures everything a skill author needs to act on the `dotnet-msbuild` plugin without opening the full report.

**What this measures.** Every runnable skill in `dotnet/skills` was run through [Vally](https://github.com/microsoft/evaluate) (0.7) on a **cross-family matrix**: 5 executor model families — `opus-4.8`, `gpt-5.5`, `sonnet-4.6`, `haiku-4.5`, `mai-flash` — each judged by a **different family** (judge ≠ executor; default judge = latest Opus, or GPT when Opus is the executor). For every skill, a **skilled** run is compared against a **baseline** (no-skill) run and scored per executor. This removes single-model and self-judging bias, so a skill that only helps one model family — or only its own family's judge — is visible.

- **Data source:** cross-family CI grid (run `29228914412` + backfills) — 5 executors × 85 runnable skills, **419 scored cells** over 84 skills.
- **Row grain:** one row per **skill**, aggregated across its (up to 5) executor cells. `avgN` is the mean trial count behind the cells (trials 1–17; higher = more statistically trustworthy). `thin-N` flags directional-only rows.

### `dotnet-msbuild` at a glance (portfolio scorecard)

| Plugin | Skills | Cells | Pass | Impact | Tie-trials | Err | avg ΔTok | avg ΔTurns | avg ΔTools | Headline |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| **dotnet-msbuild** | 14 | 70 | 56% | 0.282 | 31 | 0 | +8,450 | +0.28 | +0.01 | Deepest, most reliable plugin; thin-N |

**Insights.** 14 skill(s); mean impact **0.28**; 9/14 help ≥1 frontier model. Exemplars (win incl. frontier): `msbuild-server`, `build-perf-baseline`, `build-perf-diagnostics`, `directory-build-organization`, `including-generated-files`. Weakest: `eval-performance` (-0.24). 

**Address first:** `msbuild-antipatterns` — discovery; `eval-performance` — regression; `check-bin-obj-clash` — cost.

### How to read the table

Each skill is scored **skilled vs. baseline** on these axes:

| Signal | Column | What it means | Good |
[...]

PR #972, opened 2026-07-30
Move eight P0/P1 skills to staging

y87feng:
## Summary

Move eight cross-family P0/P1 skills out of `main` while retaining their complete pre-removal versions on the long-lived `staging` branch.

This PR removes each selected skill together with its eval/fixture directory, then cleans up only the associated CODEOWNERS, external-dependency allowlist, eval-quality allowlist, and dashboard sample-data entries.

## Why these skills

The selection combines the latest corrected cross-family classifications (#893, #895, #896, #899, #901, including the #909 judge-infrastructure correction) with independently verified 30-day Copilot CLI usage from Kusto.

**Telemetry window:** 2026-07-02 11:11:59 through 2026-07-30 18:39:32. Automation accounts are accounts invoking at least 20 of the 92 skills; globally they account for 87.5% of invocations.

| Skill | Plugin | Eval action | Non-automation users | Non-automation invocations | Automation share | Reason to stage |
|---|---|---|---:|---:|---:|---|
| `android-tombstone-symbolication` | `dotnet-diag` | P0 `FIX-REGRESSION` | 2 | 4 | 99.9% | Regresses on both frontier models and has almost no organic usage. |
| `maui-collectionview` | `dotnet-maui` | P0 `FIX-REGRESSION` | 5 | 196 | 88.2% | Regresses about 30% of trials; usage is concentrated in a small audience. |
| `msbuild-antipatterns` | `dotnet-msbuild` | P0 `FIX-DISCOVERY` | 31 | 82 | 78.5% | Negative impact and only 30% eval invocation; stage for priority repair because real demand exists. |
| `eval-performance` | `dotnet-msbuild` | P0 `FIX-REGRESSION` | 5 | 9 | 97.4% | Regresses about 60% of trials with very little organic usage. |
| `check-bin-obj-clash` | `dotnet-msbuild` | P1 `TRIM-COST` | 11 | 18 | 96.6% | Positive but thin evidence and high execution overhead; stage until it is cheaper and revalidated. |
| `migrate-static-to-wrapper` | `dotnet-test` | P0 `FIX-REGRESSION` | 1 | 1 | 99.9% | Non-positive impact, frontier regressions, and effectively no organic adoption. |
| `migrate-dotnet9-to-dotnet10` | `dotnet-upgrade` | P0 `FIX-REGRESSION` | 34 | 77 | 98.4% | Regresses about 39% of trials; stage for priority repair because real demand exists. |
| `dotnet-aot-compat` | `dotnet-upgrade` | P1 `TRIM-COST` | 7 | 34 | 89.9% | Directionally useful but extremely expensive and based on thin eval evidence. |

`staging` is a quarantine/repair branch, not a deletion of history. Skills with demonstrated demand can be fixed and revalidated there before being proposed back to `main`.

## Explicitly retained on main

- All `dotnet-blazor` skills, including `plan-ui-change` and `support-prerendering`
- `microbenchmarking`, based on strong organic usage (31 users, 127 invocations, 7 users with 5+ invocations, 10 multi-day users)
- `code-testing-agent`, based on its exceptionally broad organic usage

## Validation

- `skill-validator check` across all plugins: **passed**
- Validated **88 skills, 16 agents, and 16 plugins**
- Reference scan: **0 errors**
- `git diff --check`: **passed**
- Confirmed all eight removed skills remain present on `origin/staging`


eng/dashboard/generate-token-data.ps1
```diff
@@ -18,11 +18,11 @@ $plugins = [ordered]@{
     "dotnet"         = @("csharp-scripts", "dotnet-pinvoke", "nuget-trusted-publishing")
     "dotnet-ai"      = @("technology-selection")
     "dotnet-data"    = @("optimizing-ef-core-queries")
-    "dotnet-diag"    = @("analyzing-dotnet-performance", "android-tombstone-symbolication", "dotnet-trace-collect", "dump-collect", "microbenchmarking")
+    "dotnet-diag"    = @("analyzing-dotnet-performance", "dotnet-trace-collect", "dump-collect", "microbenchmarking")
     "dotnet-maui"    = @("dotnet-maui-doctor")
-    "dotnet-msbuild" = @("binlog-failure-analysis", "build-parallelism", "build-perf-diagnostics", "incremental-build", "msbuild-antipatterns", "msbuild-modernization")
+    "dotnet-msbuild" = @("binlog-failure-analysis", "build-parallelism", "build-perf-diagnostics", "incremental-build", "msbuild-modernization")
```
AbhitejJohn, line 23: Thanks for pulling all of this together, folks — the config cleanup itself is clean (I checked all four modified files and there are no leftover entries for the eight skills).

One gap though: we drop `msbuild-antipatterns` (and `eval-performance` / `check-bin-obj-clash`) here, but several *retained* msbuild artifacts still point at them. A few concrete ones:
- `plugins/dotnet-msbuild/agents/msbuild-code-review.agent.md:53` has a hard markdown link `[msbuild-antipatterns skill](../skills/msbuild-antipatterns/SKILL.md)` — that target is deleted, so it's now a dead link.
- `msbuild.agent.md` and `build-perf.agent.md` still advertise `eval-performance` / `msbuild-antipatterns` / `check-bin-obj-clash` in their "available skills" lists, and the review agent leans on `msbuild-antipatterns` AP-13/AP-14/AP-21 by name.

Since these agents ship on `main`, they'll route the model to skills that no longer exist. The PR's "Reference scan: 0 errors" line only covers external URLs (that's all `ReferenceScanner.cs` checks), so it wouldn't catch any of this. Could we scrub or soften the inbound references in the same change? Happy to send a small pass against the branch if that helps.

(Copilot, commenting on Abhitej's behalf.)

eng/eval-quality/underpowered-allowlist.txt
```diff
@@ -71,13 +68,8 @@ tests/dotnet11/system-text-json-net11/eval.yaml
...
-tests/dotnet-maui/maui-collectionview/eval.yaml
 tests/dotnet-maui/maui-data-binding/eval.yaml
-tests/dotnet-maui/maui-dependency-injection/eval.yaml
 tests/dotnet-maui/maui-safe-area/eval.yaml
-tests/dotnet-maui/maui-shell-navigation/eval.yaml
-tests/dotnet-maui/maui-theming/eval.yaml
-tests/dotnet-msbuild/msbuild-antipatterns/eval.yaml
 tests/dotnet-template-engine/template-smart-defaults/eval.yaml
 tests/dotnet-test/find-untested-sources/eval.yaml
 tests/dotnet-test/generate-testability-wrappers/eval.yaml
```
AbhitejJohn, line 75: Heads-up while we're in this file: `check_eval_quality.py` is already red on `main` because these three (`maui-dependency-injection`, `maui-shell-navigation`, `maui-theming`) are stale underpowered exemptions that now clear the 5-trial floor. I verified it — clean `origin/main` reports 4 errors, this branch reports 3, and the one that drops is `maui-collectionview` (which this PR removes). So this PR isn't the cause and actually improves the count.

Since we're editing the allowlist anyway, want to drop these three stale lines too and get the gate green in the same change? None of them are in the eight-skill scope, so it's purely opportunistic — fine to skip if you'd rather keep this PR tightly scoped.

(Copilot, commenting on Abhitej's behalf.)

eng/dashboard/generate-token-data.ps1
```diff
@@ -18,11 +18,11 @@ $plugins = [ordered]@{
...
     "dotnet-test"    = @("migrate-vstest-to-mtp", "run-tests", "writing-mstest-tests")
-    "dotnet-upgrade" = @("dotnet-aot-compat", "migrate-dotnet9-to-dotnet10", "migrate-nullable-references")
+    "dotnet-upgrade" = @("migrate-nullable-references")
```
AbhitejJohn, line 25: Same inbound-reference nit on the upgrade side: `plugins/dotnet-upgrade/skills/migrate-dotnet8-to-dotnet9/SKILL.md:30` still tells users to "use the `migrate-dotnet9-to-dotnet10` skill as the next step", which we're removing here — so that hand-off now dead-ends on `main`. Curious whether we want to update it to point at `migrate-dotnet10-to-dotnet11` or just drop the pointer while the skill is on `staging`.

(Copilot, commenting on Abhitej's behalf.)

AbhitejJohn (changes requested):
## MMR verdict: 🟡 Request changes

The mechanical removal and config cleanup are correct and well-executed. The blocker is that **17 retained files still carry 31 references to the eight removed skills**, including one dead markdown link and a retained agent eval that mounts a deleted skill. None of this is disclosed in the PR body, and the "Reference scan: 0 errors" line doesn't actually cover it.

### What's proven correct ✅
- **Config cleanup is complete for the 4 modified files** — 0 leftover matches for any of the 8 skill names.
- **All 8 skills retained on `origin/staging`** — `git cat-file -e origin/staging:<path>` succeeded for all 8 `SKILL.md`.
- **No new eval-quality errors** — clean `origin/main` = 4 errors, PR HEAD = 3; the one that drops is `maui-collectionview` (removed here). The remaining 3 (`maui-dependency-injection`, `maui-shell-navigation`, `maui-theming`) are pre-existing on `main` and out of scope. This PR reduces the count.

### Findings ⚠️
1. **Retained agent eval mounts a deleted skill (highest severity).** `tests/dotnet-test/agent.testability-migration/eval.yaml:20` and `:50` list `../../plugins/dotnet-test/skills/migrate-static-to-wrapper` under `environment.skills`, but that directory is deleted. Probable/latent setup failure — if the harness skips missing mounts, the scenario is still semantically invalid since it no longer loads the skill it claims to test. (This file isn't in the diff, so it can't be inline-anchored.)
2. **Dead markdown link in a shipped agent** — `plugins/dotnet-msbuild/agents/msbuild-code-review.agent.md:53` links to the deleted `msbuild-antipatterns/SKILL.md`.
3. **Retained agents advertise removed skills** — `msbuild.agent.md` / `build-perf.agent.md` list `eval-performance` / `msbuild-antipatterns` / `check-bin-obj-clash` as available; the review agent cites `msbuild-antipatterns` AP-13/14/21 by name.
4. **Prose hand-offs dead-end** — e.g. `maui-data-binding/SKILL.md:38` → "use the `maui-collectionview` skill"; `migrate-dotnet8-to-dotnet9/SKILL.md:30` → "use the `migrate-dotnet9-to-dotnet10` skill". ~28 such references across the 17 files.
5. **Misleading validation line** — "Reference scan: 0 errors" is literally true but `ReferenceScanner.cs` only scans external URLs; it never validates cross-skill/relative links, so it gives no assurance about findings 1–4.

### Not blockers
- The 3 remaining eval-quality errors are pre-existing maui exemptions — optional opportunistic cleanup.
- Scope is otherwise tight: only the 8 skills' trees + their exact config entries.

Verified empirically on a worktree at PR HEAD; cross-checked with a second model family. Inline notes below.

steveisok:
I'd like to push back on staging `android-tombstone-symbolication`. I reproduced the eval and I think it's measuring the judge, not the skill.

**The skill works.** Against `tombstone_sample.txt`, the symbol server returns a real debug ELF and `llvm-symbolizer` resolves `0x222098 → ves_icall_System_Environment_FailFast (icall.c:6244)`. Verifiable ground truth.

**Head-to-head on Opus 4.8:**

- Baseline (no skill): resolved **0 of 7** frames. Tried three msdl URL shapes, all wrong, then asserted "public symbol servers returned 404 for that build-id" — which is false.
- Skilled: resolved **7 of 7**, plus runtime version, commit SHA, and root cause.

**A blind cross-family judge scored that as `much-better` for the failing baseline**, on the grounds that the correct answer was "highly specific, unsupported." It confirmed it could not verify the symbols, could not check the false 404 claim, and could not detect fabrication. That's a −1.0 on the one stimulus where the skill is unambiguously right, and it explains the `frontier-regressed` flag on its own.

**Eval defects:**

- 6 of 8 fixture BuildIds are fabricated and 404 on msdl — only 2 stimuli are completable end-to-end.
- Skill-free Opus scores 3/3 on `no_dotnet`, `no_buildid`, and `crashlog_ios` in 38s, matching the skilled rubric verbatim. Those are coin flips, not evidence.
- `reject_skills: ["*"]` on the iOS stimulus is unfailable for baseline, failable only for skilled.
- `runs` is unset, so the sign test pools 8 different tasks as if they were repeated measures of one effect.
- The 6 deterministic `output-matches` graders separate skilled from baseline perfectly, but `adapt.mjs` derives wins/losses solely from the blind preference judge.

**The label is also inconsistent with the harness.** From #893's own numbers (~33 trials, 13L, 2T) that's ~18 wins, and `adapt.mjs` sets `regressed` only when `losses > wins`. The P0 rests on Impact 0.16 vs a 0.4 bar — but `InvestigatingResults.md` says "Do not read `meanScore` here… a verdict never turns on it." The artifacts from run `29228914412` have expired, so the judge reasoning behind those losses can't be re-read.

On usage: 30-day invocation counts are the wrong instrument for incident-driven crash triage. Its sibling `apple-crash-symbolication` is already on the underpowered allowlist — the whole symbolication category is being measured by something that can't see it.

Suggest dropping this skill from the PR and fixing the eval instead: real BuildIds in every fixture, `defaults.runs >= 3`, reweight or remove the guardrail stimuli baseline aces, and let the judge see the deterministic grader outcomes.

AbhitejJohn:
Re: @steveisok's point that `android-tombstone-symbolication` measures the judge rather than the skill — I re-ran it cross-family on the latest `main` (5 executors, judge ≠ executor family) to regenerate the evidence, since the original run's artifacts had expired. **The concern holds up.**

Run: https://github.com/dotnet/skills/actions/runs/30862999210

| Executor | Judge | W/T/L | Mean pref | Result |
|---|---|---|---|---|
| opus | claude-opus-4.8→gpt-5.5 | 3/1/3 | **+17.1%** | inconclusive (1 unmatched trajectory) |
| gpt | gpt-5.5→claude-opus-4.8 | 3/0/5 | −2.5% | no improvement (flat, p=0.36) |
| sonnet46 | →claude-opus-4.8 | 5/1/1 | **+31.4%** | inconclusive (1 unmatched trajectory) |
| haiku | →claude-opus-4.8 | 6/2/0 | **+52.5%** | **credibly better** |
| mai | →claude-opus-4.8 | 5/0/2 | **+25.7%** | inconclusive (1 unmatched trajectory) |

Takeaways:

- **No credible regression on current `main`.** 4/5 executors are directionally positive (+17% to +52%); the only non-positive one (gpt) is statistically flat (−2.5%, sign-test p=0.36, 95% CI spans 0), and `regressed=false` on all five. The original "P0 FIX-REGRESSION" verdict does **not** reproduce — consistent with the gate fix in #952/#965 ("score direction, not magnitude").
- **The apparent fails are harness artifacts.** opus/sonnet46/mai each flip to *inconclusive* solely because one trajectory couldn't be paired (unmatched), not because the skill lost. The sign test's unit is the stimulus (~8 of them), so it needs near-unanimity — a single dropped trajectory (8→7 pairs) decides the verdict. That's fragility in the measurement, not signal about the skill.
- **There is a real, fixable eval bug.** In the "no BuildId" stimulus the skilled run fabricated a BuildId borrowed from an unrelated fixture in the shared directory; the judge correctly flagged it as a dangerous hallucination. That's fixture cross-contamination, not a stable skill regression.
- **In fairness to the eval:** where the judge *can* verify ground truth, it does reward the skill — on the real symbolication tasks the skilled run downloaded correct symbols by BuildId and produced full function/line tables while baseline used the wrong URL, 404'd, and gave up. So the judge isn't blind; the verdict is just dominated by unmatched-trajectory fragility + fixture contamination + marginal presentation nitpicks (the judge repeatedly notes "both correct, gap is small").

Recommendation: **drop `android-tombstone-symbolication` from this staging PR** and fix the eval instead — dedupe fixtures/BuildIds so they can't cross-contaminate, drop the baseline-unfailable iOS guardrail stimulus, and raise power via more distinct stimuli + `runs ≥ 3`. Precision on a biased fixture just yields a precisely-wrong number, so the fixes should come before any re-measurement.

I'm re-running the other skills being staged in this PR to check whether any of them fail for the same measurement reasons, and will follow up with those results.

AbhitejJohn:
## Cross-family re-validation of the staged skills (latest `main`)

Re-ran the staged skills cross-family on latest `main` (5 executors — opus/gpt/sonnet46/haiku/mai — judge ≠ executor family). Runs: [android-tombstone](https://github.com/dotnet/skills/actions/runs/30862999210), [other 6](https://github.com/dotnet/skills/actions/runs/30882049180). Verified these still reflect current `main` (only intervening commit is #969, which touches none of these skills/evals).

**Finding: most staging verdicts measure the eval instrument, not the skill.** Tracking issue: #986.

| Skill | Original | Re-run verdict | Re-category |
|---|---|---|---|
| android-tombstone-symbolication | P0 FIX-REGRESSION | 4/5 positive, `regressed=false` | eval-defect (biased + underpowered) — fix in flight |
| check-bin-obj-clash | P1 TRIM-COST | n=1 → can't pass by construction | eval-defect (underpowered) |
| eval-performance | strengthen | n=1 → can't pass by construction | eval-defect (underpowered) |
| msbuild-antipatterns | P0 | neutral; 3/4 stimuli off-target (F#) | eval-defect (construct-invalid) |
| migrate-static-to-wrapper | P0 FIX-REGRESSION | positive; sonnet46 credibly better, 0 regressions | **keep — skill is helpful** |
| migrate-dotnet9-to-dotnet10 | P0 FIX-REGRESSION | helps haiku (+0.66); gpt dip is polish-only (both correct) | real but overstated → P2 rubric fix |
| dotnet-aot-compat | P1 TRIM-COST | n=1 → can't pass by construction | eval-defect (underpowered) |
| maui-collectionview | P0 | fixed + revalidated in merged #947 | **stale — drop** |

Why the "fails" aren't regressions: three evals have a single stimulus (one trial → sign test `p=0.5` always, can never pass); one amplifies marginal "both-correct" losses into a "regression" via direction-based sign test; one has fixture/grader/guardrail bias; one tests off-target scenarios. Details per plugin in #896 / #899 / #901 / #895.

### Suggested changes to this PR
- **Reframe** from "shelve 8 P0/P1 skills" → "fix these evals and re-baseline." The defect is mostly the instrument.
- **Drop:** `maui-collectionview` (stale) and `migrate-static-to-wrapper` (skill is helpful).
- **Keep skill, fix eval (don't stage on pass-rate):** android-tombstone (in flight), check-bin-obj-clash, eval-performance, msbuild-antipatterns, dotnet-aot-compat.
- **Only genuine follow-up:** `migrate-dotnet9-to-dotnet10` — and that's a rubric-tightening ("both correct ⇒ tie"), not a P0 shelve.
- If **cost** is the real driver for the TRIM-COST skills, state it explicitly and separately from the (invalid) pass-rate argument.

Net: of the 8, 0–1 warrant staging on the original premise; the rest are eval repairs tracked in #986.

AbhitejJohn:
Closing this PR as discussed.

PR #974, opened 2026-07-31
Fix migrate-mstest-v1v2-to-v3 activation and test-skill eval quality

Evangelink:
## Summary

Two related pieces of test-skill quality work. They can be split if you'd rather review them separately — the commits are clean (`f9dc25d` skill activation, `2fc8ab8` eval quality).

### 1. `migrate-mstest-v1v2-to-v3` never activated on five of its own stimuli

The cause is the frontmatter `description`, which is the only text the runtime uses to select a skill:

- **`DO NOT USE FOR: ... projects already on MSTest v3+`** actively told the router not to load the skill. Three of the failing fixtures (`fix-assert-areequal-*`, `fix-datarow-*`, `migrate-from-testsettings-*`) intentionally ship a project already referencing `MSTest 3.8.0`, because they test the breaking changes that only surface *after* the package bump. The description excluded exactly the cases the skill exists to handle.
- The description gated the whole skill on "the user asks to upgrade MSTest", so the `.testsettings` to `.runsettings` prompt, which never says "upgrade", matched nothing.
- `migrate-mstest-v3-to-v4` advertised the generic phrase "tests don't compile after upgrading MSTest", which matches "I upgraded from v2 to v3 and now tests don't compile" better than the correct skill did.
- Even on activation, the skill's own *Boundary Gate* ("if all MSTest references are already 3.x and no error is reported, make no changes") would have made the agent refuse the `.testsettings` conversion and fail its `file-exists` / `file-not-exists` graders.

The description is rewritten around both entry points the skill actually serves (pre-upgrade migration *and* post-upgrade breaking-change fixes) and now carries the tokens those prompts contain: `CS1501`/`CS1503`/`CS0121`, `Assert.AreEqual`/`AreNotEqual`/`AreSame` object overloads, `DataRow` strict typing (`1L` vs `1`), `MSTEST0014`, `.testsettings`/`LegacySettings`, `DeploymentEnabled`, per-test `TestTimeout`, dropped `net5.0`. "Keeps the current runner" stops "migrate to v3 but keep VSTest" being pulled toward `migrate-vstest-to-mtp`.

`migrate-mstest-v3-to-v4` is narrowed to `3.x to 4.x` and now explicitly excludes leftover 2.x-to-3.x errors; none of its own 15 stimuli depend on the vague phrasing removed. The Boundary Gate now says a 3.x package version alone does not end the migration while v1/v2-era settings or errors remain — the `already-v3` stimulus still passes, since that fixture has clean v3 packages and no `.testsettings`, so all three gate conditions still hold. `agents/test-migration.agent.md` gets a matching routing row so the plugin arm routes like the isolated arm.

Worth a look: the v3-to-v4 description was at 1022 of the 1024-character spec limit, so absorbing the new exclusion meant compacting existing wording ("constructor" → "ctor", "out parameter" → "out param"). No trigger keywords were dropped, but that description now has essentially no headroom.

### 2. Eval-quality gate findings for the test skills

The gate reported three classes of debt against `dotnet-test` and `dotnet-test-migration`.

**Underpowered evals (5).** Below five trials the pass gate's sign test cannot reach `p ≤ 0.05` at any effect size, so these evals could never return a verdict at all. Each is now at or above the floor, with its allowlist line deleted in the same change as the shrink-only ledger requires. New scenarios target behaviour the skills document but nothing measured:

| Eval | Before → after | Added |
|---|---|---|
| `coverage-analysis` | 3 → 5 | refactoring-safety ranking (the "is this safe to change?" use case named in the skill's Purpose) and branch-vs-line coverage; both reuse the existing `partial-coverage` fixture |
| `find-untested-sources` | 4 → 5 | a mixed C#/TypeScript repo — the only case that exercises the documented polyglot-vs-Roslyn engine choice; composed from two existing fixtures |
| `generate-testability-wrappers` | 4 → 5 | the ambient-context path for a project with no DI container, where `AsyncLocal<T>` and scoped disposal are the distinguishing content |
| `grade-tests` | 4 → 5 | a C# case with production code **present** — every prior C# scenario hides it, so `Unverified` was never tested as a negative, and the D band and swallowed-exception F were never graded |
| `code-testing-agent` | 3 → 6 | `defaults.runs: 2` |

`code-testing-agent` is the one that buys trials with `runs` rather than scenarios. Scenarios are normally strictly better, but each of its three drives a full generate-build-test pipeline — `npm ci` plus two Vitest runs, `pip install` plus pytest, and a `dotnet test` build — under a 60m budget, which is the documented exception.

**Orphaned fixtures (5).** `v3-sealed-timeout`, `mtp-mstest-sdk9`, `mtp-mstest-sdk10`, `mtp-mstest-hotreload-installed` and `vstest-mstest` are superseded first-generation copies; their per-scenario successors differ only in whitespace, a dropped `rollForward`, or a package version. Both owning evals are already well above the floor, so wiring them up would add no power. Deleted.

**Skills with no eval (2 of 4 addressed).** `platform-detection` and `filter-syntax` carry real rules that nothing measured, several counterintuitive enough that a baseline is likely to get them wrong: `global.json` `test.runner` outranking `TestingPlatformDotnetTestSupport` on .NET 10+, `Microsoft.NET.Test.Sdk` not being a VSTest signal, MTP properties living in `Directory.Build.props`, and xUnit v3 dropping VSTest `--filter` while MSTest on MTP keeps it. Both get a 5-scenario eval with small fixtures and no build step.

`code-testing-extensions` and `test-analysis-extensions` are **left flagged deliberately**. Their bodies are tables of paths to extension files, so a head-to-head eval would score path recall rather than user value — gaming the warning rather than fixing it. The content those files hold is already exercised through `code-testing-agent`'s three-language pipeline. The warning is a judgement call by design, so I'd rather leave it visible than answer it with a hollow eval.

**Not touched:** the gate's four ERRORS are stale `dotnet-maui` allowlist lines for evals that now meet the floor. They pre-date this branch (reproduced on a clean tree before any edit) and belong to whichever change raised those MAUI evals. Each is a one-line deletion if you'd like them folded in here.

## Related issue

N/A

## Validation

- `python eng/eval-quality/check_eval_quality.py` — test-skill warnings cleared: underpowered list no longer contains any `tests/dotnet-test/*`, orphaned fixtures 8 → 3 (all remaining are `dotnet-upgrade`), skills-without-eval 5 → 3. Error count unchanged at the same 4 pre-existing MAUI lines, so nothing new was introduced.
- `python eng/eval-quality/check_eval_quality.py --base-ref origin/main` (what CI runs on a PR) — same 4 errors, and no "new exemption" error, confirming the ledger only shrank.
- `python eng/eval-quality/selftest_eval_quality.py` — all 20 self-tests pass.
- Every touched and new eval re-parsed: trial counts confirmed (5/5/5/5/5/5 and 3×2=6), all grader regexes compile, no grader has a missing or empty required config, and every referenced fixture path resolves on disk. New fixtures are staged in git so they exist on the CI runner.
- Skill descriptions verified within the 1024-character agentskills.io limit after YAML folding (1019 and 1012), and the plugin's rendered `<available_skills>` footprint modelled with the validator's own `RenderedSkillMenuCost` formula is 4,516 of the 15,000 budget, so nothing gets truncated out of the menu.
- The skill-validator itself could not be run here: `dotnet run --project eng/skill-validator/src` fails restore downloading `@github/copilot-win32-x64` with a TLS handshake failure. The description-length and menu-budget rules were reproduced manually instead. Eval *results* have to come from CI either way.

## Checklist

- [x] I updated CODEOWNERS when adding or moving owned content.
- [x] I updated `eng/known-domains.txt` for any new external domains referenced by skill content.

Notes on the last four: no `CODEOWNERS` entry applies (new files land under existing `tests/dotnet-test/` ownership), no plugin metadata changed (`plugin.json` untouched), and no new external domains are referenced. On focus — this is two commits covering skill activation and the gate's test-skill findings; both are dotnet-test-area quality fixes, but they are separable if you'd prefer two PRs.


Evangelink (commented):
/evaluate

Evangelink:
Dug into the eval artifacts from run 30610971796. The `❌` is real but narrow: **7W/2T/2L, p=0.090**. The sign test needs `p ≤ 0.05`, so turning even one loss into a win or tie clears it (8W/2T/1L → p=0.0195; 9W/2T/0L → p=0.002). Both losses trace to skill content that actively misled the agent, and the transcripts show exactly how. Pushed 6b11ad9 to fix both.

### Loss 1 — `Migrate MSTest v1 project with assembly reference`: skilled **0.00** vs baseline **4.17**

The skill loaded fine. The agent's entire answer was:

> To give you specific migration steps, I need to see your project file. Could you share the path to your `.csproj` (or `.sln`)? I'll read it and perform the migration.

`TestProject.csproj` and `OrderServiceTests.cs` were sitting in the working directory. Judge: *"A provides a complete, actionable answer. B just asks for more info without providing any guidance."*

Cause: the Inputs table listed `Project or solution path | **Yes**`. Marked required, it reads as a precondition the agent has to obtain before doing anything — so it asked instead of looking. This is the worst failure mode for a real user: describe your project in prose, get a question back.

Fix: path is optional and discovered by globbing, Step 1 leads with locating the project, and a note forbids opening with a request for the path. **The same `Required: Yes` trap was in `migrate-mstest-v3-to-v4` and `migrate-vstest-to-mtp`**, so both are corrected too.

### Loss 2 — `Fix DataRow type mismatch errors`: skilled **3.96** vs baseline **5.00**

The breaking-changes table said the 16-argument `DataRow` cap was *"fixed in later v3 versions"* and advised *"refactor test / wrap extra params in array"*. On a project already at **3.8.0**, the agent took that as instruction and rewrote a perfectly valid 17-argument row. From the transcript:

> **DataRow >16 params**: 17 args exceeds the constructor limit → wrap extras in `params object[]`

…then `new object[] { 17 }` failed, and it second-guessed itself mid-run:

> Wait — the `new object[] { 17 }` approach won't work… But let me check if the latest 3.x actually fixed the 16-arg limit.

It finally settled on a `(object)17` cast. Churn and wasted turns on code that was already correct.

The vagueness was the defect, so it's replaced with the fact: the cap was introduced in **3.0.1** and removed again in **3.0.3** ([microsoft/testfx#1554](https://github.com/microsoft/testfx/issues/1554), where the maintainer closes with *"please feel free to update to `3.0.3`"*). On 3.0.3+ a longer row is valid and must be left alone. Added alongside it: confirm the diagnostic before editing, because rewriting valid code to dodge a limit the project isn't subject to is a defect, not caution.

### On the ⚠️ 8/11 activation

Three scenarios didn't invoke the skill — and all three **won anyway**, because the agent handled them unaided (the `.testsettings` run, for example, produced a correct `local.runsettings` with zero skill involvement). So those wins are baseline capability, not skill value.

I deliberately did **not** keyword-stuff the description to force activation. It already names `.testsettings`/`.runsettings` and the dropped-TFM cases explicitly, the description is at 1019/1024 chars, and the overfit judge currently reads ✅ Low (0.15). Pushing harder there would trade a real signal for a scored one. If these stay non-activating across runs I'd rather treat it as a retrieval question than a description-length arms race.

### Notes

- No eval prompt, fixture, or grader was touched — both fixes are about what the skill tells a real user, so the improvement has to show up against the unchanged suite.
- Skill grows ~480 tokens, still `standard` tier and below the 5,000-token warning threshold.
- The two ties are not worth chasing: `Fix Assert.AreEqual` has a **5.00 baseline** (nothing to win), and the complex-migration tie was a position-swap disagreement (forward: skilled, reverse: tie) despite skilled scoring 4.89 against a 2.95 baseline — judge noise, not a content problem.

Evangelink:
## We did not regress — those two runs were an A/A comparison

Run 1 (`30610971796`) and run 2 (`30612389899`) tested **byte-identical skill content**:

- `git diff f9dc25d 2fc8ab8 -- plugins/` is **empty** (the second commit touched only `tests/` and `eng/`).
- The run-2 artifact confirms it directly: the `skill.invoked` payload does **not** contain the guidance added in `6b11ad9`.

So `7W/2T/2L` → `4W/5T/2L` is judge noise on the same input — the A/A instability this repo documents from #952. Neither run included the fix.

One thing *is* reproducible across both runs: `Migrate MSTest v1 project with assembly reference` scored **0.00 in both**, with the identical reply *"I need to see your project file."* That's the real defect, and `6b11ad9` addresses it.

## Then I ran the fixtures against a real MSTest 3.8 project, and found something worse

**The skill was teaching two things that are false, and one fixture couldn't reproduce the bug it was named for.**

### 1. `Assert.AreEqual(object, object)` — the "compile error" is mostly a myth

The skill claimed the removal causes *"compile error on untyped assertions"*. It doesn't. v3 keeps `AreEqual<T>(T?, T?)`, so two `object`-typed arguments infer `T = object` and compile untouched. Proof — the **shipped fixture, unmodified, on MSTest 3.8**:

```
Passed!  - Failed: 0, Passed: 3, Skipped: 0, Total: 3
```

The scenario's prompt claims *"several of my tests no longer compile... CS1501"*. That is fiction, and it explains the **baseline 5.00/5.00 in both runs**: there was nothing to fix, so neither arm could distinguish itself.

The break is real but narrow — only where `T` can't be inferred, and the true diagnostics are **CS0411** and **CS1503**:

```
error CS0411: The type arguments for method 'Assert.AreEqual<T>(T?, T?)'
cannot be inferred from the usage.
```

I have to own part of this: the description I added last round listed `CS1501/CS0121`, which I'd copied from the eval prompt rather than verified. Both are wrong. Corrected.

This matters beyond scoring: following the old text, an agent rewrites assertions that were already correct — the same over-application defect as the DataRow one.

### 2. `DataRow` mismatches are not compile errors either

Verified: they build with `MSTEST0014` and fail at **run time** with *"Test data doesn't match method parameters."* Widening (`int` → `long`) still binds; narrowing doesn't. The skill now says so explicitly, because a green build is precisely what misleads here.

### 3. Fixtures rebuilt so they reproduce the real breakage

`ComparisonTests.cs` now uses two unrelated interface-typed views of one instance — **CS0411 on all three assertions**, and **4/4 passing** once `<object>` is added. It also gains two already-valid typed assertions that must be left alone; widening those still compiles, so only judgement prevents it, and two graders now check it.

`v2-nuget/UserServiceTests.cs` and `v2-complex/InventoryServiceTests.cs` had the same flaw — their graders demanded `Assert.AreEqual<object>` on assertions that never needed it, which would have **contradicted the corrected skill and made it lose**. Both rebuilt and verified in three states:

| Fixture | On 2.2.10 (shipped) | After v3 bump | Fully migrated |
|---|---|---|---|
| `v2-nuget` | builds | CS0411 ×2 | 5/5 pass |
| `v2-complex` | builds | CS0411 ×3 | 7/7 pass |
| `fix-assert-*` | n/a (ships on v3) | CS0411 ×3 | 4/4 pass |

Every claim here was verified by building and running against MSTest 3.8 and 2.2.10, not inferred.

## Note on the ceiling scenarios

`Fix Assert.AreEqual` and `Fix DataRow` both sat at **baseline 5.00 in both runs** — structurally incapable of showing skill value, which `InvestigatingResults.md` pattern 8 calls out. I fixed the *cause* (a fixture that didn't reproduce its bug) rather than deleting the scenarios or loosening graders. The Assert scenario is now strictly harder than before.

I still haven't keyword-stuffed the description to chase the ⚠️ 8/11 activation. Three non-activating scenarios won anyway, which says those wins are baseline capability, and the overfit judge reads ✅ Low.

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink:
Merge conflicts resolved and the last remaining loss diagnosed.

## The fixes are working

`migrate-mstest-v1v2-to-v3` went from **7W/2T/2L, p=0.090** → **6W/4T/1L, p=0.063**. All three scenarios I targeted flipped:

| Scenario | Before | Now |
|---|---|---|
| `Migrate MSTest v1 project with assembly reference` | ▼ -100% (skilled **0.00**) | ▲ +100% |
| `Fix Assert.AreEqual object overload errors` | = tie (baseline 5.00 ceiling) | ▲ +100% |
| `Fix DataRow type mismatch errors` | ▼ -100% | ▲ +100% |

Baseline also dropped 3.9 → 3.6, which is the rebuilt fixtures doing their job: they now actually reproduce the bugs, so the unaided arm has something real to fail at.

`p=0.063` is one trial short. With 1 loss the gate needs 7 wins (`p=0.035`); four of the current ties are ceiling scenarios where baseline scores 5.00.

## The last loss: the agent rebuilt the project from scratch

`Migrate complex MSTest v2 project` scored **0.23 vs 3.07**. The transcript shows why, and it is not a knowledge gap:

1. glob succeeded, returning `./TestProject.csproj`, `./InventoryServiceTests.cs`, `./local.testsettings`
2. the agent rebuilt those into absolute paths **under the skill's own base directory** — all three reads failed `Path does not exist`
3. it globbed that directory, found only `SKILL.md`, and concluded *"There's no actual project on disk"*
4. it then **scaffolded a substitute project from the prose description** — fewer tests, no 17-parameter row, plus a self-introduced bug it had to debug

Judge: *"Response B couldn't find the project files, created its own simplified project from scratch."*

This is partly my own doing: `6b11ad9` told the agent to glob but not what to do with the answer, and the skill-context header advertises a "Base directory for this skill". `1079c9b` closes it — open paths exactly as returned, treat a failed read of a just-found file as a wrong constructed path, never conclude the project is missing while a search is still listing it, and never scaffold a replacement. That last rule matters well beyond this eval: any plugin-loaded skill has a base directory that is not the user's repo.

## Conflicts — resolved toward main

#976 landed independently and overlapped this branch.

- **`filter-syntax/eval.yaml` (add/add)** — both sides wrote it from scratch. **Took main's**, wholesale, and deleted my four fixture dirs. Main's fixtures carry real test classes rather than bare project files, so the scenarios read like a repo instead of a lookup quiz.
- **`underpowered-allowlist.txt`** — main still lists `find-untested-sources`; this branch raises it to 5 scenarios, so the line has to go under the shrink-only rule. Kept the removal.
- **De-duplicated a scenario neither side knew about:** we both added an ambient-context case to `generate-testability-wrappers`. **Main's is stronger** — a shipped NuGet library with locked signatures and a parallel suite, versus my plain console app — so mine is dropped. I folded in its one unique guard: a `ThreadStatic` negative grader plus rubric line, since `[ThreadStatic]` is the specific wrong answer that doesn't flow across `await` and so breaks exactly the parallel-safety property main's prompt demands.
- **`grade-tests` kept both sides** — main's 62-test readability case and my D/F-band case with production code present test different things.

## Gate is now fully green

The four `dotnet-maui` allowlist errors I'd flagged as pre-existing are fixed in main, so:

```
No errors.   (exit 0, with and without --base-ref)
```

Every non-`agent.*` eval under `dotnet-test` and `dotnet-test-migration` is now at or above the 5-trial floor.

Evangelink (commented):
/evaluate

YuliiaKovalova (approved):

PR #979, opened 2026-07-31
Capture the repo's hard-won skill-authoring lessons as guidance

Evangelink:
## Why

Our skills score well, but the reasons live in PR bodies and in a handful of engineers' heads. This mines ~35 merged PRs (#830–#976), issue #899, and `eng/eval-quality/README.md` into guidance contributors can actually follow, and fixes two latent traps found along the way.

## What

**New: `.agents/skills/improve-skill-quality`** — a triage playbook for an eval that regressed, returned no verdict, or failed to activate. Its core claim is the one lesson the history repeats loudest: **classify before you rewrite.** Broken fixtures, underpowered trial counts, forced tools, stale spec keys and harness errors have all masqueraded as skill regressions. Two reference docs carry the content patterns (`writing-for-baseline-delta.md`) and the symptom → cause → fix catalogue with PR citations (`eval-triage.md`).

**Rewritten: `.agents/skills/create-skill-test`** — it documented a **pre-Vally schema** (`scenarios:` / `assertions:` / `setup.copy_test_files`) that no longer loads. All 97 eval specs in `tests/` use `stimuli:` / `graders:` / `environment`. Anyone following the skill today authored a broken eval. The rewrite also folds in the trial floor, the dormancy-guard rule, and the fixture rules (tracked by git, buildable, self-consistent, actually reproduces its bug).

**Extended: `.agents/skills/create-skill`** — description-as-router rules (trigger words, sibling partitioning, exclusion audit, menu budget) and a "write for delta over the baseline model" section.

**`CONTRIBUTING.md`** — a "what consistently separates a passing skill from a failing one" playbook, eval sizing with the sign-test arithmetic, a corrected eval snippet, and a fix to the stale *"raise `runs` in `dotnet-skills.experiment.yaml`"* advice — an experiment-level `runs` **overwrites** every eval's own value rather than defaulting it.

**`.gitignore`** — `.agents/skills/` was silently swallowing the repo's own authoring skills; the new one only landed because I checked. Replaced with `.agents/skills/*` plus negations for the five owned directories, so installed third-party skills stay ignored. (This is the same defect class as the `.gitignore`-swallowed coverage fixture in #945.)

## Validation

- `markdownlint-cli2` — 0 errors
- `python eng/eval-quality/check_eval_quality.py` — no errors
- All relative links resolve; frontmatter parses; descriptions under 1,024 chars; bodies under 500 lines
- A separate review pass fact-checked every schema, arithmetic and command claim against the repo. It found five inaccuracies (a numbering gap, the stale `runs` advice, an over-absolute `disable-model-invocation` rule that `filter-syntax` contradicts, a wrong "8 trials tolerates a loss" row, and `run-command`'s required config) — all fixed before this was pushed.

No files under `plugins/` change, so no evaluation is triggered.

### DIFF
--- a/.agents/skills/create-skill/SKILL.md
+++ b/.agents/skills/create-skill/SKILL.md
@@ -91,7 +91,7 @@ Do not over-correct: a skilled answer shorter and less actionable than the basel
 ### Step 4: Create the skill directory
 
 ```
-skills/<skill-name>/
+plugins/<plugin>/skills/<skill-name>/
 └── SKILL.md
 ```
 
@@ -114,7 +114,7 @@ Include these recommended sections:
 ### Step 7: Add optional directories (if needed)
 
 ```
-skills/<skill-name>/
+plugins/<plugin>/skills/<skill-name>/
 ├── SKILL.md
 ├── scripts/       # Executable code agents can run
 ├── references/    # Additional documentation loaded on demand
@@ -144,7 +144,11 @@ Match the owner pattern used by sibling skills in the same plugin.
 
 A skill without an `eval.yaml` has no evidence that it improves on the baseline. Use
 `create-skill-test` to add one in the same pull request, and size it for statistical power — an eval
-below five trials can never return a passing verdict.
+below five counted trials can never return a passing verdict.
+
+The exception is a helper skill with `disable-model-invocation: true`: the model cannot
+self-activate it, so an activation-graded eval compares two identical arms. Cover it through the
+evals of the skills that load it instead.
 
 ## SKILL.md Template
 
@@ -214,7 +218,7 @@ After creating a skill, verify:
 - [ ] The description names concrete triggers and excludes the nearest sibling skills
 - [ ] Every section changes a decision the unskilled model would otherwise get wrong
 - [ ] The skill states when **not** to act, and what a truthful failure report looks like
-- [ ] An `eval.yaml` exists and clears the trial floor
+- [ ] An `eval.yaml` exists and clears the trial floor (or the skill is `disable-model-invocation: true` and covered through its consumers)
 
 ## Common Pitfalls
 
--- a/.agents/skills/create-skill/SKILL.md
+++ b/.agents/skills/create-skill/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: create-skill
-description: Scaffolds new agent skills for the dotnet/skills repository. Use when creating a new skill, generating SKILL.md files, or setting up skill directory structures. Handles frontmatter generation, section templates, and validation guidance.
+description: Scaffolds new agent skills for the dotnet/skills repository. Use when creating a new skill, generating SKILL.md files, writing a skill description that the runtime will actually route to, or setting up skill directory structures. Handles frontmatter generation, section templates, and validation guidance. Do not use for fixing a skill that already fails its evaluation (use improve-skill-quality) or for writing eval.yaml (use create-skill-test).
 ---
 
 # Create Skill
@@ -17,6 +17,8 @@ This skill helps you scaffold new agent skills that conform to the Agent Skills 
 ## When Not to Use
 
 - Modifying existing skills (edit directly instead)
+- Diagnosing or fixing a skill that fails its evaluation (use `improve-skill-quality`)
+- Writing the skill's `eval.yaml` (use `create-skill-test`)
 - Creating custom agents (use the agents/ directory pattern)
 
 ## Inputs
@@ -38,25 +40,66 @@ Ensure the name:
 - Does not contain consecutive hyphens
 - Is between 1-64 characters
 
-### Step 2: Create the skill directory
+### Step 2: Write the description — it is the router
+
+The `description` is the **only** text the runtime sees when deciding whether to load the skill.
+A perfect body behind a weak description never runs.
+
+```yaml
+---
+name: <skill-name>
+description: <what it does>. USE FOR: <symptoms, error codes, artifact names, quoted user requests>. DO NOT USE FOR: <nearby-but-wrong intents, with the skill that owns them>.
+---
+```
+
+- Lead with an action verb and use the user's own words: symptoms, error codes (`CS1501`,
+  `MSTEST0014`), artifact names (`.testsettings`, `binlog`), and requests phrased as a developer
+  would type them.
+- Partition against sibling skills on the **real discriminator**, not the topic. "Does the
+  abstraction already exist?" separates two skills; "testing" does not. Add the matching exclusion
+  to **both** siblings.
+- Claim the ambiguous words that would otherwise route to a sibling. If prompts say "review my
+  tests" and a sibling owns "review", say so explicitly.
+- Check every `DO NOT USE FOR` clause against the scenarios the skill exists to serve — an
+  exclusion like "already on v3" can lock out the post-upgrade fixes that are the skill's purpose.
+- Budget: 1,024 characters per description, and the whole plugin's rendered skill menu is also
+  budgeted. A helper skill users should never invoke directly can set
+  `disable-model-invocation: true` to free menu space while staying invocable by name.
+
+### Step 3: Write for delta over the baseline model
+
+Every skill is scored head-to-head against the same model with **no skill loaded**. Content the
+model already produces unaided is worth zero; content that makes it slower or more hedged is worth
+less than zero. See
+[improve-skill-quality/references/writing-for-baseline-delta.md](../improve-skill-quality/references/writing-for-baseline-delta.md)
+for the full evidence.
+
+| Do | Instead of |
+|----|------------|
+| Encode the decision the model would otherwise get wrong | Restating API signatures it already reproduces |
+| "When A, do B, never C, verify D" tables | Lists of plausible alternatives |
+| A concrete output contract (exact command, verdict line, findings table) | "Consider…", "you may want to…" |
+| Scale output structure to input size | A 12-section dashboard for an 8-test suite |
+| Stop-conditions that prevent over-applying | Acting before measuring, rewriting working code |
+| Instructing the agent to discover repo paths | Marking discoverable paths as required inputs |
+| Reporting restore/build/test failures truthfully | Claiming success after a failed command |
+| Verifying load-bearing API claims by compiling or probing | Trusting a source read |
+| Gating rare or expensive paths behind `references/` | One large SKILL.md carrying every path |
+
+Do not over-correct: a skilled answer shorter and less actionable than the baseline's still loses.
+
+### Step 4: Create the skill directory
 
 ```
 skills/<skill-name>/
 └── SKILL.md
 ```
 
-### Step 3: Generate SKILL.md with frontmatter
+### Step 5: Generate SKILL.md with frontmatter
 
-Create the file with required YAML frontmatter:
+Create the file with the frontmatter drafted in Step 2.
 
-```yaml
----
-name: <skill-name>
-description: <description of what the skill does and when to use it>
----
-```
-
-### Step 4: Add body content sections
+### Step 6: Add body content sections
 
 Include these recommended sections:
 
@@ -68,7 +111,7 @@ Include these recommended sections:
 6. **Validation**: How to confirm the skill worked correctly
 7. **Common Pitfalls**: Known traps and how to avoid them
 
-### Step 5: Add optional directories (if needed)
+### Step 7: Add optional directories (if needed)
 
 ```
 skills/<skill-name>/
@@ -78,7 +121,7 @@ skills/<skill-name>/
 └── assets/        # Templates, images, data files
 ```
 
-### Step 6: Update CODEOWNERS
+### Step 8: Update CODEOWNERS
 
 Add entries in `.github/CODEOWNERS` for the new skill and its test directory:
 
@@ -89,12 +132,19 @@ Add entries in `.github/CODEOWNERS` for the new skill and its test directory:
 
 Match the owner pattern used by sibling skills in the same plugin.
 
-### Step 7: Validate the skill
+### Step 9: Validate the skill
 
 - Confirm frontmatter fields are valid
 - Ensure SKILL.md is under 500 lines
 - Check that file references use relative paths
 - Verify instructions are actionable and specific
+- Run `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin ./plugins/<plugin>`
+
+### Step 10: Add the eval
+
+A skill without an `eval.yaml` has no evidence that it improves on the baseline. Use
+`create-skill-test` to add one in the same pull request, and size it for statistical power — an eval
+below five trials can never return a passing verdict.
 
 ## SKILL.md Template
 
@@ -161,6 +211,10 @@ After creating a skill, verify:
 - [ ] Validation section exists with observable success criteria
 - [ ] No secrets, tokens, or internal URLs included
 - [ ] `.github/CODEOWNERS` has entries for the new skill and its test directory
+- [ ] The description names concrete triggers and excludes the nearest sibling skills
+- [ ] Every section changes a decision the unskilled model would otherwise get wrong
+- [ ] The skill states when **not** to act, and what a truthful failure report looks like
+- [ ] An `eval.yaml` exists and clears the trial floor
 
 ## Common Pitfalls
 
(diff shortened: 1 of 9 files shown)


.agents/skills/improve-skill-quality/SKILL.md
```diff
@@ -0,0 +1,187 @@
...
+- coverage fixtures are self-consistent: declared `line-rate`, summary totals, and the `<line>`
+  elements must all report the same number, or the two arms legitimately read different truths.
+
+### Step 5: Check whether the eval could ever have passed
+
+The gate is an exact one-sided sign test over **discordant** (non-tie) trials.
+
+- Below 5 trials no record can pass, however good the skill.
+- At 5–7 trials only a clean sweep passes; one tie makes a pass arithmetically unreachable.
+- Tolerating a single loss needs 8 discordant trials.
```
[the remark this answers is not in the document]
Evangelink: Good catch, and it was worse than stated — the docs were conflating the gate's two independent bars. `adapt.mjs` gates on (1) `underpowered = directions.length < MIN_CREDIBLE_TRIALS`, which counts **total** trials, and (2) `signTestPValue(wins, losses)` over the **discordant** trials. So at 6 trials `5W/1T/0L` is 5 discordant and passes at p = 0.031, and at 7 trials two ties are survivable.

Fixed in cdba3aa. Every trial table is now keyed on discordant trials rather than counted trials:

| discordant trials | records that pass | p |
|---:|---|---:|
| <= 4 | none | >= 0.0625 |
| 5-7 | zero losses only (5W/0L) | 0.031 |
| 8 | one loss survivable (7W/1L) | 0.035 |

with the counted-trial floor stated separately. I also corrected `eng/eval-quality/README.md`, which carried the same over-generalisation ("at 5, 6 or 7 trials the only record that passes is every trial a win") — these docs restate it, so leaving it would have put the two in conflict. Flagging that explicitly since it is outside the original scope of this PR.

.agents/skills/create-skill/SKILL.md
```diff
@@ -38,25 +40,66 @@ Ensure the name:
...
+| Reporting restore/build/test failures truthfully | Claiming success after a failed command |
+| Verifying load-bearing API claims by compiling or probing | Trusting a source read |
+| Gating rare or expensive paths behind `references/` | One large SKILL.md carrying every path |
+
+Do not over-correct: a skilled answer shorter and less actionable than the baseline's still loses.
+
+### Step 4: Create the skill directory
+
+```
+skills/<skill-name>/
```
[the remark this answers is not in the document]
Evangelink: Agreed — there is no top-level `skills/` directory in this repo, and the same file's Step 8 (CODEOWNERS) and Step 10 (eval) already use the real `plugins/<plugin>/skills/<skill-name>/` layout, so the scaffolding examples contradicted them.

Fixed in cdba3aa: both the Step 4 and Step 7 trees now read `plugins/<plugin>/skills/<skill-name>/`, matching `CONTRIBUTING.md`'s repository layout section.

Evangelink:
## Review rounds

Ran three independent review passes over the diff, each pointed at the implementation rather than the prose: **Copilot**, **Claude Opus 4.8**, and **GPT-5.6** (Sol + Terra). All findings verified against `eng/vally-adapter/adapt.mjs`, `eng/eval-quality/check_eval_quality.py`, `.github/workflows/evaluation.yml`, and the 97 real specs under `tests/`. Everything below is in cdba3aa.

### Accepted (10)

**The statistics were wrong, and both Copilot and Opus caught it independently.** The docs conflated the gate's two independent bars. `adapt.mjs` gates on `underpowered = directions.length < 5` (**counted** trials) *and* `signTestPValue(wins, losses) <= 0.05` (**discordant** trials). So "at 5-7 trials one tie makes a pass unreachable" is false: `5W/1T/0L` over 6 trials is 5 discordant and passes at p = 0.031. Opus then caught the residual error in my own fix — at 6 trials *two* ties leave 4 discordant and fail, so "one or two ties are survivable" only holds at 7. Every table is now keyed on discordant trials.

| # | Finding | Source |
|---|---------|--------|
| 1 | Trial arithmetic conflated counted vs discordant trials | Copilot, Opus |
| 2 | "one or two ties survivable" wrong at 6 trials | Opus |
| 3 | `filter-syntax` described as the sole answer-graded eval — `platform-detection` uses the same pattern | Opus |
| 4 | "every skill needs an eval" contradicted the `disable-model-invocation` exception | GPT |
| 5 | Minimal eval example declares 1 trial and would be rejected by the gate | GPT |
| 6 | "every buildable fixture must build" forbade the intentionally broken fixtures the same doc demonstrates | GPT, Terra |
| 7 | "do not proceed until you can quote a losing trial" blocked the no-results and underpowered paths this skill exists to handle | GPT |
| 8 | `expect_tools` presented as an activation assertion — it checks tool usage; the harness reports activation separately | GPT |
| 9 | `CONTRIBUTING` claimed evaluations run automatically on plugin changes; a maintainer must trigger `/evaluate` bound to a commit | GPT |
| 10 | First-match triage table routed fixture failures to the harness row, and treated any positive-but-failing record as a power problem although `adapt.mjs` requires `conclusive` first | Terra |

### Also corrected outside the original scope

`eng/eval-quality/README.md` carried the same over-generalisation this PR was reproducing ("at 5, 6 or 7 trials the only record that passes is every trial a win"). Since the new docs restate that section, leaving it would have put the two in conflict — so the blockquote under check 8 and the "Evals parked at the floor" paragraph are corrected. Happy to split that out if the owners prefer.

### Rejected (2)

- **`eng/version/nuget.config` is invalid XML** (Terra, High). Real bug, but not this PR: it shows up only under a two-dot `git diff origin/main`, which includes upstream drift. The three-dot PR diff does not touch the file, and `main` has **already fixed** it — this branch simply predates that commit. Nothing to do here.
- **Local-run prerequisites omit Bash** (GPT). Correct that `eng/run-skill-evals.sh` needs Bash and that the prerequisites list only mentions Node and `gh`. Pre-existing, unrelated to any line this PR touches, and fixing it properly means either a cross-platform wrapper or a documented WSL/Git Bash requirement — a separate change rather than something to bury in a docs PR.

### Validation

`markdownlint-cli2` 0 errors - `check_eval_quality.py` no errors - `adapt.test.mjs` 21/21 pass - all relative links and the new in-page anchor resolve - step numbering sequential in both rewritten skills.

.agents/skills/improve-skill-quality/references/eval-triage.md
```diff
@@ -0,0 +1,93 @@
...
+| One arm has unmatched trajectories | Comparison judge error or arm timeout biases the remaining trials | Report inconclusive, do not read it as a regression | PR #887 |
+| Fixture resolves locally, SDK-not-found in CI | Fixture pin without roll-forward | `global.json` with `rollForward: latestMajor` | PR #907 |
+
+## Fixtures
+
+| Symptom | Cause | Fix | Evidence |
+|---------|-------|-----|----------|
+| Judge penalizes the agent for "pre-existing build issues" | A fixture meant to be healthy does not compile | Build every healthy fixture before shipping the eval; a deliberately broken one must fail only for the reason its stimulus is about | PR #949 |
+| Scenario fails at setup in CI but passes locally | Fixture is on disk but not in the git index (`.gitignore` swallowed it) | Verify with `git ls-files`; the gate now blocks this | PR #945, PR #953 |
+| Judge says the response "made a critical error" about a number | Cobertura fixture is split-brain: declared `line-rate` disagrees with its `<line>` payload or summary totals | Make declared rate, summary totals and payload agree, then re-derive any rubric quoting a figure | PR #964, PR #945 |
```
AbhitejJohn: This seems rather specific to the test skills - at least for the error.  Can we filter these out? Maybe we add them to a separate section (`examples` etc) if we still want to keep them.

AbhitejJohn:
## Review — PR #979

Thanks for putting this together, Amaury — this is a thorough extraction of lessons that were previously scattered across PR bodies and a few people's heads.

### What I verified

| Claim | Result |
|-------|--------|
| Sign-test arithmetic (p = 0.031, 0.035, ≥ 0.0625) | **Confirmed.** Computed: `0.5^5 = 0.03125`, `(C(8,7)+C(8,8))/256 = 0.03516`, `0.5^4 = 0.0625` |
| `.gitignore` swallowed repo-owned skills | **Confirmed.** Old pattern was bare `.agents/skills/`; new pattern `.agents/skills/*` + 5 negations matches exactly the 5 directories in the repo |
| `check_eval_quality.py` passes | **Confirmed.** 0 errors, 97 specs checked |
| `selftest_eval_quality.py` passes | **Confirmed.** All 22 self-tests pass |
| All relative links resolve | **Confirmed.** All 7 cross-referenced paths exist on the PR branch |
| Description lengths < 1024 chars | **Confirmed.** 456, 447, 476 chars |
| Body lines < 500 | **Confirmed.** 319, 246, 208 lines |
| Old CONTRIBUTING.md eval snippet used `config:` | **Confirmed.** Line 253 was `config:` — following the old "add `defaults: runs:`" advice would silently break the spec |
| Old "raise `runs` in experiment yaml" advice was wrong | **Confirmed.** Experiment yaml comments out `runs` and explains it would overwrite every eval's own value |

### Findings

**1. Agent eval path guidance gap** (worth addressing)
The rewritten `create-skill-test/SKILL.md` says "for a skill or agent" and mentions the `agent.` prefix convention, but agent evals are explicitly excluded from the canonical experiment (`dotnet-skills.experiment.yaml` `evals:` glob skips `agent.*`), and `run-skill-evals.sh` can't pull them in. The old doc had some agent-specific evaluation coverage. A brief note in Step 10 or Common Pitfalls saying that `agent.*` evals are excluded from the sign-test flow (and explaining how to validate them instead) would help agent-eval authors.

**2. Pre-existing terminology drift in `eng/eval-quality/README.md`** (nit, not introduced by this PR)
Line 154 still says `trials = scenarios × defaults.runs` while the new authoring skills consistently use `stimuli × runs`. Worth aligning in a follow-up.

**3. `.gitignore` maintenance note** (nit)
The negation list is exactly correct today (5 dirs, 5 negations). A one-line comment noting that new repo-owned authoring skills need a negation added here would help the next person.

**4. `InvestigatingResults.md` reference** (nit)
`improve-skill-quality/SKILL.md` Step 1 links to `eng/vally-adapter/InvestigatingResults.md`, while `AGENTS.md` references `eng/skill-validator/src/docs/InvestigatingResults.md`. If these are different docs for different audiences, a brief note clarifying which is which would help.

### Summary

This is excellent work. The `create-skill-test` rewrite fixes genuinely dangerous documentation (the old pre-Vally schema would produce broken evals), the `.gitignore` fix closes a real bug, the CONTRIBUTING.md corrections prevent silent spec-load failures, and the new `improve-skill-quality` triage playbook is the kind of guidance that prevents the most common misdiagnosis in this repo. The only actionable finding is the agent-eval gap (#1).

.agents/skills/improve-skill-quality/references/eval-triage.md
```diff
@@ -0,0 +1,93 @@
...
+| A dormancy guard scores randomly across runs | `expect_activation: false` combined with `constraints.reject_skills`, making the skilled arm skill-free and identical to baseline | Use `expect_activation: false` alone | PR #945, PR #953 |
+| A reference skill shows no improvement | `disable-model-invocation: true` means the model cannot self-activate it, so an activation-graded eval compares identical arms | Cover it through a consumer skill, or grade answer content as `filter-syntax` does | PR #971, PR #976, issue #899 |
+| An eval "passes" while the skill stopped emitting its signature output | No grader asserts the mandated shape | Add a grader for the exact contract (e.g. the `Recommendation:` line) | PR #904 |
+| Overfit score high, user value unclear | Rubric items reward using the skill, or prompts echo skill vocabulary | Drop them: the harness already reports activation separately, so a rubric never needs to. Keep rubric items outcome-shaped and de-cue the prompt | PR #904 |
+| Both arms write tests and the judge compares volume | Non-activation rubric rewards raw output | Add anti-hijack criteria: do not invoke the skill, do not reward test count | PR #945 |
```
AbhitejJohn: This one might have some test specific language as well.

AbhitejJohn (approved):
Thanks for putting this together.

Evangelink:
Thanks @AbhitejJohn — all four are addressed in c8b60e8. No follow-up tickets needed; each turned out to be a few lines in files this PR already touches.

### 1. Agent eval gap — fixed, and it was worse than a doc gap

You were right that this needed addressing. Confirmed the mechanism:

```yaml
# dotnet-skills.experiment.yaml
evals:
  - tests/*/!(agent.*)/eval.yaml
```

and `eng/run-skill-evals.sh:20-22` intersects `--eval-filter` with that glob, so **naming an agent eval explicitly still drops it silently** — the runner exits successfully having run nothing. So the doc was not merely incomplete: it told authors to size an `agent.*` eval for a trial floor that never applies and to validate it with a command that is a no-op.

Three changes:
- **Step 1** states that `agent.*` sits outside the verdict flow — no verdict, no floor, dropped by the runner even when named.
- **Step 10** gives the route that does work: widen the `evals:` glob in a copied experiment file and pass it via `EXPERIMENT_FILE`, then read the trajectories rather than a verdict.
- Checklist and pitfalls scoped: `stimuli × runs clears 5` now says "for a skill eval", plus rows for sizing an agent eval for the floor and for expecting `run-skill-evals.sh` to run one.

### 2. `scenarios` vs `stimuli` drift — fixed here

Aligned check 8 in `eng/eval-quality/README.md` (`trials = stimuli × defaults.runs`, and the sentence above it). It is a two-word change in a file this PR already edits, so carrying it as a follow-up seemed like more overhead than the fix.

### 3. `.gitignore` maintenance note — added

```gitignore
# Adding a new repo-owned authoring skill? Add a negation for it here too, or it
# will be silently untracked.
```

### 4. `InvestigatingResults.md` — they are not parallel docs

Worth recording, because the answer is cleaner than "two audiences": `eng/skill-validator/src/docs/InvestigatingResults.md` opens with a banner saying the LLM eval pipeline no longer uses `skill-validator evaluate`, that it documents the **legacy** schema, and that readers should use `eng/vally-adapter/InvestigatingResults.md` instead. I also confirmed `skill-validator` no longer ships an `evaluate` verb at all.

So the triage skill links the correct one. `AGENTS.md`'s reference is also correct in its own context — that repo instruction is about keeping the validator's own bundled doc in sync when the validator changes. I added a line to the `improve-skill-quality` references section naming the legacy doc explicitly, so nobody reads the retired schema by mistake.

### Validation

markdownlint 0 errors - `check_eval_quality.py` no errors - `selftest_eval_quality.py` 22/22 - `.gitignore` negations re-verified with `git check-ignore` (repo-owned skills tracked, a stranger skill still ignored).

.agents/skills/improve-skill-quality/references/eval-triage.md
Evangelink: Agreed — that row stated its rule entirely in Cobertura terms, so an author on `dotnet-maui` or `dotnet-msbuild` would read past it even though the failure mode is not coverage-specific at all.

Generalized in 85a8c84 rather than moved to an `examples` section, so the rule stays in the table where the symptom is looked up and the concrete case survives as the illustration:

> Judge says the response "made a critical error" about **a value the fixture supplies** | The fixture states the same fact in two places and they disagree, so each arm can legitimately read a different truth — e.g. a Cobertura report whose declared `line-rate` contradicts its `<line>` payload or summary totals | Make every representation of the value agree, then re-derive any rubric or prompt that quotes it

I also generalized the one-line summary of this same rule in `SKILL.md` Step 4, so the detail and the summary do not disagree with each other.

I checked the rest of the file for the same problem. The other concrete names (`writing-mstest-tests`, `filter-syntax`, the `dotnet-test` run in the power section) all sit in the evidence/example position behind a rule that is already stated generally, so I left them — that is the shape you described wanting to keep.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

.agents/skills/improve-skill-quality/references/eval-triage.md
Evangelink: Right — that row was the other one written from inside the test plugin ("both arms write tests", "do not reward test count"), even though the failure mode is generic: whenever both arms emit the same *kind* of artifact, the judge has nothing else to separate them and falls back on counting.

Generalized in 85a8c84:

> Both arms produce **the same kind of artifact** and the judge falls back on comparing volume | The rubric rewards raw output instead of the property under test | Add anti-hijack criteria: do not invoke the skill, and do not reward quantity (number of tests, findings, or lines produced)

The concrete units are kept as the parenthetical so the guidance is still actionable, but a `dotnet-msbuild` or `dotnet-maui` author now sees a rule that applies to them.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

Evangelink:
Acting on two suppressed findings from the Copilot review summaries — flagging here because one of them changed documented guidance after your approval, @AbhitejJohn.

**`environment.skills` was documented wrongly.** The Step 5 skeleton called it "extra skills loaded in the isolated arm only", and both halves of that are false:

```yaml
# dotnet-skills.experiment.yaml
vary:
  - /environment/skills
variants:
  baseline: { environment: { skills: [] } }
  skilled:  { environment: { skills: [plugins/${eval.grandparent}/skills/${eval.parent}] } }
```

The experiment supplies that key per arm, so anything a **skill** eval declares is replaced in every arm. It cannot add a skill to one arm, and in a skill eval it does nothing at all. It is meaningful only in `agent.*` evals, which the experiment does not vary. Removed from the generic skeleton, documented where it does apply, plus a pitfall row.

While checking, I noticed the three agent specs in this repo disagree on how those entries are spelled — `tests/dotnet-msbuild/agent.msbuild/eval.yaml` uses bare names (`binlog-failure-analysis`) while `agent.test-quality-auditor` and `agent.test-migration` use relative paths (`../../plugins/dotnet-test/skills/...`). Rather than assert a resolution rule I could not verify from the repo, the doc now tells authors to copy an existing agent eval. **Worth a look separately** — one of those two forms is likely not loading what its author intended.

**Stale cross-reference:** moving the fixture row to the top of the triage table last round left the sentence beneath it saying a fixture failure "also matches the two rows above it". Reworded.

Both in 13e67d16. markdownlint and `check_eval_quality.py` clean.

YuliiaKovalova (approved):

PR #993, opened 2026-08-05
Support classic .NET test projects in dotnet-test

Evangelink:
## Summary

- detect classic non-SDK projects and `packages.config` before choosing build, test, or coverage commands
- preserve explicit `<Compile Include>` registration, legacy dependency versions, custom fixture bases, and repository MSBuild/VSTest/MSTest workflows
- gate MSTest authoring advice by the installed version, including MSTest 3.5.x fallbacks
- prevent SDK-style coverage package injection into classic projects and safely partition mixed SDK/classic solutions
- require explicit, repository-policy-compliant opt-in before testability migrations
- add regression evals and realistic net472 fixtures for generation, execution, coverage, and MSTest authoring

## Review

Three independent review rounds were completed. Findings fixed included unsafe `packages.config` classification, mixed-solution coverage execution, unconditional testability migration routing, legacy validation wording, MSTest version-gating contradictions, and fixture adapter integration.

## Evaluation follow-up

Run `31012795023` produced three non-passes with distinct causes:

- **code-testing-agent — routing/content:** 5W/6T/1L (`p=0.109`), with activation only 1/5. Its frontmatter simultaneously claimed mandatory test generation and excluded MSTest. The only loss said the baseline was better because it covered the exact `0%` boundary while the skilled run substituted `50%`. The description now keeps MSTest generation at this entry point, and the classic stimulus explicitly requires `<0`, `0`, `100`, and `>100`.
- **crap-score — statistical power:** 4W/2T/0L (`p=0.063`). Two ties left only four discordant trials, so no record could have passed. Two distinct scenarios now measure classic-project coverage safety and async `MoveNext`-to-source line mapping rather than repeating existing tasks.
- **platform-detection — eval design/activation:** 0/5 activation because the skill set `disable-model-invocation: true`; the isolated arm randomly lost the SDK 10 precedence case while the plugin arm answered it correctly. Platform detection is now directly invocable and its expanded eval measures harder project-system, runner, bridge, and precedence decisions. The guidance also now distinguishes `dotnet test` command mode from the platform actually executed through the MTP bridge.

Run `31039546010` was inconclusive because three trials errored at `session.idle`:

- Two `writing-mstest-tests` plugin trials stalled at the exact 5-minute budget—one before any response and one after two successful globs. This was classified as transient reliability; the eval now allows 6 minutes without changing skill content.
- One isolated classic coverage trial ignored the existing stop instruction, created a temporary `.coverage-run` SDK project, collected coverage from the substitute assembly, and timed out while writing that misleading report. The skill now has a prominent hard stop forbidding temporary wrappers or substitute coverage. The eval snapshots every project file and rejects any newly generated Cobertura report.
- To restore runtime headroom, `coverage-analysis` now uses two runs instead of three. Its eight distinct scenarios still produce 16 counted trials; the prior measured record was 17W/6T/1L over 24.
- Review also corrected the legacy modernization fixture from MSTest 3.6.3 to 3.8.0 and scoped every `crap-score` collection fallback to SDK-style projects.

Run `31083033945` exhausted the fixed 155-minute Vally step after 233 of 234 expected trajectories; all recorded trajectories succeeded and no verdict phase ran. Runtime analysis showed the largest costs in broad generation/build scenarios. The evals now retain every unique requirement while replacing expensive near-ties with focused no-tool stimuli:

- coverage still checks branch gaps and multi-member 80% arithmetic; the retained measured subset was 16W/1T/1L
- run-tests still covers classic/VSTest, bridged/native MTP, multi-TFM, NUnit, xUnit, TUnit, TRX, combined filters, diagnostics, and props detection; the unchanged measured subset was 10W/1T
- writing-mstest now tests API review/modernization rather than generation owned by `code-testing-agent`, including cooperative timeout syntax and correct `IsInRange(min, max, value)` ordering; the unchanged measured subset was 11W/1T
- expected workload is 219 trajectories, with the expensive restore/build/generation paths converted to inline read-only tasks and unreferenced fixtures removed

## Validation

- `dotnet run --no-build --no-restore --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin ./plugins/dotnet-test`
- XML parsing and internal Cobertura consistency checks for all added fixtures
- exercised the classic coverage safety grader under Git Bash
- `git diff --check`

Run `31112078844` confirmed the structural bottleneck: all six changed skills were still assigned to the synthetic `default` execution shard, so one Vally job owned every baseline/isolated/plugin trajectory. The workflow already supports `executionShard:` and merges shard artifacts downstream. The six evals are now split into four domain shards aligned with `max-parallel: 4`:

- `generation`: code-testing-agent
- `risk`: coverage-analysis + crap-score
- `execution`: platform-detection + run-tests
- `mstest`: writing-mstest-tests

This changes only scheduling; each skill retains the same scenarios, variants, comparison, and statistical gate. The obsolete monolithic run bound to `8c638d083` was cancelled after the sharded commit was pushed.

Run `31112078844` completed comparisons and narrowed the remaining work to two skills:

- **code-testing-agent — eval discrimination:** 2W/10T/0L (`p=0.250`). Only the broad Vitest scenario activated the skill consistently; focused generation was baseline-equivalent. Three distinct project-wide scenarios now exercise the orchestration/checklist advantage across multi-module Python, classic multi-class MSTest, and multi-package Go. At two runs each, the eval now has 18 trials and six new broad trial opportunities without relying on extra repetition.
- **platform-detection — output contract + boundary coverage:** 3W/5T/1L (`p=0.312`). The sole loss was semantically correct—“VSTest command mode executing MTP”—but put VSTest on the requested `Platform:` line. The skill now defines `Platform` as the executed platform and reports command mode separately. Two baseline-easy framework cases were replaced by harder, valid SDK 9 boundaries: MTP runner without the dotnet-test bridge, and MSTest.Sdk explicitly opting back into VSTest with `UseVSTest=true`.

Both replacement platform fixtures restore successfully, and the generation fixtures compile/build. The generation shard now evaluates 18 trials; platform detection remains at 9 trials.

Run `31117103578`, the first sharded retry, never reached Vally: `prepare-validator` failed while downloading a GitHub Action with repeated `Service Unavailable` errors. This was transient infrastructure, not an eval result.


Evangelink (commented):
/evaluate

YuliiaKovalova (approved):

tests/dotnet-test/writing-mstest-tests/eval.yaml
```diff
@@ -204,6 +204,37 @@ stimuli:
...
+          return new Customer(input.Name);
+      }
+      ```
+    graders:
+      - type: output-matches
+        config:
+          pattern: Assert\.ThrowsException<ArgumentNullException>
+      - type: output-matches
+        config:
+          pattern: (3\.5\.2|installed|version|compatible)
```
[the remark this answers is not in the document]
Evangelink: Good catch — tightened the grader in a48fd80a to require an explicit MSTest 3.5.2 mention.

tests/dotnet-test/run-tests/fixtures/classic-mstest/OrderServiceTests.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    [TestClass]
+    public class OrderServiceTests
+    {
+        [TestMethod]
+        public void CalculateTotal_EmptyOrder_ReturnsZero()
+        {
+            Assert.AreEqual(0m, 0m);
+        }
+    }
```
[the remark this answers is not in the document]
Evangelink: Fixed in a48fd80a. The fixture now computes and asserts a two-line-item total instead of comparing constants.

tests/dotnet-test/code-testing-agent/fixtures/classic-mstest/src/DiscountService.cs
```diff
@@ -0,0 +1,42 @@
...
+        public DiscountService(IProductRepository repository)
+        {
+            _repository = repository;
+        }
+
+        public decimal Apply(int productId, decimal percentage)
+        {
+            if (percentage < 0m || percentage > 100m)
+            {
+                throw new ArgumentOutOfRangeException("percentage");
```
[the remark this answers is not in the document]
Evangelink: Fixed in a48fd80a by using nameof(percentage) for the ArgumentOutOfRangeException parameter name.

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

tests/dotnet-test/code-testing-agent/fixtures/classic-mstest/tests/DiscountServiceTests.cs
```diff
@@ -0,0 +1,32 @@
+using FizzWare.NBuilder;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 2718e640 by importing Contoso.Discounts so the classic fixture resolves the production types.

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

tests/dotnet-test/writing-mstest-tests/eval.yaml
```diff
@@ -343,153 +335,66 @@ stimuli:
...
         config:
-          pattern: Assert\.(IsGreaterThan|IsLessThan|IsInRange|IsGreaterThanOrEqualTo|IsLessThanOrEqualTo)
+          pattern: Assert\.IsGreaterThan
       - type: output-matches
         config:
-          pattern: \bsealed\b
+          pattern: Assert\.IsLessThan
+      - type: output-matches
+        config:
+          pattern: Assert\.IsInRange\(60,\s*90,\s*score\)
```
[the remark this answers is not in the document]
Evangelink: No change needed here: MSTest 3.10 defines `Assert.IsInRange(minValue, maxValue, value)`, so `Assert.IsInRange(60, 90, score)` is the correct order. The value-first mapping you found is in a separate xUnit-migration reference and is stale; changing this grader would reward invalid MSTest syntax.

tests/dotnet-test/platform-detection/eval.yaml
```diff
@@ -21,15 +166,18 @@ stimuli:
     graders:
       - type: output-matches
         config:
-          pattern: (?im)^\s*\**\s*Platform:\s*\**\s*VSTest
+          pattern: (?im)^\s*\**\s*dotnet test mode:\s*\**\s*VSTest
+      - type: output-matches
```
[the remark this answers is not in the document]
Evangelink: This refers to an older fixture state. On the current head, `sdk10-globaljson-vstest-wins/global.json` explicitly selects `VSTest`, and its project is MSTest with the MTP runner and bridge enabled. The expected result is therefore VSTest command mode, MTP execution platform, and MSTest framework.

tests/dotnet-test/platform-detection/eval.yaml
```diff
@@ -56,9 +204,9 @@ stimuli:
           pattern: (?im)^\s*\**\s*Framework:\s*\**\s*xUnit
       - type: prompt
     rubric:
-      - Concluded MTP from the global.json test.runner setting
+      - Concluded MTP because global.json selects native MTP mode and the project enables the xUnit MTP runner
       - Identified the framework as xUnit from the xunit.v3 package reference
-      - Did not require an MSBuild property such as TestingPlatformDotnetTestSupport to be present, since global.json settles it on .NET 10+
+      - Did not require TestingPlatformDotnetTestSupport because native MTP mode does not use the VSTest bridge
```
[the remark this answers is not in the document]
Evangelink: This scenario and fixture were replaced in `43f7d776`. The current eval no longer references `sdk10-globaljson-mtp`; it now covers the harder SDK 9 runner-without-bridge and `MSTest.Sdk` `UseVSTest=true` boundaries.

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

plugins/dotnet-test/skills/coverage-analysis/references/test-execution.md
```diff
@@ -94,22 +125,21 @@ must be passed after the `--` separator. In .NET 10+, `--coverage` is a top-leve
...
-        --coverage `
-        --coverage-output-format cobertura `
-        --coverage-output $rawDir
-} else {
-    # .NET 9: pass Microsoft.Testing.Platform arguments after the -- separator
-    dotnet test "<ENTRY>" `
-        --results-directory $rawDir `
-        -- --coverage --coverage-output-format cobertura --coverage-output $rawDir
+foreach ($entry in $coverageEntries) {
+    if ($major -ge 10) {
```
JanKrivanek: This branches only on the SDK major version, but .NET 10 can still run in VSTest mode. In that mode, `--project` and top-level `--coverage` are the wrong syntax. Detect the runner mode separately and keep the positional path plus arguments after `--` for VSTest-bridge execution.

tests/dotnet-test/platform-detection/fixtures/sdk10-native-mtp-use-vstest-conflict/TestProject.csproj
```diff
@@ -1,6 +1,6 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="MSTest.Sdk/3.8.0">
```
JanKrivanek: `MSTest.Sdk` 3.8.0 brings MTP 1.6.x, below the MTP 1.7+ required by native .NET 10 `dotnet test`. Removing `UseVSTest` therefore still leaves this fixture incompatible, while the rubric attributes the failure only to the runner-mode conflict. Use a compatible `MSTest.Sdk` version or require both problems in the diagnosis.

tests/dotnet-test/code-testing-agent/eval.yaml
```diff
@@ -1,125 +1,176 @@
...
         config:
-          command: sh -c "cd fixtures/typescript-vitest-cart && npx vitest run --coverage"
+          command: sh -c "diff -u .eval-baseline/DiscountServiceTests.cs fixtures/classic-mstest/tests/DiscountServiceTests.cs && diff -u .eval-baseline/packages.config fixtures/classic-mstest/tests/packages.config && test $(grep -c 'Compile Include=\"DiscountServiceBoundaryTests.cs\"' fixtures/classic-mstest/tests/Discounts.Tests.csproj) -eq 1 && test $(grep -c 'Compile Include=\"TieredDiscountPolicyTests.cs\"' fixtures/classic-mstest/tests/Discounts.Tests.csproj) -eq 1"
           expected_exit_code: 0
-          timeout: 10m
+          timeout: 1m
+      - type: file-not-contains
+        config:
+          path: fixtures/classic-mstest/tests/DiscountServiceBoundaryTests.cs
+          value: ThrowsExactly
```
JanKrivanek, line 118: This only excludes `ThrowsExactly`, but the fixture pins MSTest 3.5.2, where `Assert.Throws<T>` is unavailable too. An uncompilable generated suite can therefore pass this grader. Require `Assert.ThrowsException<T>` or add a compilation check.

plugins/dotnet-test/skills/coverage-analysis/references/test-execution.md
Evangelink: Fixed in a09eca068. Coverage collection now detects the effective `global.json` and branches on native MTP versus VSTest command mode; .NET 10 VSTest-bridge execution keeps the positional path and passes coverage arguments after `--`.

tests/dotnet-test/platform-detection/fixtures/sdk10-native-mtp-use-vstest-conflict/TestProject.csproj
Evangelink: Fixed in a09eca068. The fixture now uses MSTest.Sdk 4.1.0, so native .NET 10 MTP is compatible when `UseVSTest` is disabled and the evaluated incompatibility isolates the intended runner-mode conflict.

tests/dotnet-test/code-testing-agent/eval.yaml
Evangelink: Fixed in a09eca068. The classic MSTest 3.5.2 eval now requires `Assert.ThrowsException<T>` in both generated files and rejects both `Assert.Throws<T>` and `Assert.ThrowsExactly<T>`.

JanKrivanek (commented):
/evaluate

plugins/dotnet-test/skills/coverage-analysis/references/test-execution.md
JanKrivanek: Resolved in a09eca068 and covered by the focused regression added in dbc194a9a. Coverage guidance now reads global.json to distinguish native MTP from VSTest mode; SDK 10 VSTest mode keeps the positional path and passes MTP coverage arguments after `--`.

tests/dotnet-test/platform-detection/fixtures/sdk10-native-mtp-use-vstest-conflict/TestProject.csproj
JanKrivanek: Resolved in a09eca068 by upgrading this fixture to `MSTest.Sdk/4.1.0`. I also verified it with SDK 10: as written it fails on the native-MTP/`UseVSTest` conflict, while overriding `UseVSTest=false` proceeds through native MTP, so the fixture no longer has the unrelated MTP 1.6 compatibility problem.

tests/dotnet-test/code-testing-agent/eval.yaml
JanKrivanek: Resolved in a09eca068 and hardened further in dbc194a9a. Both generated files must use `Assert.ThrowsException<T>` and reject `Assert.Throws<T>`/`Assert.ThrowsExactly<T>`; the shell check is whitespace-tolerant. The pinned MSTest 3.5.2 assembly was also verified to expose `ThrowsException` but not `Throws`.

tests/dotnet-test/code-testing-agent/fixtures/go-multipackage/go.mod
```diff
@@ -0,0 +1,3 @@
+module example.com/checkout
+
+go 1.23
```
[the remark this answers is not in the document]
JanKrivanek: Resolved in 99b8b256d by changing the fixture directive from Go 1.23 to the repository baseline of Go 1.22. The fixture uses no Go 1.23-specific language or library features.

JanKrivanek (commented):
/evaluate

JanKrivanek (approved):
Good to go now

JanKrivanek (commented):
/evaluate

Evangelink (commented):
/evaluate

JanKrivanek (approved):

JanKrivanek (commented):
/evaluate

JanKrivanek (commented):
/evaluate

JanKrivanek (commented):
/evaluate

JanKrivanek (commented):
/evaluate

JanKrivanek (commented):
/evaluate

AbhitejJohn (approved):

AbhitejJohn:
@JanKrivanek, @Evangelink I re-checked the merged result (`5a06b20c`) after the final fixes. The earlier `Assert.IsInRange` and MSTest-generation routing issues are fixed. These items still remain:

1. **Coverlet fails in native MTP mode.** `plugins/dotnet-test/skills/coverage-analysis/references/test-execution.md:129-141` and `:186-193` always use a positional path plus `--collect:"XPlat Code Coverage"`, even when `$dotnetTestMode` is `native-MTP`. Coverlet's own VSTest integration docs state: **"`coverlet.collector` is not compatible with the Microsoft Testing Platform (MTP v2)"** and direct native MTP users to `coverlet.MTP` with `--coverlet`. Could this path stop with a clear compatibility error, switch to `coverlet.MTP`, or deliberately run in VSTest mode?

2. **`crap-score` still chooses coverage syntax by SDK major.** `plugins/dotnet-test/skills/crap-score/SKILL.md:76-77` gives top-level `--coverage` to every .NET 10+ project. This conflicts with the fixed guidance at `coverage-analysis/references/test-execution.md:146-170`, where SDK 10 VSTest mode keeps the positional path and passes MTP arguments after `--`. I also reconfirmed this from SDK `10.0.100-preview.6.25358.103` in VSTest mode:

```text
ABSENT:--project
ABSENT:--solution
ABSENT:--coverage
```

Could `crap-score` reuse the same runner-mode decision instead of duplicating an SDK-only table?

3. **The broad generation eval does not enforce `status.md`.** `code-testing-agent/SKILL.md:24-28` says every broad workflow is incomplete without research, plan, and status artifacts. In `tests/dotnet-test/code-testing-agent/eval.yaml`, all four broad stimuli have hard checks for `research.md` and `plan.md` (`:52-57`, `:142-147`, `:190-195`, `:243-248`), but none checks `.testagent/status.md`. Three rubrics mention status; the sparse-workspace rubric does not. Could all four broad stimuli add the deterministic status check, and could the fourth rubric match the contract?

4. **Eval-quality guidance is stale.** `eng/eval-quality/README.md:403-413` still says `platform-detection` is one of two hidden reference skills with a direct eval. The merged plugin makes it model-invocable, and the current gate proves the new state:

```text
1 reference skill(s) carry a direct-activation eval
    dotnet-test/filter-syntax
No errors.
```

Could the README leave only `filter-syntax` in this warning?

5. **The runner-property examples are malformed XML.** `platform-detection/SKILL.md:116,118` uses `<EnableMSTestRunner>true>` and `<EnableNUnitRunner>true>`. Could these use full elements such as `<EnableMSTestRunner>true</EnableMSTestRunner>` or unambiguous `EnableMSTestRunner=true` notation?

6. **A related migration reference still has the old range-argument order.** `plugins/dotnet-test-migration/skills/migrate-xunit-to-mstest/references/mapping-cheatsheet.md:126-127` maps to `Assert.IsInRange(value, low, high)` / `IsNotInRange(value, low, high)`. MSTest uses `(minValue, maxValue, value)`, as the fixed `writing-mstest-tests/SKILL.md:282` now shows. Could this shared mapping be corrected too?

I ran the merged eval-quality checks: all 24 self-tests passed, 99 eval specs were checked, and the gate reported no errors. The new Go and Python fixtures also compile/import successfully. These findings were independently reviewed across GPT, Claude, and Gemini model families.

PR #994, opened 2026-08-05
Add Ignite UI Blazor skill

kdinev:
## Summary

This pull request adds a new skill for integrating Ignite UI components into Blazor applications, along with an evaluation test for the skill. The main focus is to provide clear, file-oriented guidance for adding and configuring Ignite UI in various Blazor project types, especially split Blazor Web Apps.

## Validation

Introduced a new evaluation YAML (`eval.yaml`) that tests the skill’s ability to provide concrete, file-specific integration steps for using Ignite UI in a split Blazor Web App. The evaluation checks for explicit package references, service registration in both server and client, correct `_Imports.razor` usage, proper asset linking, and file-oriented instructions.

## Checklist


AbhitejJohn:
Tagging @danroth27 and @javiercn to help review.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+license: MIT
+name: use-igniteui-blazor
+description: >
+  Add, configure, or review Ignite UI component support in Blazor applications.
```
kotlarmilos, line 5: I think maximum of 1024 chars, please check if it fits

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+dotnet add package IgniteUI.Blazor.GridLite   # OSS lightweight grid (MIT)
+```
+
+## 2. `Program.cs`
+
+```csharp
+builder.Services.AddIgniteUIBlazor();   // all modules available
+```
+
+Pass `typeof(Igb<Name>Module)` values to eagerly pre-load a specific set instead:
```
kotlarmilos, line 36: Do you want to provide an example?

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,53 @@
+name: use-igniteui-blazor
+description: Evaluates the dotnet-blazor/use-igniteui-blazor skill
+type: capability
+defaults:
+  timeout: 10m
+  runs: 5
+stimuli:
```
kotlarmilos, line 7: This eval runs the same scenario five times, so it does not test the other supported setups, for example single project blazor server, MAUI blazor hybrid, GridLite only setup

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
AbhitejJohn: The eval infra should flag if it exceeds.

kotlarmilos:
> ❌ [use-igniteui-blazor] Skill description is 1,050 characters — maximum is 1,024. Shorten the description in SKILL.md frontmatter.

AbhitejJohn:
@kdinev , @kotlarmilos : Would you mind moving these changes to a branch against this repo please? That would help get PR evaluations going. They do not run against branches in forks.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
kdinev: @kotlarmilos The code snippet on lines 37-40 is the example.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
kdinev: @kotlarmilos @AbhitejJohn Brought the description down to 924 chars.

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
kdinev: @kotlarmilos I believe all review comments have been addressed (including the copilot one) in the latest commit. If everything looks OK now, I will work on moving this to a different branch, which is not in a fork. However evals should be able to be ran with a manual workflow approval off of forks.

kdinev:
@kotlarmilos Could you create a new branch in the repo, and I'll PR the change against that branch and that branch will be PRed against main for the workflow runs like @AbhitejJohn has requested? Alternative option is to give repo rights to our team to create branches and I'll handle it myself.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,111 @@
+---
+license: MIT
+name: use-igniteui-blazor
+description: >
+  Add, configure, or review Ignite UI component support in Blazor applications.
```
danroth27: Is it worth distinguishing between Ignite UI for Blazor and Ignite UI for Blazor Lite to avoid confusion with the commercially licensed components?

```suggestion
  Add, configure, or review Ignite UI for Blazor Lite component support in Blazor applications.
```

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+Add it to both `_Imports.razor` files in split Blazor Web App solutions.
+
+## 4. Host page — CSS and script
+
+Host page is `wwwroot/index.html` (WASM/MAUI), `Pages/_Host.cshtml` (Server), or `Components/App.razor` (Web App).
+
+```html
+<link href="_content/IgniteUI.Blazor/themes/light/bootstrap.css" rel="stylesheet" />
+...
+<script src="_content/IgniteUI.Blazor/app.bundle.js"></script>
```
danroth27, line 74: Is this manual script tag still required? `IgniteUI.Blazor.Lite` 0.1.1 already ships an `IgniteUI.Blazor.Lite.*.lib.module.js` Blazor JavaScript initializer that loads `app.bootstrap.js` and the component bundles. I verified in a default .NET 10 Blazor Web App that the components load and render correctly without adding `app.bundle.js` to `App.razor`. Could this guidance rely on the initializer instead, and document any older framework or hosting-model cases where a manual script is still necessary?

kdinev: @danroth27 Good catch! That was required, but the state of the skill is from before the change you mention. I will update accordingly.

kotlarmilos:
@AbhitejJohn I don't have push permissions to create a new branch on the repo. Could you please help us with that?

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
jsakamotoIGJP: Just to add some context.
As you said, app.bundle.js is loaded automatically through the .lib.module.js initializer.

On the other hand, our docs still show a script tag for cases that use Ignite UI for Blazor global functions such as igRegisterScript. The explicit reference gives a reliable point to run that code.

If those functions aren't needed during page load, app.bundle.js doesn't need to be referenced explicitly.

Please take a look related discussion: https://github.com/IgniteUI/igniteui-blazor/issues/233

damyanpetev: Yup, @jsakamotoIGJP made all valid points;
While we could leave the guidance that the script is only needed in some cases, I'd much rather drop it (leave it as legacy elsewhere), tackle the user scripts registration and completely update guidance for new code in this skill. Will need a lil bit to tackle, and thus delay the skill a bit if you don't mind. Will update on progress as soon as I can.

PS: The `app.bundle.js` is a bit a legacy artifact on our end - it was shipped before modules became a thing, and like these things often do, it stuck. Safety in stability and so on.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+<script src="_framework/blazor.web.js"></script>   <!-- or blazor.server.js / blazor.webassembly.js / blazor.webview.js -->
```
PureWeen, line 75: 💡 **Regression** — For MAUI Blazor Hybrid, `BlazorWebView` starts Blazor natively, so the host page should preserve `<script src="_framework/blazor.webview.js" autostart="false"></script>`. The current snippet presents `blazor.webview.js` as interchangeable with the auto-starting framework scripts, so an agent producing a complete MAUI host page could emit an incomplete tag. A one-line MAUI-specific callout would avoid that risk.

_Flagged by: 3/3 reviewers after dispute; calibrated as non-blocking after source-level verification._

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
...
+      I got a grid that was completely unstyled. My index.html currently has no Ignite UI tags in it at all.
+
+      Give me the exact package, startup and index.html changes for a grid-only setup, and don't have me pull in things I don't need.
+    graders:
+      - type: output-contains
+        config:
+          substring: IgniteUI.Blazor.GridLite
+      - type: output-not-contains
+        config:
+          substring: AddIgniteUIBlazor
```
PureWeen, line 138: ⚠️ **Testing** — These `output-not-contains` checks (`AddIgniteUIBlazor` here and `app.bundle.js` below) also reject a correct explanation such as "do not call `AddIgniteUIBlazor()` or add `app.bundle.js` for GridLite-only." This was reproduced against the repository-pinned Vally 0.14.0 grader: correct negated prose receives score 0. Match actual usage syntax, such as an `AddIgniteUIBlazor(` invocation or an `<script ...app.bundle.js>` tag, instead of banning the bare terms.

_Flagged by: 2/3 reviewers; empirically reproduced._

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
...
+          substring: app.bundle.js
+      - type: output-not-contains
+        config:
+          substring: '<link href="_content/IgniteUI.Blazor.GridLite/css/themes/light/bootstrap.css"'
+      - type: output-contains
+        config:
+          substring: _content/IgniteUI.Blazor/themes/light/bootstrap.css
+      - type: output-matches
+        config:
+          pattern: '(?:both|each)[\s\S]{0,80}(?:[Ss]erver[\s\S]{0,80}[Cc]lient|[Cc]lient[\s\S]{0,80}[Ss]erver)'
```
PureWeen, line 45: ⚠️ **Testing** — This pattern only requires `both` or `each` near `Server` and `Client` text anywhere in the response; it is not tied to `AddIgniteUIBlazor` or either `Program.cs`. Direct probes showed that it passes generic architecture prose and negated advice such as "do not register in both Server and Client," while sufficiently separated correct wording can fail. Tie the assertion to positive registration in both projects rather than word proximity alone.

_Flagged by: 2/3 reviewers; empirically reproduced._

PureWeen (commented):
### Adversarial review summary

Two eval correctness defects should be addressed:

- ⚠️ The GridLite-only negative substring graders penalize correct explanatory negations. This was reproduced using the repository-pinned Vally 0.14.0 implementation.
- ⚠️ The split-app Server/Client proximity regex produces both false positives and false negatives and does not establish dual service registration. This was reproduced with direct regex probes.

One non-blocking documentation clarification is also suggested:

- 💡 Preserve the MAUI-specific `autostart="false"` requirement when showing `blazor.webview.js`. First-party template and runtime source confirm the contract, although no native runtime reproduction was performed.

The five scenarios cover the intended project variants, but the two grader defects weaken the reliability of the grid-only and split-app cases. Existing review feedback about legacy `app.bundle.js` guidance and GridLite package/setup distinctions was not duplicated.

_Methodology: 3 independent reviewers with adversarial consensus, followed by targeted empirical revalidation. Review event: COMMENT._

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
...
+      - type: prompt
+    rubric:
+      - Identifies the package references for both IgniteUI.Blazor.Lite and IgniteUI.Blazor.GridLite
+      - Says AddIgniteUIBlazor() must be called in both the Server and Client Program.cs files for a split Blazor Web App
+      - Adds @using IgniteUI.Blazor.Controls to _Imports.razor so the components are available everywhere that needs them
+      - Includes both the Ignite UI theme stylesheet and app.bundle.js in the host page, with app.bundle.js before the Blazor framework script
+      - Does not suggest the GridLite stylesheet, which is only needed for a grid-only setup
+      - Keeps the answer specific to the relevant files instead of giving only general guidance
```
PureWeen, line 53: ⚠️ **Testing** — The only Blazor Web App stimulus never checks the skill’s load-bearing static-SSR guidance. An answer can provide every requested package, registration, import, and asset detail while saying nothing about an interactive render mode, yet a default static-SSR page would still render no usable Ignite UI component. Add a distinct Web App failure scenario where blank components must be diagnosed as missing `@rendermode Interactive*` (or extend this rubric, though a separate scenario gives a cleaner signal).

_Flagged by: 3/3 reviewers._

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
+name: use-igniteui-blazor
+description: Evaluates the dotnet-blazor/use-igniteui-blazor skill
+type: capability
+defaults:
+  timeout: 10m
+  runs: 5
+stimuli:
```
PureWeen, line 7: 💡 **Testing** — This eval has no dormancy contract even though the skill explicitly routes general render-mode and architecture questions to `create-blazor-project`. Because this skill also contains broad Blazor/render-mode vocabulary, it can incorrectly activate for a generic “which Blazor render mode should I choose?” request and the eval would never detect that routing regression. Add an `expect_activation: false` stimulus with no Ignite UI terms at that nearest sibling boundary.

_Flagged by: 3/3 reviewers._

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
...
+          substring: _content/IgniteUI.Blazor/themes/dark/material.css
+      - type: output-contains
+        config:
+          substring: app.bundle.js
+      - type: output-contains
+        config:
+          substring: blazor.webassembly.js
+      - type: output-matches
+        config:
+          pattern: 'Igb[A-Za-z]+Module'
```
PureWeen, line 189: 💡 **Testing** — `Igb[A-Za-z]+Module` accepts any single module-shaped token, including `IgbGridModule`, even though this scenario explicitly rules out grid usage and asks about inputs, a combo, and a dialog. That lets a contradictory answer satisfy the deterministic module check. Require the requested component modules, for example `Igb(Input|Combo|Dialog)Module` with enough assertions to establish the intended preload set, while leaving the “preload does not gate rendering” semantics to the prompt rubric.

_Flagged by: 3/3 reviewers._

PureWeen (commented):
### Additional eval review

A focused second pass using the repository’s eval-authoring rules found three additional non-duplicate improvements:

- ⚠️ **3/3 reviewers:** positively test the Blazor Web App static-SSR/render-mode failure that the skill claims to solve.
- 💡 **3/3 reviewers:** add an `expect_activation: false` boundary case for generic render-mode/architecture questions with no Ignite UI requirement.
- 💡 **3/3 reviewers:** tighten the generic module-name regex so a grid or unrelated module cannot satisfy the input/combo/dialog scenario.

The broader proposal to convert all mention checks into syntax/order checks was rejected: those semantic requirements are already valid prompt-grader responsibilities, and the concrete negative-grader/order defects are covered by existing threads. Prior feedback about the exact-five-stimulus floor and the split prompt naming `AddIgniteUIBlazor()` was also not duplicated.

_Test coverage assessment: the project-type matrix is broad, but it omits one core positive behavior (interactive render mode) and all routing-restraint coverage. Methodology: 3 independent reviewers with adversarial consensus. Review event: COMMENT._

kotlarmilos:
Moving this work to #1111 so the evaluation workflows can run against a branch in `dotnet/skills`.

kotlarmilos (changes requested):
Three issues need to be addressed before this can be approved.

1. `plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md:74-78`

`IgniteUI.Blazor.Lite` now loads its runtime through its Blazor JavaScript initializer, so `app.bundle.js` is not universally required and its absence does not generally prevent components from rendering. Please update this section to describe the initializer-based setup and any narrow legacy or global-function exception, then remove the `app.bundle.js` requirements from the corresponding eval graders and rubrics.

2. `plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md:46`

This should be scoped to split Web Apps using Interactive WebAssembly or Auto. A single-project Interactive Server Blazor Web App has no client `Program.cs` and should register `AddIgniteUIBlazor()` only in its server project. Please distinguish the single-project and split cases here and in the project-type table, and add an eval for the modern single-project Web App configuration.

3. `tests/dotnet-blazor/use-igniteui-blazor/eval.yaml:136-141`

These exclusions also reject correct negated guidance such as “do not call `AddIgniteUIBlazor()` or add an `app.bundle.js` script.” Please match actual usage syntax instead, such as an `AddIgniteUIBlazor(` invocation and an HTML script tag referencing `app.bundle.js`, so explanatory negations remain valid.

PR #997, opened 2026-08-07
Add golden end-to-end test generation scenarios

Evangelink:
## Why

The test plugin needs stable end-to-end scenarios for three workflows: creating and wiring the first .NET test project, empirically finding and closing gaps in an existing suite, and safely testing public static APIs that depend on ambient state.

## What changed

- Add public `scaffold-dotnet-test-project` for central packages, bounded project references, `.sln`/`.slnf`/`.slnx` registration, no-op reuse, and missing CI wiring.
- Extend `test-gap-analysis` so it can add focused tests only for empirically survived mutations. Its implementation scenario replays concrete mutations after generation.
- Add public `testability-obstacle` for nested, async-safe ambient overrides across time, filesystem, environment, and identity, plus built-in fake-time delays and a Python dormancy guard.
- Extend the testability migration agent through deterministic test generation when requested.
- Incorporate review fixes for SDK-style solution GUIDs, decimal literals, and unambiguous regex graders.

## Evaluation history and response

Early consumer-oriented revisions mostly tied and then regressed `code-testing-agent`; they also pulled stable sibling skills into evaluation. Those changes were removed. `code-testing-agent`, its generator, `generate-testability-wrappers`, and `migrate-static-to-wrapper` remain byte-for-byte identical to `main`.

Specialized runs then established repeatable signal:

- Run `31178542305`: gap analysis 3W/4T/0L; testability 3W/9T/0L.
- Same-commit runs `31185716948` and `31188608506` showed substantial judge variance, but also stable decisions.
- Run `31185716948`: testability passed 9W/7T/0L; scaffold was one p-value step short at 8W/4T/2L.
- Run `31188608506`: gap analysis passed 9W/4T/1L; testability was one win short at 6W/9T/1L.

This revision removes only stable negative/inert signal and adds independent proven-family evidence:

- Remove project-oriented/no-solution scaffolding, which lost 4/4 because both arms succeeded and judges rewarded extra unrequested tests.
- Add SDK `.slnx` project registration, complementing the repeatedly winning `.sln`/`.slnf` CI-wiring scenarios.
- Add a retry-policy guard/exception mutation scenario to the diverse gap-analysis eval.
- Remove machine identity (net-neutral with the only testability loss) and coupon randomness (4/4 ties). Historical records without them would pass in both same-commit runs.

## Validation

- `skill-validator check --plugin ./plugins/dotnet-test`
- Built the new `.slnx` and retry-policy fixtures
- Built all retained .NET fixtures and ran populated xUnit suites


tests/dotnet-test/testability-obstacle/eval.yaml
```diff
@@ -0,0 +1,183 @@
...
+          dest: fixtures/archive-writer
+    graders:
+      - type: run-command
+        config:
+          command: sh -c "dotnet test fixtures/archive-writer/tests/ArchiveWriter.Tests.csproj"
+          expected_exit_code: 0
+          timeout: 10m
+      - type: run-command
+        config:
+          command: sh -c "! grep -q 'DateTimeOffset.UtcNow\\|File.WriteAllText' fixtures/archive-writer/src/ArchiveWriter.cs"
```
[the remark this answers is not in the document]
Evangelink: Good catch - fixed in b3e6b1757. All three graders now use grep -E alternation, avoiding BRE and YAML escape ambiguity.

Evangelink (commented):
/evaluate

plugins/dotnet-test/skills/scaffold-dotnet-test-project/SKILL.md
```diff
@@ -0,0 +1,182 @@
+---
+name: scaffold-dotnet-test-project
+description: >-
+  Create the first .NET test project. USE FOR: "solution has no tests", first
+  xUnit tests, Tests.csproj/ProjectReference, add an omitted test project to
+  .sln/.slnx/.slnf, central packages, or tests missing from CI. DO NOT USE FOR:
+  a suitable existing project (stop; use code-testing-agent) or migration.
```
AbhitejJohn: Thanks for adding the omitted-wiring scenarios. Could we narrow this exclusion to a suitable project that is **already registered in the requested build entry point**? As written, the routing text says to stop for any suitable existing project, but the eval at lines 129-187 expects this skill to activate when an existing project is missing from `Catalog.sln` or `TextTools.slnf`. The body already has the correct narrower stop condition at lines 62-65, but the runtime routes from this frontmatter.

plugins/dotnet-test/agents/testability-migration.agent.md
```diff
@@ -73,6 +73,21 @@ Use the `migrate-static-to-wrapper` skill to:
 5. Verify the project builds
 6. Report what was changed and what remains
 
+### Phase 4: Test
+
+When the request includes tests, use `testability-obstacle` or
+`code-testing-agent` to:
```
AbhitejJohn: Could we split these two paths? After Phase 3, the seam is already injected, while `testability-obstacle` explicitly says not to use it when a dependency is already injected and to use `code-testing-agent` instead. Would it work to use only `code-testing-agent` here, and treat `testability-obstacle` as an alternative targeted path that skips Generate/Migrate rather than as the final phase after them?

plugins/dotnet-test/agents/testability-migration.agent.md
```diff
@@ -73,6 +73,21 @@ Use the `migrate-static-to-wrapper` skill to:
 5. Verify the project builds
 6. Report what was changed and what remains
 
+### Phase 4: Test
+
+When the request includes tests, use `testability-obstacle` or
```
AbhitejJohn: Could we add one agent-level scenario that requests migration **and** deterministic tests? This PR adds the full Phase 4 path and removes the old inter-phase confirmation behavior, but `tests/dotnet-test/agent.testability-migration/eval.yaml` still has only an analysis-only scenario and a targeted TimeProvider migration scenario. Neither exercises this new path, so the stated agent capability can regress without an eval noticing.

plugins/dotnet-test/skills/testability-obstacle/SKILL.md
```diff
@@ -0,0 +1,194 @@
+---
+name: testability-obstacle
+description: >-
+  Make a static C# API testable and add tests. USE FOR: deterministic Guid,
```
AbhitejJohn: Small routing consistency question: could this lead describe C# behavior with ambient dependencies rather than only a static API? The body supports constructor injection for instance classes, and the eval at lines 144-177 expects this skill for the instance `LeaseRenewer`/`TimeProvider` case. `USE FOR: time` may still route correctly, so I see this as an ambiguity rather than a proven activation failure.

AbhitejJohn (commented):
I rechecked the latest PR head. The four inline findings still apply.

plugins/dotnet-test/agents/testability-migration.agent.md
Evangelink: Added in 16b334cbc. The agent eval now includes a time migration plus deterministic xUnit scenario that must run the generated tests.

plugins/dotnet-test/skills/testability-obstacle/SKILL.md
Evangelink: Fixed in 16b334cbc. The router now leads with C# ambient-dependent behavior and explicitly covers constructor injection for instance classes.

plugins/dotnet-test/skills/scaffold-dotnet-test-project/SKILL.md
Evangelink: Fixed in 16b334cbc. The exclusion now stops only when the suitable project is already registered in the requested build entry point.

plugins/dotnet-test/agents/testability-migration.agent.md
Evangelink: Fixed in 16b334cbc. Phase 4 now uses code-testing-agent only; testability-obstacle is a separate targeted path that skips Generate/Migrate.

Evangelink (commented):
/evaluate

JanKrivanek (approved):
Looks good to go

PR #1010, opened 2026-08-13
Automate Vally dependency updates

JanKrivanek:
## Summary
- move the Vally and Copilot CLI pins into a trusted npm manifest consumed by the evaluation workflow
- configure weekly Dependabot updates for Vally while keeping Copilot manually pinned
- cover the trusted manifest install path in the workflow tests

## Testing
- `python -m unittest eng.evaluation.test_token_failover`
- `actionlint -shellcheck= -pyflakes= .github/workflows/*.yml`

### DIFF
--- a/.github/workflows/evaluation-run.yml
+++ b/.github/workflows/evaluation-run.yml
@@ -364,8 +364,9 @@ jobs:
           rm -rf "$RUNNER_TEMP/evaluation-tools"
           mkdir -p "$RUNNER_TEMP/evaluation-tools"
           cp "$RUNNER_TEMP/trusted-validator-src/eng/evaluation-tools/package.json" \
+            "$RUNNER_TEMP/trusted-validator-src/eng/evaluation-tools/package-lock.json" \
             "$RUNNER_TEMP/evaluation-tools/"
-          npm install \
+          npm ci \
             --prefix "$RUNNER_TEMP/evaluation-tools" \
             --userconfig "$RUNNER_TEMP/evaluation.npmrc" \
             --registry https://registry.npmjs.org/
--- a/.github/workflows/evaluation-run.yml
+++ b/.github/workflows/evaluation-run.yml
@@ -329,6 +329,7 @@ jobs:
           persist-credentials: false
           sparse-checkout: |
             eng/evaluation/nuget.config
+            eng/evaluation-tools
             eng/skill-validator
             eng/vally-adapter
             global.json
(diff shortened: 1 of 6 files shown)


AbhitejJohn (approved):
**Approved** — clean infra change; all CI green, 10/10 local tests pass, no blockers from cross-model review. Two observations noted inline.

eng/evaluation-tools/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "dotnet-skills-evaluation-tools",
+  "private": true,
+  "description": "Pinned command-line tools used by the skill evaluation workflow.",
+  "dependencies": {
+    "@github/copilot": "1.0.75",
```
AbhitejJohn, line 7: Direct versions are pinned but transitive dependencies float — no `package-lock.json` is committed. Production `npm install` (evaluation-run.yml:368) resolves transitives at build time, so the `trusted, pinned toolchain` guarantee covers only the top-level packages. A committed lockfile with `npm ci` would close this gap; without one, the floating transitives are an accepted trade-off worth noting.

eng/evaluation/test_token_failover.py
```diff
@@ -270,6 +277,27 @@ def test_eval_discovery_precedes_tool_install_and_token_selection(self) -> None:
             install_script,
         )
 
+    def test_evaluation_tool_manifest_has_secretless_smoke_test(self) -> None:
+        workflow_text = TEST_WORKFLOW.read_text(encoding="utf-8")
+        self.assertEqual(workflow_text.count('"eng/evaluation-tools/**"'), 2)
```
AbhitejJohn: This raw-text count assertion will break if the same glob appears in a YAML comment or a third trigger section. The parsed `on.pull_request.paths` and `on.push.paths` lists would be a more resilient target.

JanKrivanek: Updated the test to parse the workflow and assert that `eng/evaluation-tools/**` appears exactly once in each of `on.pull_request.paths` and `on.push.paths`, avoiding comment or unrelated-section false positives.

.github/workflows/evaluation-run.yml
```diff
@@ -354,19 +355,20 @@ jobs:
...
+          mkdir -p "$RUNNER_TEMP/evaluation-tools"
+          cp "$RUNNER_TEMP/trusted-validator-src/eng/evaluation-tools/package.json" \
+            "$RUNNER_TEMP/evaluation-tools/"
           npm install \
             --prefix "$RUNNER_TEMP/evaluation-tools" \
             --userconfig "$RUNNER_TEMP/evaluation.npmrc" \
-            --registry https://registry.npmjs.org/ \
-            @microsoft/vally-cli@0.9.0 \
-            @github/copilot@1.0.75
+            --registry https://registry.npmjs.org/
```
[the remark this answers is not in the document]
JanKrivanek: Agreed. I committed `eng/evaluation-tools/package-lock.json` and changed the trusted evaluation install to `npm ci`, so both direct and transitive dependencies are now deterministic.

eng/evaluation-tools/package.json
JanKrivanek: Addressed by committing the generated lockfile and switching both the production evaluation install and its smoke-test install to `npm ci`. Dependabot can now update the Vally pin and resolved graph together while the Copilot direct pin remains excluded by the allow-list.

eng/evaluation-tools/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "dotnet-skills-evaluation-tools",
+  "private": true,
+  "description": "Pinned command-line tools used by the skill evaluation workflow.",
+  "dependencies": {
```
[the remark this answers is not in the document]
JanKrivanek: Added `version: 0.0.1` to the private tooling manifest, matching the repository convention and keeping npm package metadata complete.

.github/workflows/evaluation-run.yml
```diff
@@ -354,19 +355,21 @@ jobs:
...
+          cp "$RUNNER_TEMP/trusted-validator-src/eng/evaluation-tools/package.json" \
+            "$RUNNER_TEMP/trusted-validator-src/eng/evaluation-tools/package-lock.json" \
+            "$RUNNER_TEMP/evaluation-tools/"
+          npm ci \
             --prefix "$RUNNER_TEMP/evaluation-tools" \
             --userconfig "$RUNNER_TEMP/evaluation.npmrc" \
-            --registry https://registry.npmjs.org/ \
-            @microsoft/vally-cli@0.9.0 \
-            @github/copilot@1.0.75
+            --registry https://registry.npmjs.org/
```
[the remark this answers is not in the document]
JanKrivanek: Fixed in `1d30e3075`: `eng/evaluation-tools/package-lock.json` is checked in, the trusted checkout copies both manifests, and production now installs with `npm ci`. The matching secretless `evaluation-tools` smoke job also uses `npm ci` and is passing on the current PR revision.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

AbhitejJohn (approved):

PR #1015, opened 2026-08-14
Fix bundled skill reference loading

### LINKED ISSUE #1008: Skills reference `references/*.md` by relative path, which silently fails when the skill lives outside the workspace
## Summary

Skills instruct the agent to load `references/*.md` using **workspace-relative** paths. When the
skill is installed outside the workspace (the normal case for VS Code agent plugins), workspace-scoped
search tools cannot resolve those paths. The agent concludes the references "aren't available",
silently falls back to the abbreviated inline recipes, and reports a degraded result as if it were complete.

## Environment

- VS Code + GitHub Copilot agent mode
- Skills installed at `~/.vscode/agent-plugins/github.com/dotnet/skills/...`
- Workspace open at an unrelated path (`~/Desktop/project/orders-api-review`)
- macOS

## Repro

1. Install the plugins; open a workspace anywhere *other than* the plugin directory.
2. Invoke `analyzing-dotnet-performance`.
3. Agent follows Step 1 and tries to resolve `references/critical-patterns.md`.
4. Workspace-scoped file search returns 0 matches — the glob is rooted at the workspace folder, and the
   skill directory is outside it.
5. Agent proceeds in fallback mode.

The files exist the whole time:

```
$ ls plugins/dotnet-diag/skills/analyzing-dotnet-performance/references/
async-patterns.md  collections-and-linq.md  critical-patterns.md
io-and-serialization.md  memory-and-strings.md  regex-patterns.md
structural-patterns.md
```

## Why it happens

`references/…` is relative, and the skill never says *relative to what*. The agent has no basis to
resolve it against the SKILL.md file's own directory, so it falls back to workspace-relative — the only
root its search tools accept.

## `analyzing-dotnet-performance` makes this materially worse

Step 1 instructs the agent to **stop trying** after the first failed lookup:

> **If reference files are not found** (e.g., in a sandboxed environment or when the skill is embedded as
> instructions only), **skip file loading and proceed directly to Step 3** using the scan recipes listed
> inline below. Do not spend time searching the filesystem for reference files — if they aren't at the
> expected relative path, they aren't available.

This converts a recoverable path-resolution failure into a permanent one. One well-intentioned
efficiency guardrail guarantees the degraded path.

## Impact

The fallback is **silent**. Nothing in the output tells the user the scan was partial.

In a real review, loading the references after the fact surfaced findings the inline recipes have no
detection for at all — System.Text.Json source-generation, anonymous types blocking source-gen, and
`System.Threading.Channels` for producer/consumer. The inline recipe list contains no equivalent checks,
so these were not missed through agent error; they were unreachable.

## Scope — not limited to this skill

`microbenchmarking/SKILL.md` uses the same relative-link convention with a much stronger mandate:

> **You MUST read the reference files relevant to the task before writing any code** — your training data
> likely contains outdated or incorrect BDN patterns.

...and links as `[references/comparison-strategies.md](references/comparison-strategies.md)`. Same
resolution failure, higher stakes: the skill's own premise is that unaided model knowledge of BDN is
wrong. Silently skipping the references yields exactly the outdated output the skill exists to prevent.

Other skills shipping `references/`: `dotnet-trace-collect`, `coverage-analysis`, and others.

## Suggested fix

1. **State the resolution rule explicitly.** Reference paths resolve against the SKILL.md file's own
   directory. Skills should say so, and instruct agents to read them with a direct file-read tool using
   an **absolute** path, not a workspace-scoped glob/search tool.
2. **Require one directory listing before declaring absence.** Replace "do not spend time searching" with
   a bounded check: list `<skill_dir>/references/` once. Cheap, deterministic, eliminates the false negative.
3. **Make degradation visible.** If references genuinely can't be loaded, require the agent to state so in
   its output. A partial scan presented as complete is worse than a failed one.
4. Consider templating the skill directory (e.g. `${SKILL_DIR}/references/…`) so the intended root is
   unambiguous.

Item 3 matters most: the current failure mode is invisible to the user.

AbhitejJohn:
## Summary

Fix silent degradation when a skill cannot load its bundled reference files. Bundled paths remain relative to the directory that contains `SKILL.md`; skills no longer treat one failed lookup as proof that the files are unavailable.

## Changes

- Update `analyzing-dotnet-performance` and `clr-activation-debugging` to:
  - resolve bundled files from the skill root, not the user workspace;
  - use one bounded listing of the known `references/` directory after a direct-read failure;
  - avoid workspace file or text search for the skill installation;
  - disclose reduced reference coverage in the final result.
- Update `create-skill` authoring guidance and its checklist so new skills use relative bundled-file paths and cannot degrade silently.

## Validation

- `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin ./plugins/dotnet-diag`
  - All checks passed: 7 skills, 1 agent, 1 plugin.
  - 99 eval specifications checked; no errors.
- `git diff --check`

Fixes #1008

### DIFF
--- a/plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
+++ b/plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
@@ -35,25 +35,33 @@ Scan C#/.NET code for performance anti-patterns and produce prioritized findings
 
 ## Workflow
 
-### Step 1: Load Reference Files
+### Step 1: Load Critical Reference
 
-Resolve bundled paths from the directory that contains this `SKILL.md`, not from the user's workspace. Load these reference files, which contain detailed detection recipes and grep commands:
+Resolve bundled paths from the directory that contains this `SKILL.md`, not from the user's workspace. Load this reference file first:
 
 - `references/critical-patterns.md`
-- `references/async-patterns.md`
-- `references/memory-and-strings.md`
-- `references/regex-patterns.md`
-- `references/collections-and-linq.md`
-- `references/io-and-serialization.md`
-- `references/structural-patterns.md`
 
 If a direct read fails, list this skill's `references/` directory once and retry only when the listing shows the expected file. Do not use workspace file or text search to locate the skill installation.
 
-If one or more expected reference files remain unavailable, use the loaded references where possible and the inline recipes in Step 3 for the missing coverage. Include `Reference coverage: reduced; unavailable: <paths>; used inline recipes for missing references.` in the final report, with `<paths>` replaced by the missing relative paths.
-
 ### Step 2: Detect Code Signals and Select Topic Recipes
 
-Scan the code for signals that indicate which pattern categories to check. Use the `## Detection` sections from loaded reference files and the inline recipes in Step 3 for categories whose reference files are unavailable.
+Scan the code for signals that indicate which pattern categories to check. Use the `## Detection` section from the critical reference when available and the inline recipes in Step 3 for initial signal detection.
+
+After detecting signals, load only the topic-specific references selected by scan depth:
+
+- `critical-only`: No additional references (use only `critical-patterns.md`)
+- `standard` (default): Load references matching detected signals from this list:
+  - `references/async-patterns.md` — async/Task/ValueTask signals
+  - `references/memory-and-strings.md` — Span/Memory/string allocation signals
+  - `references/regex-patterns.md` — Regex signals
+  - `references/collections-and-linq.md` — Dictionary/List/LINQ signals
+  - `references/io-and-serialization.md` — JsonSerializer/HttpClient/Stream signals
+  - `references/structural-patterns.md` — always loaded (unsealed classes checked regardless)
+- `comprehensive`: Load all six topic-specific references above
+
+For coverage reporting, the selected references are `references/critical-patterns.md` plus only the topic-specific references selected above. If any selected reference remains unavailable after retry, use the inline recipes in Step 3 for the missing coverage. Include `Reference coverage: reduced; unavailable: <paths>; used inline recipes for missing references.` in the final report, with `<paths>` replaced by the missing relative paths of the selected references only.
+
+Use the `## Detection` sections from loaded reference files and the inline recipes in Step 3 for categories whose reference files are unavailable.
 
 | Signal in Code | Topic |
 |----------------|-------|
--- a/plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
+++ b/plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
@@ -12,6 +12,8 @@ license: MIT
 
 Scan C#/.NET code for performance anti-patterns and produce prioritized findings with concrete fixes. Patterns sourced from the official .NET performance blog series, distilled to customer-actionable guidance.
 
+> **Resolving file paths:** All `references/` and `scripts/` paths in this skill are relative to the skill’s own directory, not the workspace. Before loading any reference file, locate this skill’s installation directory by searching for a directory named `analyzing-dotnet-performance` containing a `SKILL.md` file. It may be under the workspace (`.agents/skills/` or `plugins/`), or in an external location such as a VS Code extensions folder. Resolve all `references/` and `scripts/` paths from that directory.
+
 ## When to Use
 
 - Reviewing C#/.NET code for performance optimization opportunities
--- a/plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
+++ b/plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
@@ -49,11 +49,11 @@ Resolve bundled paths from the directory that contains this `SKILL.md`, not from
 
 If a direct read fails, list this skill's `references/` directory once and retry only when the listing shows the expected file. Do not use workspace file or text search to locate the skill installation.
 
-If the reference files remain unavailable, proceed to Step 3 with the inline recipes and include `Reference coverage: unavailable; used inline recipes only.` in the final report.
+If one or more expected reference files remain unavailable, use the loaded references where possible and the inline recipes in Step 3 for the missing coverage. Include `Reference coverage: reduced; unavailable: <paths>; used inline recipes for missing references.` in the final report, with `<paths>` replaced by the missing relative paths.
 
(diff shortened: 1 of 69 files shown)


AbhitejJohn (commented):
/evaluate 0d9c3dff328abeee84a4cb7ca46fd3785972b9cd

ChrisAhna (approved):

plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
```diff
@@ -35,11 +35,13 @@ Scan C#/.NET code for performance anti-patterns and produce prioritized findings
 
 ## Workflow
 
-### Step 1: Load Reference Files (if available)
+### Step 1: Load Reference Files
 
-Try to load `references/critical-patterns.md` and the topic-specific reference files listed below. These contain detailed detection recipes and grep commands.
+Resolve bundled paths from the directory that contains this `SKILL.md`, not from the user's workspace. Load `references/critical-patterns.md` and the topic-specific reference files listed below. These contain detailed detection recipes and grep commands.
```
Evangelink: The topic-specific references aren't actually listed below—the only literal path here is `references/critical-patterns.md`. Because the fallback listing is allowed only after a direct read fails, that successful read gives the agent no way to discover `async-patterns.md`, `memory-and-strings.md`, `regex-patterns.md`, `collections-and-linq.md`, `io-and-serialization.md`, or `structural-patterns.md`. Please enumerate those relative paths (or explicitly allow listing `references/` before topic selection) so the bundled recipes are reliably loaded.

Evangelink (changes requested):
One bundled-reference loading issue needs to be addressed before approval.

plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
AbhitejJohn: Thanks for catching this. I explicitly listed all seven bundled reference paths in `99922ce6`, so topic selection no longer depends on discovering files after the first successful read.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
```diff
@@ -35,11 +35,21 @@ Scan C#/.NET code for performance anti-patterns and produce prioritized findings
...
+- `references/async-patterns.md`
+- `references/memory-and-strings.md`
+- `references/regex-patterns.md`
+- `references/collections-and-linq.md`
+- `references/io-and-serialization.md`
+- `references/structural-patterns.md`
+
+If a direct read fails, list this skill's `references/` directory once and retry only when the listing shows the expected file. Do not use workspace file or text search to locate the skill installation.
+
+If the reference files remain unavailable, proceed to Step 3 with the inline recipes and include `Reference coverage: unavailable; used inline recipes only.` in the final report.
```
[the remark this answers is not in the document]
AbhitejJohn: Good catch. In `5854dfdc`, the performance skill now reports reduced coverage whenever one or more expected references are missing, names those paths, and uses inline recipes only for the missing coverage.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

plugins/dotnet-diag/skills/clr-activation-debugging/SKILL.md
```diff
@@ -98,13 +98,15 @@ After capturing, **remove the env var or registry key** to stop logging.
...
+Resolve bundled paths from the directory that contains this `SKILL.md`, not from the user's workspace. Load the reference files in this order — they contain the detailed log format, decision flow, and CLSID registry documentation:
 
 1. `references/log-format.md` — Log line format, fields, and all known log message types
 2. `references/activation-flow.md` — The shim's decision tree for runtime selection
 3. `references/com-activation.md` — COM (DllGetClassObject) activation specifics, CLSID registry layout
 
-If reference files are not available, proceed using the inline knowledge below.
+If a direct read fails, list this skill's `references/` directory once and retry only when the listing shows the expected file. Do not use workspace file or text search to locate the skill installation.
+
+If the reference files remain unavailable, proceed with the inline knowledge below and include `Reference coverage: unavailable; used inline guidance only.` in the final diagnosis.
```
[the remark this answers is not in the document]
AbhitejJohn: Good catch. In `5854dfdc`, any missing expected reference now requires reduced-coverage disclosure with the missing paths, while available references still remain in use.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

plugins/dotnet-diag/skills/analyzing-dotnet-performance/SKILL.md
```diff
@@ -35,15 +35,25 @@ Scan C#/.NET code for performance anti-patterns and produce prioritized findings
...
+Resolve bundled paths from the directory that contains this `SKILL.md`, not from the user's workspace. Load these reference files, which contain detailed detection recipes and grep commands:
 
-**If reference files are not found** (e.g., in a sandboxed environment or when the skill is embedded as instructions only), **skip file loading and proceed directly to Step 3** using the scan recipes listed inline below. Do not spend time searching the filesystem for reference files — if they aren't at the expected relative path, they aren't available.
+- `references/critical-patterns.md`
+- `references/async-patterns.md`
+- `references/memory-and-strings.md`
+- `references/regex-patterns.md`
+- `references/collections-and-linq.md`
+- `references/io-and-serialization.md`
+- `references/structural-patterns.md`
```
Evangelink: This now instructs the agent to load all seven references before Step 2 detects topic signals or applies the requested scan depth. That regresses the previous topic-specific loading behavior: even `critical-only` must read unrelated async, regex, collections, and I/O guidance (about 32 KB total) before being told to check only critical patterns. Please keep the paths enumerated, but make loading conditional—for example, load the critical reference first, then load only the references selected by scan depth and detected signals (all of them for `comprehensive`). The reduced-coverage note should likewise cover only the references selected for that run.

AbhitejJohn: Thanks for catching this. I accepted the feedback and changed the workflow so only `critical-patterns.md` loads before signal detection; topic references now load according to scan depth and detected signals, and reduced-coverage reporting applies only to references selected for that run.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

Evangelink (approved):

PR #1045, opened 2026-08-24
Add general .NET vectorization skill

### LINKED ISSUE #894: [skills-eval] dotnet-experimental: 3 skills, 47% pass — 1 to strengthen, 2 keep
## Context — cross-family skill evaluation: `dotnet-experimental`

This issue is **self-contained**: it captures everything a skill author needs to act on the `dotnet-experimental` plugin without opening the full report.

**What this measures.** Every runnable skill in `dotnet/skills` was run through [Vally](https://github.com/microsoft/evaluate) (0.7) on a **cross-family matrix**: 5 executor model families — `opus-4.8`, `gpt-5.5`, `sonnet-4.6`, `haiku-4.5`, `mai-flash` — each judged by a **different family** (judge ≠ executor; default judge = latest Opus, or GPT when Opus is the executor). For every skill, a **skilled** run is compared against a **baseline** (no-skill) run and scored per executor. This removes single-model and self-judging bias, so a skill that only helps one model family — or only its own family's judge — is visible.

- **Data source:** cross-family CI grid (run `29228914412` + backfills) — 5 executors × 85 runnable skills, **419 scored cells** over 84 skills.
- **Row grain:** one row per **skill**, aggregated across its (up to 5) executor cells. `avgN` is the mean trial count behind the cells (trials 1–17; higher = more statistically trustworthy). `thin-N` flags directional-only rows.

### `dotnet-experimental` at a glance (portfolio scorecard)

| Plugin | Skills | Cells | Pass | Impact | Tie-trials | Err | avg ΔTok | avg ΔTurns | avg ΔTools | Headline |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| **dotnet-experimental** | 3 | 15 | 47% | 0.474 | 7 | 0 | +17,562 | +0.56 | +0.53 | Healthy; keep-polish |

**Insights.** 3 skill(s); mean impact **0.47**; 3/3 help ≥1 frontier model. Exemplars (win incl. frontier): `exp-test-maintainability`. 

**Address first:** nothing critical — polish the `STRENGTHEN` rows below.

### How to read the table

Each skill is scored **skilled vs. baseline** on these axes:

| Signal | Column | What it means | Good |
| --- | --- | --- | --- |
| Breadth | `Families ✓` (n/5) | How many of the **5 model families** the skill helps (per-family pass count, max `5/5`) | 4–5 / 5 |
| Where | `Passed on` | *Which* families passed. **Frontier = latest Opus + latest GPT** are **bold**; Sonnet 4.6 / Haiku 4.5 / MAI Flash are mid/low-weight | frontier ✓ |
| Magnitude | `Impact` (−1…+1) | How strongly the judge prefers *skilled* over *baseline* | ≥ 0.4 |
| Decisiveness | `Ties▵` | Trials where the judge saw **no difference** → skill is *inert* | low |
| Safety | `Loss▵` | Trials where **skilled was WORSE** than baseline → skill misfires | ~0 |
| Reliability | `Err` | Trials that errored/crashed in setup or judging | 0 |
| Efficiency | `ΔTok` / `ΔTurns` / `ΔTools` | Extra tokens / agent turns / tool calls vs baseline | ≤ 0 |
| Confidence | `avgN` | Mean trials behind the verdict; low N = directional only | ≥ 3 |
| Invocation | `Call%` | Share of skilled trials where the model actually **invoked the skill** | ~100% |

> **`Families ✓` is cell-level (max `5/5`); `Ties▵`/`Loss▵` are trial-level tallies summed across *all* families** (including the ones where the skill failed). A high `Families ✓` next to non-zero `Loss▵` is not a contradiction — see `Passed on` and the Action text for where losses landed.

**Action buckets** (each skill has one primary action; `[flags]` note secondary concerns):

| Bucket | Priority | Meaning |
| --- | --- | --- |
| FIX-RELIABILITY | 🔴 P0 | Errored trials / no verdict — stabilize the harness before trusting the score |
| FIX-DISCOVERY | 🔴 P0 | Model doesn't invoke it (`Call% < 50%`) — a triggering/description problem |
| FIX-REGRESSION | 🔴 P0 | Skilled is worse than baseline on many trials — the skill misfires |
| ADD-DECISIVENESS | 🟠 P1 | Called ~100% but ties dominate, ~0 impact — inert; needs sharper behavioral steps |
| TRIM-COST | 🟠 P1 | Passes but with heavy token/turn overhead — trim verbosity |
| EXEMPLAR | 🟢 keep | Broad, strong, reliable win — use as a template |
| EFFICIENT-WIN | 🟢 protect | Wins *and* cuts turns/tools — the ideal shape |
| KEEP-POLISH | 🟢 | Solid majority win; minor polish + more trials |
| STRENGTHEN | 🟡 P2 | Marginal/mixed lift — sharpen triggers & success criteria |

### Per-skill actions

| Skill | Families ✓ | Passed on (frontier **bold**) | Impact | Ties▵ | Loss▵ | Err | avgN | Call% | ΔTok | ΔTurns | ΔTools | Action |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `exp-simd-vectorization` | 3/5 | **GPT**, Haiku, MAI | 0.57 | 1 | 2 | 0 | 3.8 | 100% | +36,894 | +1.27 | +1.33 | **KEEP-POLISH** · 🟢 **Solid majority win** on **GPT**, Haiku, MAI. Losses/ties are concentrated in the **non-passing** cells (Opus, Sonnet), not the 3 passing ones — the passes are clean. To firm up: (1) raise trials on Opus, Sonnet; (2) investigate why a frontier model (Opus) didn't pass — that's the highest-value gap; (3) minor wording polish only. *([frontier-miss], [frontier-regressed])* |
| `exp-test-maintainability` | 2/5 | **Opus**, Haiku | 0.51 | 1 | 2 | 0 | 4 | 100% | +7,893 | +0 | -0.25 | **EFFICIENT-WIN** · ✅ **Better *and* cheaper.** Wins on **Opus**, Haiku while saving 0.25 tools. What's good: it adds value without bloat — the ideal shape. Protect the brevity; don't let it grow. ⚠️ Frontier miss on GPT — worth a look. *([frontier-miss], [efficient])* |
| `exp-mock-usage-analysis` | 2/5 | **GPT**, Haiku | 0.34 | 5 | 2 | 0 | 6 | 100% | +7,899 | +0.4 | +0.5 | **STRENGTHEN** · 🟡 **Marginal/mixed** (passes 2/5 — **GPT**, Haiku). Diagnosis: mostly *ties* — too generic/non-prescriptive. Try: (1) sharpen the trigger so it fires only where it wins; (2) add 1–2 opinionated, concrete steps that change behaviour; (3) add trials to separate signal from noise. *([frontier-miss])* |

---

<sub>Generated from the cross-family Call-to-Action report (`CALL-TO-ACTION.md` §4–§5; companion `IMPACT-ANALYSIS.md`). Regenerate the underlying tables with `node deep-metrics.mjs "$env:TEMP\cf-ci" agg-ci` → `node gen-cta-tables.mjs agg-ci`. Numbers are directional where `avgN` is low; treat single-trial cells as hypotheses to confirm with more runs.</sub>

tannergooding:
Replaces #1044, recreated from a branch hosted directly in `dotnet/skills` so the full CI workflow can run.

Ports dotnet/runtime's vectorization guidance into the general `dotnet-advanced` plugin and retires the experimental SIMD skill.

The port keeps portable API selection, safe span loads and stores, self-contained width dispatch, vectorized remainder handling, fallback validation, and measured wider-width decisions aligned with the official .NET guidance. It also adds ten stimuli covering review, authoring, framework reuse, portability, memory safety, and routing boundaries.

Targeted local model runs exposed several places where concise guidance was not enough:

- Claude Opus 5 produced the intended safe span and vector-tail shape, while GPT-5.5 initially used `LoadUnsafe`/`StoreUnsafe` with a scalar epilogue. Adding the small safe inner-loop template changed the GPT-5.5 result to `Vector128.Create`/`CopyTo` with a preloaded overlapping vector tail.
- Claude Opus 5, GPT-5.5, Gemini 3.1 Pro, and MAI Code 1.1 Flash initially used the naive `Vector256 && Length >= Count` followed by `Vector128 && Length >= Count` cascade. The skill now explicitly defers to the official comprehensive templates and requires each widest-first hardware block to select its vector helper or dedicated small-input helper and return. Gemini and MAI reruns then produced that self-contained dispatch shape.
- Claude Opus 5, GPT-5.5, and MAI migrated the broken `SumValues` fixture from managed-reference loading to safe span creation with a masked vector tail. Gemini retained `MemoryMarshal.GetReference`/`LoadUnsafe` despite marking its own result as passing, which motivated deterministic graders for safe loading and fallback guards rather than relying only on a judge rubric.
- Review probes also showed that some initial rubrics would have rewarded repairing handwritten SIMD when an existing accelerated span operation was the better answer. Those rubrics now prefer framework reuse first and grade an explicit-loop repair only when it remains necessary.

These were targeted local stimulus runs rather than the official Vally comparison; the PR evaluation remains the source of truth for measured baseline delta.

Closes #894.

> [!NOTE]
> This pull request description was drafted with GitHub Copilot.


AbhitejJohn:
Thanks a bunch for the thorough skill and eval coverage here. I reviewed the skill, fixtures, and eval design across Claude, GPT, and Gemini. The SIMD guidance is strong overall. I found three areas worth adjusting:

1. **Managed-reference rule and `LastMatch` fixture**

   `SKILL.md` currently says not to create a managed reference past the end of an object, and the `LastMatch` rubric treats its one-past-end reference as inherently unsafe.

   ECMA-335 permits a managed reference to the address where an element just past the end of an array would be stored, provided it isn't dereferenced. The .NET runtime addendum states this directly:

   > "Managed pointers which point at the address just past the end of an object, or the address where an element just past the end of an array would be stored, are permitted but not dereferenceable."

   Sources: [ECMA-335 II.14.4.2](https://github.com/stakx/ecma-335/blob/master/docs/ii.14.4.2-managed-pointers.md) and the [runtime specification addendum](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md).

   Since `LastMatch` accepts an arbitrary `ReadOnlySpan<int>`, span-end and array-end aren't always the same address. Could we avoid depending on that distinction and make the fixture demonstrate the unambiguous defect: subtracting from element zero and creating a managed reference before the start of the span? The rubric could then ask the agent to identify that final subtraction and recommend subtracting only when another element remains.

2. **The evals sometimes verify an exact implementation instead of the required behavior**

   Several hard graders require or reject specific source tokens:

   - `ConditionalSelect` requires one tail strategy for `SumValues`.
   - `ToZeroSmall` requires a private helper to have one exact name.
   - Checks for `Vector128.Create` and `CopyTo` reject other correct load/store forms.
   - Negative checks for `LoadUnsafe`, `MemoryMarshal`, `System.Runtime.Intrinsics.X86`, and `Vector256` can reject valid, bounds-safe implementations. They can also match comments rather than executable code.

   For example, a solution can vectorize every full block and use a correct scalar epilogue without using `ConditionalSelect`. It would satisfy the prompt's requirement to retain SIMD, pass the boundary and overflow tests, and still fail the token grader. Similarly, renaming `ToZeroSmall` to `HandleSmallInput` would fail without changing behavior.

   This also conflicts with the repo's eval-authoring guidance in `.agents/skills/create-skill-test/SKILL.md`:

   > "Test outcomes, not methods"  
   > "Accept any valid approach."

   The existing `run-command` graders already provide stronger evidence. They execute exact-width, nonmultiple, empty, large, hardware-disabled, and AVX2-disabled paths. Could we remove the technique and private-name checks, then use the prompt rubric for the remaining outcomes: bounds safety, scalar equivalence, portable fallback, no double processing, and retained SIMD for full-vector inputs?

   I'd also combine conditional rubric pairs such as "prefer an existing span API" and "if retaining explicit SIMD..." into one independently evaluable item that accepts either valid result.

3. **Make the framework-first rule decisive**

   In the Product scenario, a skilled trajectory selected custom SIMD while another trajectory used `TensorPrimitives`. The skill already recommends high-level APIs, but the current wording still leaves room to continue into handwritten vector code after a matching framework operation is found.

   Could we make that rule explicit: use the highest-level API that preserves the contract, then stop? For explicit SIMD, it would also help to require the response to name the framework APIs considered and explain why none preserves the required behavior. Wider `Vector256<T>`, `Vector512<T>`, or `Vector<T>` paths should remain measurement-driven.

After these skill and eval changes, a fresh official evaluation would give useful evidence for activation and behavior on the revised content.

tannergooding:
> Managed-reference rule and LastMatch fixture

This is an area where even if the edge is technically allowed, it remains an inherently unsafe and dangerous operation that leads to a large number of bugs and potential security issues, especially where lesser LLMs infer it to be extendable to other cases or make mistakes.

Our official guidance, especially with the work around `unsafe evolution` and reducing the use of unnecessary unsafe is that it should essentially never be done and so having the skill reject it is ultimately better.

> For example, a solution can vectorize every full block and use a correct scalar epilogue without using ConditionalSelect. It would satisfy the prompt's requirement to retain SIMD, pass the boundary and overflow tests, and still fail the token grader. Similarly, renaming ToZeroSmall to HandleSmallInput would fail without changing behavior.

The `ToZeroSmall` is a case where it should probably be relaxed.

The "scalar epilogue" would be a failure of the skill to correctly vectorize the method (i.e. is invalid). Leaving scalar epilogues is essentially never required and is something that a skill that is explicitly meant to vectorize the code should not be leaving behind.

I'll look at doing cleanup for other cases mentioned as appropriate.

tannergooding:
Did some cleanup.

The main thing worth pointing out is that the requirements for `Create` (over Load), `CopyTo` (over Store), and the vectorized tails are a matter of safety and overall correctness. This is a known area where models try to do the "wrong thing" because they have wrong historical assumptions (and is something that Gemini and other models were consistently getting wrong without the guidance).

Even where these things are "technically equivalent", it is explicitly unnecessary risk and complexity that we want to push models away from for such an advanced skill. The simpler safe thing is the correct thing and what we want to push as the idiomatic way to handle it, how we want existing code fixed up to be as well to reduce the known risk and bugs that occur.

AbhitejJohn (approved):
Thanks for the detailed responses and updates. Taking your comment as the authoritative design intent, I agree that the skill can prohibit one-past-end managed references as a conservative policy, even though the runtime permits them when they are not dereferenced.

I think the current text still differs from that intent. `SKILL.md` says a collection can leave the reference untracked, and the `LastMatch` rubric requires that explanation. Would you be open to stating this as an intentional safety policy - permitted by the runtime, but prohibited by this guidance because it is fragile and easy to misuse - without attributing it to lost GC tracking?

On the eval checks, I agree that `Vector128.IsHardwareAccelerated`, `Vector128.Create(span)`, `CopyTo(span)`, and the prohibition on unmeasured wider paths enforce requirements that the skill states explicitly. In particular, the disabled-intrinsics test proves equivalent results, but it cannot prove that the scalar branch ran because `Vector128` operations have a correct software fallback. Keeping the `Vector128.IsHardwareAccelerated` check makes sense.

My remaining eval question is specifically about requiring the literal `ConditionalSelect` token. A tail can remain fully vectorized and avoid duplicate work by masking the overlapping vector with bitwise operations. Conceptually:

```csharp
Vector128<int> tail = Vector128.Create(finalFullSpan);
Vector128<int> mask = CreateMaskForUnprocessedLanes(overlap);
sum += tail & mask;
```

Here, lanes already processed by the main loop are zero in `mask`, and lanes that belong to the remainder are all bits set. This is equivalent for the reduction to `ConditionalSelect(mask, tail, Vector128<int>.Zero)`. It has no scalar epilogue and does not count overlapping lanes twice.

Would it make sense for the rubric to accept either `ConditionalSelect` or an equivalent vector masking operation, while the existing nonmultiple-length execution cases verify that no lanes are duplicated?

I also read the latest official result as `0 improved, 2 not proven improved`. That does not prove a correctness regression, but it also does not establish an improvement at this head.

tannergooding:
> I think the current text still differs from that intent. SKILL.md says a collection can leave the reference untracked, and the LastMatch rubric requires that explanation. Would you be open to stating this as an intentional safety policy - permitted by the runtime, but prohibited by this guidance because it is fragile and easy to misuse - without attributing it to lost GC tracking

Have a fix for that up now.

> My remaining eval question is specifically about requiring the literal ConditionalSelect token. A tail can remain fully vectorized and avoid duplicate work by masking the overlapping vector with bitwise operations. Conceptually:

The conceptual simplification can actually be worse for codegen, quality, and extensibility. We explicitly recognize and optimize the `ConditionalSelect` pattern and emit just `mask & x` if that's allowed and otherwise the general pattern is the one that gets broader and more general lightup for things like embedded masking. We sometimes recognize a bitwise-and as well, but it doesn't have all the same general handling involved and can get morphed in a way that breaks the pattern too. -- Something we'll continue improving over time, but where its a lot simple to push agents down the "golden path".

AbhitejJohn (approved):

AbhitejJohn:
Thanks for the fixes tanner, LGTM. Tagging @artl93 for any additional thoughts. Otherwise, looks good to go.

artl93:
@AbhitejJohn - If your referring to expertise on vectorization, I cannot comment. I rely on Tanner.

AbhitejJohn:
@artl93 : Sounds good, merging this in then :) Thanks for the addition Tanner, looking forward to more.

PR #1075, opened 2026-08-27
Fix MSTest version and retry guidance

Evangelink:
## Summary

- Corrects `Assert.IsInstanceOfType` guidance across MSTest 3.0-3.3, 3.4-3.11, and 4.x.
- Uses a bounded delayed backoff for flaky external-service retries and preserves real test/cleanup behavior instead of placeholder bodies.
- Rewrites the two losing rubrics around observable correctness and installed-version decisions; no stimuli were added after observing the result.

## Evidence and decision

PR #1054's valid Luna run ([32852289045](https://github.com/dotnet/skills/actions/runs/32852289045)) measured 7W/5T/2L. The two losses were concrete:

- **Runtime type mismatch:** the skilled answer treated all MSTest 3.x releases as supporting the generic `out` overload. Microsoft Learn's API monikers place that overload in 3.4-3.11; v4 returns the typed value directly, while 3.0-3.3 need the non-generic assertion followed by a safe post-assert cast.
- **Environment retry/cleanup:** the skilled answer emitted comment-only test bodies and immediate `[Retry(3)]`; the baseline retained representative operations and used the documented nonzero exponential backoff.

`migrate-static-to-wrapper` was also investigated from PR #1057's valid Luna run ([32983373666](https://github.com/dotnet/skills/actions/runs/32983373666)): 4W/5T/0L. All five tied transcripts reached functionally equivalent correct code, and the existing rubrics already assert behavior, scope, tests, and truthful validation. No skill or eval change is justified there; adding breadth now would only chase the observed p-value.

## Validation

- `skill-validator check --plugin ./plugins/dotnet-test`
- YAML parse and `git diff --check`


Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

YuliiaKovalova (approved):

PR #1077, opened 2026-08-27
Fix nested testability override restoration

Evangelink:
## Summary

- teaches `testability-obstacle` an explicit exception-safe `AsyncLocal<T>` scope that captures and restores the previous provider in LIFO order;
- makes the restoration test discriminating by choosing an outer value different from the production fallback, so an incorrect `Dispose()` that clears the slot cannot pass;
- strengthens only the existing environment-restoration stimulus with disposal, parallel-overlap, and no-process-environment-mutation checks;
- leaves routing unchanged because the only plugin-arm miss was the identity-generation stimulus, while the description already names `Guid`, identity generation, static API preservation, and nested/parallel overrides.

## Evidence and root cause

Authoritative evidence is PR #1057, workflow run [32983373666](https://github.com/dotnet/skills/actions/runs/32983373666), evaluated commit `1cc012d18effd5c5cd3fed3c85135fecb4daccc2`.

- Sonnet: 3W/4T/1L over 8 stimuli (`d=4`, `p=.312`, net `+25%`); repeated evidence 6W/6T/4L.
- Luna: 6W/2T/0L (`d=6`, `p=.016`, net `+75%`).
- Isolated activation: 8/8; plugin activity: 7/8.
- The repeatable Sonnet loss was **Restore environment overrides after exceptions and nesting**: 0W/0T/2L.

The first losing judge wrote:

> Both solutions appear to satisfy every requested behavior. A has a modest quality edge through a compositional immutable override stack and a clean warning-free final verification, while B is also substantively correct and well-tested.

The second wrote:

> Both solutions meet the central requirements and pass their stated tests. A supplies a robust immutable stacked implementation and broader directly relevant async/nesting coverage. B is a compact valid design, but its explanation of AsyncLocal child-flow behavior is imprecise and its final claim that tests are clean conflicts with the remaining warning shown in its timeline.

Both skilled trials used an outer `false`, inner `true`, and an absent environment fallback of `false`. Their post-exception `Assert.False` therefore still passed if `Dispose()` incorrectly cleared the slot instead of restoring the outer provider. The preferred baseline used outer `true`, inner `false`, so clearing the slot was observable. This change encodes that discriminator directly rather than adding another stimulus or repeated-run padding.

## Validation

- `dotnet artifacts\bin\SkillValidator\debug\skill-validator.dll check --plugin .\plugins\dotnet-test`
- `dotnet exec artifacts\bin\SkillValidator.Tests\debug\SkillValidator.Tests.dll -class SkillValidator.Tests.AnalyzeSkillTests -class SkillValidator.Tests.EvalSchemaTests -maxThreads 2 -reporter quiet -noLogo`
- `dotnet build tests\dotnet-test\testability-obstacle\fixtures\static-feature-flags\tests\StaticFeatureFlags.Tests.csproj --no-restore --nologo --verbosity minimal -nr:false`
- compiled runtime probe covering nested exception restoration and parallel async-flow isolation

### DIFF
--- a/plugins/dotnet-test/skills/testability-obstacle/SKILL.md
+++ b/plugins/dotnet-test/skills/testability-obstacle/SKILL.md
@@ -107,20 +107,73 @@ Use built-in fake-time-aware overloads instead of inventing an `IDelay` wrapper:
 Test delayed behavior by starting the operation, proving it is incomplete,
 advancing `FakeTimeProvider`, then awaiting it. Never wait for wall-clock time.
 
-For a nested ambient override, disposing the inner scope must restore the outer
-value, not clear the slot. Capture the previous value per scope:
+For a nested ambient override, each scope owns the value that was active when it
+started. Dispose scopes in LIFO order with `using` (which emits `try/finally`) or
+an explicit `finally`; disposing the inner scope restores the outer value, never
+an unconditional `null`. For an environment-backed static API, use this shape:
 
 ```csharp
-public static IDisposable OverrideClock(Func<DateTimeOffset> clock)
+public static class FeatureFlags
 {
-    var previous = s_clock.Value;
-    s_clock.Value = clock;
-    return new Scope(() => s_clock.Value = previous);
+    private static readonly AsyncLocal<Func<string, string?>?> s_environment = new();
+
+    public static bool IsEnabled(string name)
+    {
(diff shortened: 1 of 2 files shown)


Evangelink (commented):
/evaluate

YuliiaKovalova (approved):

PR #1079, opened 2026-08-27
Separate dormancy evidence and improve affected evals

Evangelink:
## Summary

Correctly dormant skilled arms are intentionally behaviorally equivalent to baseline, so their paired preference is a tie or judge noise by construction. Those scenarios were still voting in the exact sign test and could turn a valid skill improvement into a random preference loss.

This change introduces adapter schema version 4:

- `expect_activation: false` scenarios remain fully retained in raw trials, judge evidence, retry/error and pairing integrity, `comparisonTrialEvidence`, and `completionTransitions`, but are excluded from `signTest`, `netWin`, and `stimulusVoteCount`.
- `excludedScenarioEvidence` and per-scenario preference eligibility make the exclusion explicit rather than silently dropping evidence.
- `activationContract` evaluates explicit dormancy against isolated target-skill activation. Unexpected activation blocks a pass and is reported distinctly; plugin activity remains diagnostic because it cannot identify the emitting skill.
- Eval-quality power checks count preference-eligible stimuli only and report dormancy-contract counts separately.
- PR reporting, dashboard data/rendering, authoring skills, and investigation guidance explain the preference/activation split.

## Evaluation-informed skill fixes

The first complete schema-v4 run produced 48 model/skill results: 24 improved, 20 measured but not proven improved, and 4 activation-contract failures. There were no invalid, underpowered, or preference-regression results. The follow-up reads the exact losing trial evidence and addresses all 16 unique non-passing skills without adding repeated runs or weakening graders.

- **Activation boundaries:** `coverage-analysis` now defers single named method/class/file CRAP work to `crap-score`; `generate-testability-wrappers` stays dormant when code already consumes an interface or built-in abstraction.
- **Analysis quality:** `assertion-quality`, `detect-static-dependencies`, `test-anti-patterns`, `test-gap-analysis`, and `test-smell-detection` now enforce precise matcher semantics, reconciled occurrence counts, coherent severity/counting, complete branch/exception inventories, canonical mutation verification, and clean-framework stop conditions.
- **Execution and testability:** `migrate-static-to-wrapper`, `platform-detection`, `run-tests`, `scaffold-dotnet-test-project`, `testability-obstacle`, and `writing-mstest-tests` now preserve clock-read semantics, distinguish runner selection from command integration, use correct MTP/VSTest syntax, validate the exact requested tree, require executed seam tests, and use version-correct MSTest assertions/cancellation APIs.
- **Framework migrations:** `migrate-mstest-v3-to-v4`, `migrate-xunit-to-mstest`, and `migrate-xunit-to-xunit-v3` now resolve package versions, preserve VSTest/MTP intentionally, verify no-op migrations, retain parallelization and base-class behavior, and avoid unrelated file or version changes.
- **Eval strengthening:** existing xUnit migration stimuli now verify effective runner configuration, inherited base calls, byte-for-byte no-op behavior, and actual test execution. No new stimuli or runs were added.

Internal rubber-duck and code-review passes identified and drove corrections for MTP-to-MSTest translation, transitive Test.Sdk handling, no-op package-only detection, alternate MSBuild property locations, parameter-name-independent graders, `TestContext.CancellationToken` availability, and mechanical-versus-behavioral clock guidance.

## Compatibility and migration

- Existing eval YAML needs no migration; `expect_activation: false` automatically opts into activation-contract-only evidence.
- In schema version 4, legacy top-level aliases (`wins`, `ties`, `losses`, `winRate`, `stimulusVoteCount`, and `trialCount`) mean preference-eligible evidence. Consumers needing an all-stimulus view should combine `scenarioEvidence` with `excludedScenarioEvidence`, or read `scenarios[]` / `comparisonTrialEvidence`.
- Historical schema-version-3 dashboard data remains readable with its original all-stimulus semantics.
- Current non-agent dormancy evals retain at least five preference-eligible stimuli, so this creates no new underpowered allowlist debt.
- Skill/eval follow-up changes use the existing Vally schema and existing stimulus set; they require no consumer migration.

## Related issue

N/A

## Validation

- `node --test eng\vally-adapter\adapt.test.mjs eng\vally-adapter\consolidate.test.mjs eng\vally-adapter\retry-executor-timeouts.test.mjs eng\dashboard\dashboard-freshness.test.js` — 72 passed
- `python eng\eval-quality\selftest_eval_quality.py` — 27 passed
- `python eng\eval-quality\check_eval_quality.py` — 98 specs checked, no errors
- `dotnet run --project eng\skill-validator\src\SkillValidator.csproj -p:CopilotSkipCliDownload=true -- check --plugin .\plugins\dotnet-test` — passed
- `dotnet run --project eng\skill-validator\src\SkillValidator.csproj -p:CopilotSkipCliDownload=true -- check --plugin .\plugins\dotnet-test-migration` — passed
- Focused effective-MSBuild-property and no-op file-integrity grader checks passed
- `git diff --check` passed
- Two rubber-duck/code-review rounds completed; all material findings were incorporated

## Checklist

- [x] I updated CODEOWNERS when adding or moving owned content.
- [x] I updated `eng/known-domains.txt` for any new external domains referenced by skill content.


eng/dashboard/generate-benchmark-data.ps1
```diff
@@ -237,6 +237,11 @@ foreach ($verdict in $results.verdicts) {
         $activationScenarios.Add([ordered]@{
             scenarioName = $scenario.scenarioName
             expectation  = if ($isReferenceSkill) { "reference" } elseif ($expectActivation) { "active" } else { "dormant" }
+            preferenceGateEligible = if ($scenario.PSObject.Properties['preferenceGateEligible']) {
+                $scenario.preferenceGateEligible -ne $false
+            } else {
+                $expectActivation
+            }
```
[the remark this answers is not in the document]
Evangelink: Good catch — fixed in 89380de34. Scenarios without schema-v4 eligibility metadata now default to eligible, preserving schema-v3 dashboard semantics while explicit v4 exclusions remain excluded.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

Evangelink (commented):
/evaluate f8329e8293a8fc62c41582d23fd943ddc1c38555

eng/vally-adapter/adapt.mjs
```diff
@@ -260,7 +261,7 @@ function readNonActivationStimuli(evalFile, repoRoot) {
     }
     return t;
   };
-  const isFalsey = (v) => /^(false|no|off)\b/i.test(v.trim());
+  const isFalsey = (v) => /^(false|no|off|n)\b/i.test(v.trim());
```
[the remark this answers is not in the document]
Evangelink: Accepted and fixed in 0075511d7. Dormancy now requires a complete YAML false token, optionally followed by whitespace or a comment; prefix values such as `off-target` remain preference eligible, with regression coverage for both cases.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

eng/dashboard/dashboard.js
```diff
@@ -357,12 +361,16 @@
     const count = Number.isFinite(gate.stimulusVoteCount)
       ? gate.stimulusVoteCount
       : (gate.wins || 0) + (gate.ties || 0) + (gate.losses || 0);
+    const excluded = Number.isFinite(gate.excludedStimulusCount)
+      ? gate.excludedStimulusCount
+      : 0;
     return `
-      <div><strong>${count}</strong> stimulus vote${count === 1 ? '' : 's'} &middot;
+      <div><strong>${count}</strong> preference-eligible stimulus vote${count === 1 ? '' : 's'} &middot;
         <strong>${gate.wins || 0}W/${gate.ties || 0}T/${gate.losses || 0}L</strong></div>
```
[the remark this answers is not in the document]
Evangelink: Accepted and fixed in 44b59cce7. The converter now emits the exclusion marker only for schema v4+, and the dashboard uses its presence to label legacy all-stimulus votes separately from preference-eligible votes.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

Evangelink (commented):
/evaluate 44b59cce783724b218a3178e4a3ee8c2bece4ab1

eng/vally-adapter/adapt.mjs
```diff
@@ -260,7 +261,8 @@ function readNonActivationStimuli(evalFile, repoRoot) {
     }
     return t;
   };
-  const isFalsey = (v) => /^(false|no|off)\b/i.test(v.trim());
+  const isFalsey = (v) =>
+    /^(?:false|no|off|n)(?:\s*(?:#.*)?)$/i.test(v.trim());
```
AbhitejJohn: This is blocking because schema v4 makes this zero-dependency YAML scan part of the pass/fail gate. It must classify `expect_activation` exactly as PyYAML does in `check_eval_quality.py`; otherwise the pre-run power gate can approve five preference stimuli while the adapter later counts four and reports every run as underpowered. PyYAML does not accept single-letter `n` or mixed-case forms such as `fAlse`.

```suggestion
  const isFalsey = (v) =>
    /^(?:false|False|FALSE|no|No|NO|off|Off|OFF)\s*(?:#.*)?$/.test(v.trim());
```

Please also harden this parser before it becomes load-bearing: a valid zero-indent sequence (`stimuli:` followed by `- name:` at column 0) currently exits at `if (ind === 0)` before reading any item, and flow mappings such as `- {name: A, expect_activation: false}` are not read. Both cases silently return dormancy stimuli to the preference gate and make the activation contract vacuous. Add focused parity tests against these forms.

eng/vally-adapter/adapt.mjs
```diff
@@ -1068,6 +1101,35 @@ function comparisonToVerdict(report, identity, roles, nonActivationStims) {
...
+    .map((scenario) => ({
+      scenarioName: scenario.scenarioName,
+      expected: "dormant",
+      observed: scenario.skillActivationIsolated?.activated ? "activated" : "dormant",
+      satisfied: !scenario.skillActivationIsolated?.activated,
+    }));
+  const activationContractFailures = activationContractScenarios.filter(
+    (scenario) => !scenario.satisfied,
+  );
+  const activationContract = {
```
AbhitejJohn, line 1173: Please fail loudly when a dormancy annotation matches no observed stimulus. A rename, typo, or parser gap currently removes both its preference exclusion and its pass-blocking activation contract with no evidence in `results.json`.

```suggestion
  const observedStimulusNames = new Set(stimulusNames);
  const unmatchedDormancyStimuli = [...nonActivation]
    .filter((name) => !observedStimulusNames.has(name))
    .sort();
  const activationContract = {
```

Add `unmatchedDormancyStimuli` to this object and the invalid-verdict shape, and emit an adapter warning when it is non-empty. This keeps the verdict shape stable and makes an unevaluated contract visible without changing the current pass rule.

eng/vally-adapter/consolidate.mjs
```diff
@@ -236,15 +243,18 @@ function scenarioTable(verdict, weakOnly = false) {
...
   for (const scenario of scenarios) {
     const { netWin, wins, ties, losses } = scenarioStats(scenario);
     const icon = netWin > 0 ? "▲" : netWin < 0 ? "▼" : "=";
     const magnitude = typeof scenario.meanScore === "number" ? scenario.meanScore : 0;
+    const eligibility = scenario.preferenceGateEligible === false
+      ? "Excluded (activation contract)"
+      : "Eligible";
     rows.push(
-      `| ${icon} ${td(html(scenario.scenarioName))} | ${pct(netWin)} | ${pct(magnitude)} | ${wins}/${ties}/${losses} |`,
+      `| ${icon} ${td(html(scenario.scenarioName))} | ${eligibility} | ${pct(netWin)} | ${pct(magnitude)} | ${wins}/${ties}/${losses} |`,
```
AbhitejJohn, line 257: The table now marks dormancy rows as excluded, but `representativeEvidence()` below still chooses the first weak/tied scenario without checking that flag. A correctly dormant scenario ties by construction, so it can become the illustrative judge evidence for a passing verdict even though the gate explicitly ignored it. I reproduced a row showing `5W/0T/0L; 1 dormancy excluded` next to evidence that both arms produced the same refusal.

Please order candidates so preference-eligible scenarios come first. An excluded dormancy scenario should be selected only when `skillActivationIsolated.activated === true` (a real contract violation), or as a final fallback when no eligible scenario has evidence. Add a rendering test that puts an alphabetically first dormant tie before an eligible tied scenario and asserts that only the eligible evidence is quoted.

tests/dotnet-test-migration/migrate-xunit-to-xunit-v3/eval.yaml
```diff
@@ -196,6 +208,15 @@ stimuli:
       - type: output-matches
         config:
           pattern: UseMicrosoftTestingPlatformRunner
+      - type: run-command
+        config:
+          command: dotnet msbuild TestProject.csproj -getProperty:UseMicrosoftTestingPlatformRunner
+          expected_exit_code: 0
+          stdout_contains: 'true'
```
AbhitejJohn: This grader rejects a correct migration that writes `<UseMicrosoftTestingPlatformRunner>True</...>`. `dotnet msbuild -getProperty` echoes the value verbatim, while `stdout_contains` is case-sensitive and also accepts bad substrings such as `nottrue`.

```suggestion
          command: sh -c "dotnet msbuild TestProject.csproj -getProperty:UseMicrosoftTestingPlatformRunner | grep -qix true"
          expected_exit_code: 0
```

This command accepts valid Boolean casing and requires the whole output line to be `true`.

AbhitejJohn (changes requested):
Blocking correctness fixes are required before merge. Schema v4 makes dormancy parsing and reporting gate-critical, but the parser still has silent fail-open forms and one report path quotes evidence that the gate excludes. The new xUnit migration grader also rejects valid MSBuild Boolean casing. Each inline comment gives a concrete failure case and fix.

AbhitejJohn:
There is one additional **reporting regression** in the per-shard job summary at `.github/workflows/evaluation-run.yml:926-985`.

Schema v4 changes the result semantics, but this summary still uses the schema-v3 presentation:

- The header says every distinct stimulus gives one gate vote. Dormancy stimuli are now retained but excluded from the preference gate.
- The icon/label logic uses `STATE` and `PREFERENCE_REGRESSED` only. An activation-contract failure has `STATE=VALID_NO_CHANGE` and `PREFERENCE_REGRESSED=false`, so the summary renders it as `➖ Not proven improved` instead of the distinct `⛔ Activation contract failed` result shown by the adapter, consolidated PR report, and dashboard.
- The scenario table does not show preference eligibility, so a dormancy row can display a win/tie/loss without telling the reader that it did not vote in the preference gate.

The verdict and merge-gate calculation remain correct; this regression affects the shard summary's observability and can send an investigator toward preference quality instead of the real routing violation.

Please update this summary to use the v4 terminology, render `stateReason.code == "activation_contract_failed"` as a blocking `⛔` result, and add a preference-eligibility column consistent with `consolidate.mjs`.

eng/vally-adapter/adapt.mjs
Evangelink: Fixed in 47bbc32ec. The scanner now matches only PyYAML’s accepted false spellings, requires YAML comment separation, and handles both zero-indent sequence items and flow-mapping items with focused parity tests.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

tests/dotnet-test-migration/migrate-xunit-to-xunit-v3/eval.yaml
Evangelink: Fixed in 47bbc32ec. The grader now pipes the effective property through `grep -qix true`, accepting valid MSBuild Boolean casing while rejecting substring matches.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

eng/vally-adapter/consolidate.mjs
Evangelink: Fixed in 47bbc32ec. Representative evidence now prioritizes preference-eligible scenarios, then actual dormancy-contract violations, with correctly dormant excluded evidence used only as a final fallback; the regression test covers an earlier dormant tie.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

eng/vally-adapter/adapt.mjs
Evangelink: Fixed in 47bbc32ec. Unmatched annotations are now retained as `activationContract.unmatchedDormancyStimuli` in valid and invalid verdict shapes, emitted as adapter warnings, and surfaced in consolidated warnings without changing the current pass rule.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

Evangelink:
@AbhitejJohn all comments have been addressed, there are still 11 evals that needs tweaking but I'd like to merge this one as-is and then look at how to improve the rest. Please approve or suggest other changes!

AbhitejJohn (approved):

PR #1090, opened 2026-08-30
Document efficient repository cloning

jamesmontemagno:
## Summary

- recommend a blobless, single-branch clone for contributors who need `main` history
- document a shallow, single-branch option for CI and read-only use
- explain the history tradeoff of the shallow option

This helps users avoid downloading disconnected generated dashboard branch history without changing repository behavior or expanding the focused Actions change in #1089.

## Validation

- `npx --yes markdownlint-cli2 CONTRIBUTING.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,27 @@ This repository contains shared building blocks for coding agents:
 
 Because these artifacts can affect many users and workflows, we prioritize correctness, clarity, and long term maintainability over speed.
 
+## Cloning efficiently
+
+This repository has generated dashboard branches with substantial history. For a
+contributor clone that retains the `main` branch history, use:
+
+```bash
+git clone --filter=blob:none --single-branch https://github.com/dotnet/skills.git
+```
+
+`--single-branch` excludes the disconnected generated branches, while
+`--filter=blob:none` downloads file contents on demand.
+
+For CI or read-only use that needs only the latest `main` snapshot, use:
+
+```bash
+git clone --depth=1 --single-branch https://github.com/dotnet/skills.git
+```
+
+The shallow option omits older `main` history, so it is not recommended when that
+history is needed for development.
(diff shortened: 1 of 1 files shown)


AbhitejJohn:
I'm wondering if we'd want to move these dashboard branches off to another repo. I've been considering moving some of the test code over to another repo as well given how harnesses pull down the repo to load skills from plugins today. 
/cc: @JanKrivanek for thoughts as well.

jamesmontemagno:
@AbhitejJohn The review feedback is now fully addressed in commit `df2fb6f7a69f32344381930f362fe507bd5167ad`: both clone examples explicitly select `main`, and the blobless-clone wording now clarifies that the initial checkout still downloads required files. `markdownlint-cli2 CONTRIBUTING.md` passes.

I confirmed my account has triage/pull access but not push access to `dotnet/skills`. A maintainer can promote the updated branch with:

```bash
git fetch https://github.com/jamesmontemagno/skills.git motz-bookish-system
git push origin FETCH_HEAD:refs/heads/motz-bookish-system
```

GitHub cannot change an existing PR's head repository, so after the upstream branch exists we should open a replacement PR from `dotnet:motz-bookish-system`, then close this fork PR. I am leaving this open until the replacement exists.

JanKrivanek:
> I'm wondering if we'd want to move these dashboard branches off to another repo. I've been considering moving some of the test code over to another repo as well given how harnesses pull down the repo to load skills from plugins today. /cc: [@JanKrivanek](https://github.com/JanKrivanek) for thoughts as well.

Btw. moving data to different repo will mean that action will need to have write token for that other repo. Though - not a new thing - as we already moved out the session logs.

Would we move out the tests as well - then data can again live on a branch and above would not be a concerns. The contribution experience would suffer a bit + gating might get cumbersome.

All that being said - moving just the data branches might be best as of now

PR #1098, opened 2026-08-31
Cover xUnit-to-MSTest Parallelize eval pattern

Evangelink:
## Summary

The existing eval covered xUnit parallelization, but the coverage analyzer could not recognize the bare `Parallelize` code pattern. This PR adds deterministic evidence for that pattern and strengthens the migration skill after Luna produced valid but not-proven results.

The first Luna run exposed skill-content issues, not activation, reliability, or power failures:

> "Response B ... never explicitly confirms the [ClassInitialize]/[ClassCleanup] pattern ... which reduces trust in its reported outcome."

> "Response B ... stumbled with a stray copy of files into the skill directory that broke the build ... before recovering."

> "Response B ... never shows the final AnnotatedTests.cs source, leaving all attribute conversions unverified."

The initial fix moved the data-driven case from loss to tie and the skip/trait/timeout case from loss to win. A second fix made class-fixture lifetime and generic typed-return semantics explicit. The following run improved Luna from 7W/3T/3L to 6W/6T/1L, with `p=0.0625`; its sole remaining loss was evidence clarity rather than implementation correctness:

> "A's timeline includes an explicit readback ... showing TestContext property, TestContext.WriteLine ... B ... provides slightly less concrete evidence of the exact converted result."

The final evidence fix produced a credible Luna improvement: **9W/3T/1L, `p=0.0107`, net +61.5%**, with complete isolated and plugin activation, no comparison errors, and a passing activation contract.

The skill now:

- maps `IClassFixture<T>` explicitly to a static field with `[ClassInitialize]` and `[ClassCleanup]`, never per-test lifecycle hooks;
- preserves assigned typed return values with generic MSTest type-assertion overloads;
- maps `ITestOutputHelper` and `_output.WriteLine` explicitly to `TestContext` and `TestContext.WriteLine`, and requires those exact APIs in completion evidence;
- applies assembly-level parallelization only when concurrency is observable in the current project, while still translating explicit settings;
- treats loaded skill and reference directories as read-only guidance rather than project workspaces;
- requires concise, concrete evidence for high-risk mappings and final test counts.

The earlier parallelization loss was not used to change the skill: its baseline escaped the evaluation workspace and modified the source fixture, so the treatment arm started from an already-migrated MSTest project and correctly followed the no-op rule.

## Related issue

N/A

## Validation

- `python eng/eval-quality/check_eval_quality.py` - passed with existing repository warnings only.
- `eng/skill-coverage/Measure-SkillCoverage.ps1 -PluginName dotnet-test-migration -SkillName migrate-xunit-to-mstest -Format Json` - reports 100% coverage with no uncovered points.
- `git diff --check` - passed.
- Repository skill-check, lint, coverage, and validation workflows pass on `7082c3176`.
- Official evaluation on `7082c3176` passes for both models; Luna is `VALID_PASS` at 9W/3T/1L, `p=0.0107`, net +61.5%.
- Local `skill-validator check` remained blocked during restore by a repeated TLS handshake failure downloading `@github/copilot-win32-x64` from npm; the repository skill-check passed.

## Checklist

- [ ] I searched existing issues and pull requests to avoid duplicates.

### DIFF
--- a/plugins/dotnet-test-migration/skills/migrate-xunit-to-mstest/SKILL.md
+++ b/plugins/dotnet-test-migration/skills/migrate-xunit-to-mstest/SKILL.md
@@ -128,7 +128,7 @@ Apply the mechanical and semantic rewrites in one edit pass when the inventory m
 - Keep constructor setup and `IDisposable`/`IAsyncDisposable` when valid. Map `IAsyncLifetime` to `[TestInitialize]`/`[TestCleanup]`.
 - `IClassFixture<T>` means one fixture instance per test class, shared by all methods in that class. Map it to a static `T` field created once by a static `[ClassInitialize]` method that accepts `TestContext`, and dispose it once from static `[ClassCleanup]`. Never use `[TestInitialize]`/`[TestCleanup]` for this mapping because that creates one fixture per test method.
 - For `ICollectionFixture<T>`, preserve both sharing and serialization. Prefer a static `Lazy<T>` helper used by each member class; add `[DoNotParallelize]` only when the source collection disabled parallelization. Use assembly initialization only when the fixture is genuinely assembly-wide.
-- Replace `ITestOutputHelper` with injected or property-based MSTest `TestContext`.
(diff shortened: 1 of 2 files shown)


Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

Evangelink (commented):
/evaluate

JanKrivanek (approved):

PR #1111, opened 2026-09-03
Add Ignite UI Blazor skill

kotlarmilos:
## Summary

This moves the changes from #994 to a branch in `dotnet/skills` so the evaluation workflows can run. It adds guidance for integrating Ignite UI components into Blazor applications and an evaluation for split Blazor Web Apps.

## Validation

The evaluation checks package references, service registration in the server and client projects, `_Imports.razor` usage, asset linking, and file-specific setup instructions.


plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Pass `typeof(Igb<Name>Module)` values to eagerly pre-load a specific set instead:
+
+```csharp
+builder.Services.AddIgniteUIBlazor(
+    typeof(IgbInputModule), typeof(IgbComboModule), typeof(IgbDialogModule));
+```
+
+Module names always follow `Igb{ComponentName}Module`. In `IgniteUI.Blazor.Lite` a component registers its own module on first render, so the explicit list trims the initial payload rather than gating rendering.
```
AbhitejJohn: Thanks for documenting this path. Could we correct the payload guidance here? The shipped implementation leaves `ModulesToLoad` empty for `AddIgniteUIBlazor()` and lets components request modules on first render. Passing `Igb...Module` types preloads those modules during startup, so it increases the initial transfer rather than trimming it. The fifth eval stimulus and rubric repeat the same inversion, so could we update both to describe this as a startup warm-up/first-render latency trade-off?

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+Host page is `wwwroot/index.html` (WASM/MAUI), `Pages/_Host.cshtml` (Server), or `Components/App.razor` (Web App).
+
+```html
+<link href="_content/IgniteUI.Blazor/themes/light/bootstrap.css" rel="stylesheet" />
+...
+<script src="_content/IgniteUI.Blazor/app.bundle.js"></script>
+<script src="_framework/blazor.web.js"></script>   <!-- or blazor.server.js / blazor.webassembly.js / blazor.webview.js -->
+```
+
+Both tags are required: without the stylesheet components render unstyled, without `app.bundle.js` they do not render at all. `app.bundle.js` must come **before** the Blazor framework script.
```
AbhitejJohn: Could we scope this requirement to `IgniteUI.Blazor.Lite` and add the GridLite-only contract explicitly? The GridLite package README says no service registration or manual JavaScript bundle is required, and the package has no `app.bundle.js`. As written, a GridLite-only user can follow this unconditional step and add invalid Lite setup. A short GridLite branch saying "no `AddIgniteUIBlazor()`, no `app.bundle.js`; add the namespace and GridLite stylesheet only" would also align the skill with stimulus 4.

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
...
+      I only need a data grid out of Ignite UI — none of the other components — so I added just the IgniteUI.Blazor.GridLite package to my Blazor WebAssembly app and put the grid component on a page.
+
+      I got a grid that was completely unstyled. My index.html currently has no Ignite UI tags in it at all.
+
+      Give me the exact package, startup and index.html changes for a grid-only setup, and don't have me pull in things I don't need.
+    graders:
+      - type: output-contains
+        config:
+          substring: IgniteUI.Blazor.GridLite
+      - type: output-not-contains
```
AbhitejJohn: Could we replace these bare negative substring graders with checks for prescriptive usage, or leave this distinction to the rubric? A correct answer such as "do not call `AddIgniteUIBlazor()` and do not add `app.bundle.js` for GridLite" fails both hard graders. The same negation-blind issue appears at lines 169-171, and this repo already documents and fixes this failure mode in `tests/dotnet-test/test-anti-patterns/eval.yaml`. Since the skilled arm is more likely to know and explicitly rule out these names, the current checks can bias the comparison against it.

tests/dotnet-blazor/use-igniteui-blazor/eval.yaml
```diff
@@ -0,0 +1,200 @@
+name: use-igniteui-blazor
+description: Evaluates the dotnet-blazor/use-igniteui-blazor skill
+type: capability
+defaults:
+  timeout: 10m
+  runs: 5
```
AbhitejJohn, line 6: Could we add at least one more distinct preference stimulus before treating this eval as sufficient evidence? `check_eval_quality.py` reports exactly five preference stimuli, so one tie leaves only four discordant votes and makes a passing verdict impossible. `runs: 5` measures reliability for the same five tasks; it does not add statistical power. A separate `expect_activation: false` case for a plain Blazor request with no Ignite UI would also verify the routing boundary in the description, although that dormancy case would not count toward the preference floor.

AbhitejJohn (commented):
Thanks for putting this together. I completed a multi-model review and also checked the evaluation design. The structural checks pass, but there is no skilled-versus-baseline result for this head yet, and the current five-stimulus design has no tolerance for a tie. I left focused comments on the new issues that are not already covered by the existing review threads.

kdinev:
@kotlarmilos I created one more PR about adding more stimuli, as well as a small adjustment to the skill the tests uncovered. #1151 
@AbhitejJohn I believe all review comments are addressed now. I cannot resolve the conversations though, since this is not my PR anymore.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+```html
+<link href="_content/IgniteUI.Blazor.GridLite/css/themes/light/bootstrap.css" rel="stylesheet" />
+```
+
+## 5. Render mode (Blazor Web App only)
+
+Ignite UI components need an interactive render mode; static SSR renders nothing usable.
+
+```razor
+@rendermode InteractiveServer   @* or InteractiveWebAssembly / InteractiveAuto *@
```
AbhitejJohn, line 117: Could we move the alternatives out of this directive line? Razor requires `@rendermode` to end at the line break, so copying this snippet verbatim fails with `RZ1017: Unexpected literal following the 'rendermode' directive`. I reproduced it in a new Interactive Server Blazor Web App. `@rendermode InteractiveServer` on its own line builds; the alternatives can stay in the surrounding prose.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,120 @@
...
+  JavaScript interop (see use-js-interop), authentication (see configure-auth),
+  prerendering (see support-prerendering), or layout/component design questions that need
+  no Ignite UI setup.
+---
+
+# Application Setup & Component Registration
+
+## 1. NuGet package
+
+Before adding packages, inspect the target projects' existing package references. If `IgniteUI.Blazor` or `IgniteUI.Blazor.Trial` is already referenced, keep that package strategy and do not add Lite or GridLite. Only switch package families when the user explicitly asks, replacing conflicting references rather than keeping both.
```
AbhitejJohn, line 24: Nice improvement on checking existing package references. Could we also check the target framework before adding these packages and state the current minimum? The shipped Lite 0.1.1 and GridLite 0.9.1 packages contain compile assemblies only for `net8.0`, `net9.0`, and `net10.0`. I tested `net7.0`: restore succeeds without a compatibility warning because the package has applicable build assets, but `using IgniteUI.Blazor.Controls` then fails with `CS0246` because there is no compatible compile asset. A short .NET 8+ prerequisite would prevent that misleading path for older existing Blazor apps.

AbhitejJohn (approved):
Thanks for the updates. The earlier package, GridLite, module-loading, eval-grader, and coverage concerns are addressed, and the latest evaluation is strong across both tested models.

I'm approving this. I left two focused follow-up notes: move the render-mode alternatives out of the `@rendermode` directive line because the exact snippet triggers `RZ1017`, and state the current .NET 8–10 package requirement so older Blazor projects get a clear prerequisite.

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
kdinev: @AbhitejJohn Addressed in #1176

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
kdinev: @AbhitejJohn Addressed in #1176

kotlarmilos (commented):
LGTM!

plugins/dotnet-blazor/skills/use-igniteui-blazor/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+name: use-igniteui-blazor
+description: >
+  Add, configure, or review Ignite UI for Blazor Lite component support in Blazor applications.
+  USE FOR: installing IgniteUI.Blazor.Lite or IgniteUI.Blazor.GridLite,
+  registering AddIgniteUIBlazor() in Blazor Server, WASM, Hybrid, or split
+  Blazor Web App projects, adding @using IgniteUI.Blazor.Controls, wiring the
+  theme stylesheet, picking the right host page, locating the GridLite
+  stylesheet path, explaining single-project vs split Server/Client Web App
+  setup differences, and checking where an interactive render mode is needed
+  for Ignite UI components to work.
```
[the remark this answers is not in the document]
kdinev: This suggestion would make the description exceed the maximum limit. I don't think this update is needed.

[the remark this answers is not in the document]
AbhitejJohn: Agreed, it does seem like overkill.

AbhitejJohn (approved):
Approving. I've re-triggered a fresh eval on this just so we have a baseline for the final state. If we think we are all good @kotlarmilos , you can feel free to merge. Thanks for the contribution @kdinev .

PR #1140, opened 2026-09-08
Add NUnit to MSTest migration skill

Evangelink:
## Summary

Adds an NUnit 3/4 to MSTest v4 migration skill so test projects can change frameworks without silently changing their target framework, VSTest/MTP runner, fixture lifetime, data cases, or concurrency behavior.

The skill classifies all 48 attributes in NUnit's official attribute index and provides detailed mappings for assertions, lifecycle, data sources, theories, pairwise/combinatorial generation, cooperative cancellation, STA execution, platform and culture conditions, metadata, and behaviors with no exact MSTest equivalent. It is routed through the test-migration agent and backed by a 14-scenario capability eval with focused fixtures for the highest-risk semantic conversions.

Notable decisions:

- NUnit's default single fixture instance is treated as a migration hazard because MSTest creates a class instance per test.
- VSTest cooperative cancellation uses `TestContext.CancellationToken`; retaining NUnit's injected method parameter fails under the VSTest adapter.
- NUnit retry counts include the initial attempt, so `[Retry(n)]` maps to MSTest `[Retry(n - 1)]`.
- Pairwise cases are preserved exactly rather than expanded to a Cartesian product.
- Thread identity, explicit execution, and other behaviors without an equivalent MSTest attribute are reported as deliberate rewrites or manual follow-up instead of approximated silently.

## Related issue

N/A

## Validation

- `python eng/eval-quality/check_eval_quality.py` - completed with no errors.
- `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin ./plugins/dotnet-test-migration` - all plugin, skill, and agent checks passed using the installed stable .NET SDK.
- Ran `dotnet test` for every supported eval fixture - all expected healthy fixtures passed, including the NUnit and MSTest semantic probes.
- Verified every attribute from NUnit's official 48-attribute index appears in the migration reference.
- `git diff --cached --check` - no whitespace errors.

## Checklist

- [x] Existing plugin-wide CODEOWNERS entries cover the new skill and test directory; no update was required.
- [x] No new external domains were referenced, so `eng/known-domains.txt` did not require an update.

### DIFF
--- a/plugins/dotnet-test-migration/.claude-plugin/plugin.json
+++ b/plugins/dotnet-test-migration/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dotnet-test-migration",
   "version": "0.1.8",
-  "description": "Skills and an orchestrator agent for migrating .NET test frameworks and platforms: MSTest and xUnit version upgrades, xUnit-to-MSTest conversion, and VSTest to Microsoft.Testing.Platform.",
+  "description": "Skills and an orchestrator agent for migrating .NET test frameworks and platforms: MSTest and xUnit version upgrades, xUnit/NUnit-to-MSTest conversion, and VSTest to Microsoft.Testing.Platform.",
   "skills": ["./skills/"],
   "agents": [
     "./agents/test-migration.agent.md"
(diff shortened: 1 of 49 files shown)


Evangelink (commented):
/evaluate

JanKrivanek (approved):

Evangelink (commented):
/evaluate

Issue #1160, opened 2026-09-13
[evaluation] Diagnose GPT-5.3 Codex stopping after skill activation


Evangelink:
## Problem

Current GPT-5.3 Codex evaluation traces frequently load the requested skill and then immediately return a final response without inspecting, editing, or validating the workspace.

In scheduled run https://github.com/dotnet/skills/actions/runs/34680113000, 12 of 18 isolated `code-testing-agent` trajectories performed exactly one tool call—the `skill` invocation—then stopped. The equivalent GPT-5.6 Luna run had 0 of 18 such trajectories. GPT-5.3 averaged 11.9 tool calls in the baseline arm but only 3.9 in the skilled arm for this eval.

A representative migration trace loaded guidance that explicitly said “Skill activation is not completion,” “inspect the staged files,” and “do not ask the user for a discoverable path,” then emitted `phase: final_answer` asking the user to provide files without making another tool call.

## Why this may be infrastructure-level

- The runtime injects the loaded `SKILL.md` as a new user-role `<skill-context>` message.
- A second model request is made, so this is not simply a missing host loop iteration.
- OpenAI's Codex guidance requires preservation of assistant `phase`, reasoning items, and tool-call/result ordering; losing those can materially degrade continuation.
- The evaluation matrix selects the model but does not pin reasoning effort or context tier.
- The evidence was produced by Copilot runtime 1.0.80, while newer runtime versions exist.

## Proposed investigation

Run the same implementation-heavy stimuli under a controlled factorial:

1. baseline, automatic activation, explicit skill invocation, and the same skill body pasted directly into the prompt;
2. current pinned runtime versus the latest supported runtime;
3. low, medium, and high reasoning effort;
4. isolated skill versus full plugin;
5. primary and independent secondary judges, while prioritizing deterministic completion graders.

Add trace telemetry for:

- skill selected and body injected;
- first non-skill action after activation;
- activation-only terminal responses;
- finish reason and assistant phase;
- preserved reasoning/tool-call items;
- workspace reads, writes, validation commands, and deterministic completion.

## Acceptance criteria

- The harness can distinguish routing failure, post-activation continuation failure, tool failure, and content-quality failure.
- Implementation prompts that stop after only loading a skill are surfaced explicitly rather than appearing merely as low-quality output.
- Model, runtime, reasoning effort, context tier, and tool-schema identity are recorded with every result.
[...]

PR #1163, opened 2026-09-13
Surface activation-only evaluation failures

Evangelink:
## Summary

GPT-5.3 Codex evaluation traces can load the requested skill and then terminate without making any subsequent workspace or execution tool call. The existing reports classify those runs only through their quality outcome, which makes a host/model continuation failure look like a skill-content failure.

This change adds diagnostic post-activation telemetry to each isolated and plugin scenario. It counts activated runs that continued with a non-skill tool, completed after skill calls only, failed after skill calls only, or could not be classified. Failed activation-only completions are surfaced in PR warnings, next-action guidance, and dashboard activation evidence.

The telemetry is intentionally diagnostic: it does not change the preference gate, and successful advice-only responses are not warned.

## Related issue

Relates to #1160

## Validation

- `node --test .\eng\vally-adapter\*.test.mjs .\eng\dashboard\*.test.js` - 82 tests passed.
- `node --check` passed for the modified JavaScript modules.
- PowerShell parser validation passed for `eng/dashboard/generate-benchmark-data.ps1`.
- `git diff --check` passed.
- Replayed the telemetry helper against scheduled GPT-5.3 artifacts: 17 activated `code-testing-agent` runs, 12 failed activation-only completions, and 5 continued runs.

## Checklist

- [x] I updated CODEOWNERS when adding or moving owned content. No ownership changes were required.
- [x] I updated all marketplace manifests when plugin metadata changed. No plugin metadata changed.
- [x] I updated `eng/known-domains.txt` for any new external domains referenced by skill content. No skill content or external domains changed.


eng/vally-adapter/consolidate.mjs
```diff
@@ -377,6 +411,10 @@ function nextAction(verdict) {
   if (isPreferenceRegression(verdict)) {
     return "Inspect losing stimuli and fix skill behavior; this is not objective completion proof.";
   }
+  const postActivation = postActivationStats(verdict);
+  if (postActivation.hasFailures) {
```
JanKrivanek, line 413: When `isPreferenceRegression()` is true, the earlier return skips this check. The report can show activation-only failures but still tell the author to "fix skill behavior." Move this diagnostic before the preference-loss branch so users investigate the model/runtime first, without changing the verdict.

eng/vally-adapter/consolidate.mjs
```diff
@@ -206,6 +206,27 @@ function activationCell(verdict) {
   return `isolated ${stats.isolated}/${stats.total}${plugin}`;
 }
 
+function postActivationStats(verdict) {
+  const expected = (verdict.scenarios ?? []).filter(
+    (scenario) => scenario?.expectActivation !== false,
```
JanKrivanek, line 212: `expectActivation: false` means the target skill should stay dormant, not every skill in the plugin. Another plugin skill can load and stop, but this filter hides its failure from the PR warning while the dashboard still shows it. Count observed plugin failures regardless of the target skill's activation expectation.

JanKrivanek (approved):
Overall looks useful and done in the right way. Please consider the inline comments

eng/vally-adapter/consolidate.mjs
Evangelink: Fixed in `ef73d0235`. Isolated failures remain limited to expected target activation, but plugin-arm activation-only failures are now counted across every scenario because a sibling plugin skill may be the one that loaded and stopped.

eng/vally-adapter/adapt.mjs
```diff
@@ -419,6 +419,52 @@ function mean(nums) {
...
+    const skillToolCallCount =
+      metrics?.toolCallBreakdown?.skill ?? activationCount;
+    if (!Number.isFinite(toolCallCount) || !Number.isFinite(skillToolCallCount)) {
+      summary.unclassifiedRuns += 1;
+      continue;
+    }
+
+    if (toolCallCount > skillToolCallCount) {
+      summary.continuedRuns += 1;
+      continue;
```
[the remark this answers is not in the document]
Evangelink: Fixed in `ef73d0235`. Continuation now uses the ordered trajectory event stream and requires a non-skill tool call after `skill_activation`; pre-activation tool calls no longer count, and aggregate-only legacy records fail closed to unclassified unless they prove activation-only completion.

eng/vally-adapter/consolidate.mjs
Evangelink: Fixed in `ef73d0235`. Activation-only failure diagnosis now takes precedence over preference-regression guidance, while the verdict itself remains unchanged.

Evangelink (commented):
/evaluate

JanKrivanek (approved):

AbhitejJohn:
@Evangelink Thanks a bunch for the telemetry work here. I found two follow-up edge cases after replaying the new paths:

1. In `consolidate.mjs`, `nextAction()` returns for `preferenceRegressed` before it checks activation-only failures. A result can therefore show `Activation-only stop` but still tell the contributor to “fix skill behavior.” A passing result also returns at that check, which makes the later `actions.push("Inspect activation-only stops")` path unreachable. Could we add overlap tests and make the guidance explicit for preference-loss and passing states?

2. In `adapt.mjs`, `metrics?.toolCallBreakdown?.skill ?? activationCount` can create a false activation-only warning when a breakdown exists without a `skill` entry. I reproduced it with `skillActivationCount: 1`, `toolCallCount: 1`, and `toolCallBreakdown: { view: 1 }`; the helper classified the view-only tool record as a failed activation-only completion. Could we treat a missing `skill` entry as zero when the breakdown exists, or mark inconsistent metrics as unclassified, and add a focused test?

Neither changes the main result of the PR, but they would make the diagnostic fail safely and keep the requested next action aligned with the warning.

Issue #1127, opened 2026-09-05
[dotnet-msbuild] Codex drops binlog MCP because tools is an array instead of a per-tool settings object


grevgeny:
## Observed behavior

With `dotnet-msbuild` 0.1.9 installed and enabled, Codex CLI 0.153.4 rejects the bundled `binlog` MCP declaration. Skills load, but binlog tools are absent. The runtime logs:

```text
failed to parse plugin manifest MCP object server
server="binlog" error="invalid type: sequence, expected a map"
```

## Root cause

`plugins/dotnet-msbuild/.codex-plugin/plugin.json` includes:

```json
"mcpServers": {
  "binlog": {
    "type": "stdio",
    "command": "dotnet",
    "args": ["dnx", "Microsoft.AITools.BinlogMcp", "--yes", "--prerelease"],
    "tools": ["*"]
  }
}
```

In Codex, `tools` is a map of tool names to per-tool settings, not an allow-list. The configuration is rejected before the server starts. Omitting the optional field permits parsing and leaves `enabled_tools`/`disabled_tools` unset.

The installed manifest matches current upstream 0.1.9. The inline declaration is visible in [the manifest introduced by PR #1070](https://github.com/dotnet/skills/blob/be7b56028cf14cc5b9482b29959d7d96433bad05/plugins/dotnet-msbuild/.codex-plugin/plugin.json#L11). Codex documents per-tool settings separately from `enabled_tools` and `disabled_tools` in its [configuration reference](https://developers.openai.com/codex/config-reference).

## Minimal reproduction

These command-line overrides are transient; they do not edit the user's configuration or launch the MCP server:

```sh
codex -c 'mcp_servers.audit_binlog={command="dotnet",args=["dnx","Microsoft.AITools.BinlogMcp","--yes","--prerelease"],tools=["*"]}' mcp get audit_binlog --json
```

Observed exit code: **1**.

```text
Error: failed to load bootstrap configuration
Caused by:
    invalid type: sequence, expected a map
    in `mcp_servers.audit_binlog.tools`
```

Remove only the `tools` field:

```sh
codex -c 'mcp_servers.audit_binlog={command="dotnet",args=["dnx","Microsoft.AITools.BinlogMcp","--yes","--prerelease"]}' mcp get audit_binlog --json
```

Observed exit code: **0**; the returned server is enabled with the expected command/arguments and no tool allow-list or deny-list.

[...]

Issue #1006, opened 2026-08-13
Adopt Agent Plugins 1.0


aaronpowell:
### Why this matters

Agent Plugins 1.0 is an open, vendor-neutral standard for packaging reusable Agent Skills and MCP servers so they can be discovered and loaded consistently across compatible clients.

Adopting the standard would make this plugin easier to consume across GitHub Copilot, Codex, Cursor, VS Code, and other compatible clients without requiring client-specific repackaging or duplicated component structures.

Resources:

- [Agent Plugins](https://agent-plugins.org/)
- [Agent Plugins 1.0 specification](https://agent-plugins.org/specification)
- [GitHub Changelog: Agent Plugins 1.0 in VS Code, Copilot CLI, and the Copilot app](https://github.blog/changelog/2026-08-12-agent-plugins-1-0-in-vs-code-copilot-cli-and-the-copilot-app/)

### Requested changes

Please update this plugin to conform to the Agent Plugins 1.0 specification:

- Move `plugin.json` to the root of the plugin directory.
- Add the `$schema` field:

````json
{
    "$schema": "https://agent-plugins.org/schemas/1.0.0/plugin.schema.json"
  }
````

- Align the manifest fields with the Agent Plugins 1.0 schema.
- Place portable Agent Skills under the plugin-root `skills/` directory.
- Place MCP server configuration in the plugin-root `mcp.json` file, where applicable.
- Use the `com.github.copilot/` namespace for GitHub Copilot-specific components and metadata that are outside the portable Agent Plugins contract, such as agents, hooks, instructions, and extensions (including canvases).

The intended structure is similar to:

````text
plugin/
  ├── plugin.json
  ├── skills/
  │   └── <skill>/
  │       └── SKILL.md
  ├── mcp.json
  └── com.github.copilot/
      └── <GitHub Copilot-specific components>
````

### Compatibility considerations

If this repository also supports Claude Code's plugin format, please retain the Claude-specific manifest and layout as needed. Claude Code does not currently support the Agent Plugins 1.0 specification, so `.claude/plugin.json` may still be required for Claude Code compatibility.

However, the portable Agent Plugins manifest should live at the plugin root, and portable components should follow the standard locations described in the specification.

If the current plugin uses a non-standard location for any of the following, please consider moving or exposing them through the standard layout:

- Skills should be discoverable under the plugin-root `skills/` directory.
- MCP configuration should be available through the plugin-root `mcp.json` file.
[...]

PR #1164, opened 2026-09-13
Define and validate native Codex plugin support

### LINKED ISSUE #1161: [codex] Define and validate native Codex support for plugin agents
## Problem

The repository publishes `.codex-plugin/plugin.json` manifests containing GitHub-style `agents` entries and ships `.agent.md` files with `tools` and `handoffs`, but those constructs are not portable Agent Plugins 1.0 components and are not documented as native OpenAI Codex plugin agent packaging.

Current specifications distinguish these systems:

- Agent Plugins 1.0 defines portable Skills and MCP servers.
- OpenAI's plugin extension documents skills, MCP, apps/UI, and hooks, but not packaged custom agents or static handoff graphs.
- Native Codex custom agents use `.codex/agents/*.toml` with `developer_instructions` and model/tool configuration.
- GitHub Copilot custom agents use `.agent.md` and host-specific frontmatter.

The validator's agent tool-portability table currently covers Copilot, Claude Code, and Gemini CLI, but not native Codex. #1127 demonstrates that a Codex-specific manifest incompatibility can pass repository validation and silently remove functionality.

## Impact

Users can reasonably assume that installing the Codex marketplace plugin exposes the same agents and handoffs as Copilot, even though current documentation does not establish that behavior. Tool aliases and metadata may also be silently ignored or misinterpreted.

## Proposed work

- Document the supported Codex surface explicitly: portable skills/MCP versus host-specific agents.
- Decide whether native Codex agent definitions should be generated or maintained separately under a supported `.codex/agents/*.toml` installation path.
- Define how dynamic delegation replaces Copilot/IDE `handoffs` where static handoffs are unavailable.
- Add Codex-specific manifest/schema validation instead of assuming root, Claude, and Codex manifests can always be identical.
- Add an integration lane using a supported Codex CLI version to verify plugin installation, skill discovery, MCP parsing, and any claimed agent behavior.
- Extend tool portability validation only after mapping official Codex tool/config semantics.

## Acceptance criteria

- Repository documentation does not imply `.agent.md` portability to native Codex.
- Every feature claimed for Codex is covered by a native Codex installation/discovery test.
- Unsupported top-level manifest fields are rejected or moved under an appropriate host extension.
- Codex-specific tool and MCP configuration is validated against a supported Codex parser.
- The relationship among portable skills, Copilot custom agents, native Codex custom agents, and Agents SDK handoffs is documented.

Evangelink:
Fixes #1161

## Summary

- separate portable skills/MCP support from GitHub Copilot-specific `.agent.md` agents and Codex-native `.codex/agents/*.toml`
- remove ignored `agents` and `lspServers` fields from Codex compatibility manifests
- fix the #1127 binlog MCP regression by removing the invalid `tools: ["*"]` array from the Codex manifest
- validate Codex compatibility-manifest field shapes, required skill declarations, and per-tool MCP settings in `skill-validator check`
- add a pinned Codex 0.154.0 smoke workflow for marketplace installation, skill discovery, MCP discovery, and a real `binlog_overview` call
- constrain smoke cleanup to a dedicated child directory even when `CODEX_SMOKE_HOME` is overridden
- document portable components, native Codex agents, Copilot handoffs, and application-level Agents SDK handoffs precisely

Native Codex TOML agents are intentionally not generated: current Codex plugin installation does not place files in the documented project or user agent-discovery directories, and Copilot `tools`/`handoffs` do not have a mechanical Codex equivalent.

## Validation

- `dotnet test eng\skill-validator\SkillValidator.slnx --no-restore` - 662 passed
- `skill-validator check` across all 16 plugins - passed
- `pwsh -NoProfile -File eng\version\Test-Sync-PluginVersions.ps1` - 65 assertions passed
- touched-file `dotnet format --verify-no-changes` - passed
- `markdownlint-cli2` on changed Markdown - passed
- actionlint 1.7.7 on the new workflow - passed
- Codex CLI 0.154.0 installed all 15 marketplace plugins, discovered `dotnet-msbuild:binlog-failure-analysis` and the `binlog` MCP server, generated a binlog, and called `binlog_overview`
- a sentinel outside the dedicated smoke child survived cleanup with `CODEX_SMOKE_HOME` overridden
- version prediction advances all six affected plugins

## Follow-up decision

A future native Codex-agent offering still needs a documented distribution mechanism for `.codex/agents/*.toml` and an explicit semantic design for replacing Copilot tool lists and static handoffs.


eng/skill-validator/tests/Check/PluginMcpManifestTests.cs
```diff
@@ -266,6 +298,28 @@ public void DotnetMsbuildPluginExposesBinlogFromEveryManifest()
...
+                continue;
+
+            var plugin = PluginDiscovery.ParsePluginJson(rootManifest);
+            Assert.NotNull(plugin);
+
+            var result = PluginProfiler.ValidatePlugin(plugin);
+            Assert.DoesNotContain(
+                result.Errors,
+                error => error.Contains(".codex-plugin/plugin.json", StringComparison.Ordinal));
+        }
```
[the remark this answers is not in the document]
Evangelink: Accepted and addressed in c343fd9. Added a pinned Codex 0.154.0 smoke workflow that installs all marketplace plugins in an isolated `CODEX_HOME`, verifies skill and MCP discovery, generates a binlog, and calls `binlog_overview` without credentials.

eng/skill-validator/src/Check/PluginProfiler.cs
```diff
@@ -98,6 +117,59 @@ public static PluginCheckResult ValidatePlugin(PluginInfo plugin)
...
+        {
+            if (server.Value.ValueKind != System.Text.Json.JsonValueKind.Object)
+            {
+                errors.Add($"{relativePath} MCP server '{server.Name}' must be an object.");
+                continue;
+            }
+
+            if (server.Value.TryGetProperty("tools", out var tools) &&
+                tools.ValueKind != System.Text.Json.JsonValueKind.Object)
+            {
```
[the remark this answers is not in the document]
Evangelink: Good catch - fixed in c343fd9. Per-tool entries now validate object shape, supported keys, approval modes, and positive integer output token limits, with regression coverage in `CodexManifestWithInvalidPerToolSettingsErrors`.

eng/codex/codex-plugin-smoke.mjs
```diff
@@ -0,0 +1,338 @@
...
+  DOTNET_GENERATE_ASPNET_CERTIFICATE: "false",
+  DOTNET_NOLOGO: "1",
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: "1",
+  NUGET_PACKAGES: nugetPackages,
+};
+delete environment.CODEX_API_KEY;
+delete environment.OPENAI_API_KEY;
+
+try {
+  rmSync(stateRoot, { recursive: true, force: true });
```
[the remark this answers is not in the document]
Evangelink: Accepted and fixed in db03227. `CODEX_SMOKE_HOME` is now treated only as a parent for a fixed `codex-plugin-smoke` child, so cleanup cannot recursively delete the supplied directory; missing path arguments now also fail explicitly.

Evangelink:
Addressed the latest non-inline review follow-ups in `db03227`:

- Codex compatibility-manifest fields now validate their runtime value kinds, and repository-required `name`/`skills` declarations and skill paths are checked.
- Full MCP server compatibility remains intentionally delegated to the pinned native Codex smoke lane rather than duplicating Codex's evolving 30-field server parser in `skill-validator`; static validation still covers the known incompatible `tools` shape and per-tool settings.
- The README now distinguishes application-level OpenAI Agents SDK handoffs from Copilot UI handoffs and Codex plugin/custom-agent packaging.

eng/skill-validator/src/Check/PluginProfiler.cs
```diff
@@ -98,6 +128,290 @@ public static PluginCheckResult ValidatePlugin(PluginInfo plugin)
...
+
+    private static void ValidateCodexHooks(string relativePath, JsonElement hooks, List<string> errors)
+    {
+        if (hooks.ValueKind is JsonValueKind.String or JsonValueKind.Object)
+            return;
+
+        if (hooks.ValueKind != JsonValueKind.Array ||
+            hooks.EnumerateArray().Any(item => item.ValueKind is not (JsonValueKind.String or JsonValueKind.Object)))
+        {
+            errors.Add($"{relativePath} field 'hooks' must be a string, object, or an array of strings or objects.");
```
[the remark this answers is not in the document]
Evangelink: Accepted and fixed in 3403cf9. Hooks arrays must now be homogeneous string paths or homogeneous inline objects, with a mixed-array regression case; `defaultPrompt` already delegates to the string-or-string-array validator, which rejects mixed arrays.

eng/codex/codex-plugin-smoke.mjs
```diff
@@ -0,0 +1,347 @@
...
+      ),
+      `Codex did not discover ${expectedSkill}`,
+    );
+
+    const thread = await request("thread/start", {
+      cwd: repositoryRoot,
+      ephemeral: true,
+      model: "qwen3-coder",
+      modelProvider: "ollama",
+    });
```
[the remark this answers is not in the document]
Evangelink: Accepted and fixed in 3403cf9. The smoke thread now uses the clean Codex installation's default provider; the direct MCP call does not require a model turn, and the credential-free smoke test still completes successfully.

JanKrivanek (commented):
The overall direction looks good. Please explore and strongly consider using the pinned Codex runtime for compatibility checks instead of reimplementing its parser. A second, hand-written parser is a maintenance liability and can drift from Codex over time, as the mismatches below show. Keep repository-specific policy checks in the validator and strengthen native coverage across all plugins.

eng/codex/codex-plugin-smoke.mjs
```diff
@@ -0,0 +1,345 @@
...
+    const discoveredSkills = skills.data.flatMap((entry) => entry.skills);
+    assert.ok(
+      discoveredSkills.some(
+        (skill) =>
+          skill.name === expectedSkill &&
+          skill.pluginId === `${pluginName}@${marketplaceName}`,
+      ),
+      `Codex did not discover ${expectedSkill}`,
+    );
```
JanKrivanek, line 264: **P2 - Verify skill discovery for every installed plugin.**

This only asserts discovery of `dotnet-msbuild:binlog-failure-analysis`. Installing every marketplace plugin does not establish that their skills are available: Codex 0.154.0 successfully installs and enables a plugin whose declared skills directory is empty, then returns no skills and no discovery errors. Another plugin can therefore lose its entire skill surface while this smoke test stays green.

Please compare the discovered skills against an expected inventory for every installed plugin, and fail on discovery errors as well. Checking only for errors or successful installation is not sufficient.

eng/skill-validator/src/Check/PluginProfiler.cs
```diff
@@ -98,6 +128,295 @@ public static PluginCheckResult ValidatePlugin(PluginInfo plugin)
...
+
+        foreach (var path in paths)
+        {
+            if (string.IsNullOrWhiteSpace(path) || !path.StartsWith("./", StringComparison.Ordinal))
+            {
+                errors.Add($"{relativePath} field '{field}' path '{path}' must start with './'.");
+                continue;
+            }
+
+            if (!PluginDiscovery.TryGetSafeSubdirectory(pluginDirectory, path, out var resolved, out var pathError))
```
JanKrivanek, line 301: **P2 - Reject paths that Codex silently ignores.**

The prefix and normalized-containment checks do not match Codex's lexical path rules. Codex 0.154.0 rejects bare `./` and any `..` path component, even when normalization stays inside the plugin. With skills in a custom `packs` directory, both `./` and `./packs/../packs/` passed this validator without errors but exposed zero skills in native Codex.

Please reject bare `./` and parent-directory components before normalizing the path, with regression coverage for these cases, or use native Codex to verify these compatibility rules.

eng/skill-validator/src/Check/PluginProfiler.cs
```diff
@@ -98,6 +128,295 @@ public static PluginCheckResult ValidatePlugin(PluginInfo plugin)
...
+                continue;
+            }
+
+            foreach (var setting in tool.Value.EnumerateObject())
+            {
+                switch (setting.Name)
+                {
+                    case "approval_mode":
+                        if (setting.Value.ValueKind != JsonValueKind.String ||
+                            !CodexMcpToolApprovalModes.Contains(setting.Value.GetString()!))
```
JanKrivanek: **P2 - Accept null for optional MCP tool settings.**

This rejects `approval_mode: null`, and the `output_token_limit` check below similarly rejects null. Both fields are optional in the pinned Codex parser, where explicit null behaves like omission. Codex 0.154.0 installs this configuration and discovers its enabled MCP server, while the new validator reports blocking errors.

Please allow null for these optional settings while retaining validation of non-null values, and update the new test that expects a null approval mode to be invalid.

eng/codex/codex-plugin-smoke.mjs
Evangelink: Accepted and fixed in e38b9c0. The smoke test now derives the expected inventory for every marketplace plugin, fails on any `skills/list` discovery error, and requires the discovered plugin skill set to match all 98 repository skills exactly.

eng/skill-validator/src/Check/PluginProfiler.cs
Evangelink: Accepted and fixed in e38b9c0. Codex manifest paths now reject bare `./` and any lexical `..` component before normalization, with regression coverage for both examples.

eng/skill-validator/src/Check/PluginProfiler.cs
Evangelink: Accepted and fixed in e38b9c0. Explicit null is now treated like omission for both optional per-tool settings, while non-null approval modes and token limits retain strict validation.

eng/skill-validator/src/Check/PluginProfiler.cs
```diff
@@ -98,6 +128,310 @@ public static PluginCheckResult ValidatePlugin(PluginInfo plugin)
...
+                    "Codex ignores unknown compatibility-manifest fields; keep host-specific components out of this manifest.");
+                continue;
+            }
+
+            switch (property.Name)
+            {
+                case "name":
+                case "version":
+                case "description":
+                case "apps":
```
AbhitejJohn, line 194: Could we route `apps` and the path forms of `hooks` through the same path validation used for `skills` and `commands`? Codex 0.154.0 requires these paths to start with `./` and rejects parent traversal, but the new validator checks only their JSON kinds. I reproduced this with `"apps":"apps.json"` and `"hooks":["../outside-hooks.json"]`; `skill-validator check` still exited 0 with `All checks passed`, while Codex silently drops both paths. This leaves the exact declared-but-not-loaded failure mode that this validation is meant to prevent.

README.md
```diff
@@ -2,18 +2,29 @@
...
 
 > [!TIP]
 > **Compare skill value:** [Open the Skill Value dashboard](https://dotnet.github.io/skills/)
 > See token use, elapsed time, activation, and not-passed rates by plugin, skill,
 > executor model, and judge model.
 
+Plugin support is component-specific:
+
+- Skills and MCP servers are the portable component types defined by
+  [Agent Plugins 1.0](https://agent-plugins.org/specification/1.0.0).
```
AbhitejJohn: This versioned specification URL currently returns HTTP 404; the canonical `https://agent-plugins.org/specification` URL returns 200. Could we update this occurrence and the same link in `CONTRIBUTING.md`?

.github/workflows/codex-plugin-smoke.yml
```diff
@@ -0,0 +1,65 @@
...
+      - "plugins/*/plugin.json"
+      - "plugins/*/.codex-plugin/plugin.json"
+      - "plugins/*/.mcp.json"
+      - "plugins/*/mcp.json"
+      - "plugins/*/skills/**"
+      - "eng/codex/**"
+      - ".github/workflows/codex-plugin-smoke.yml"
+  workflow_dispatch:
+
+permissions:
```
AbhitejJohn, line 27: Nice end-to-end coverage of the local integration path. Could we add the same PR-scoped `concurrency` group with `cancel-in-progress: true` that the other validation workflows use? This job can run for up to 15 minutes and triggers on most skill changes, so repeated pushes otherwise keep obsolete Codex install/build runs active.

.github/workflows/codex-plugin-smoke.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          mkdir -p artifacts/codex-cli
+          curl -fsSLo artifacts/codex-cli/codex.tar.gz \
+            "https://github.com/openai/codex/releases/download/rust-v${CODEX_VERSION}/codex-x86_64-unknown-linux-musl.tar.gz"
+          echo "${CODEX_SHA256}  artifacts/codex-cli/codex.tar.gz" | sha256sum -c -
+          tar -xzf artifacts/codex-cli/codex.tar.gz -C artifacts/codex-cli
+          mv artifacts/codex-cli/codex-x86_64-unknown-linux-musl artifacts/codex-cli/codex
+          chmod +x artifacts/codex-cli/codex
+
+      - name: Run Codex plugin smoke test
```
AbhitejJohn: Could we call out the scope of this smoke test explicitly? It removes `CODEX_API_KEY` and `OPENAI_API_KEY`, then calls `mcpServer/tool/call` directly. That gives strong end-to-end coverage of plugin installation, discovery, and MCP transport, but it does not cover authentication, model inference, prompt-based skill selection, or model-driven tool selection. The current docs only say "a real MCP tool call," which can read as a full Codex conversation test.

CONTRIBUTING.md
```diff
@@ -52,17 +52,23 @@ If your skill does not fit any existing plugin, consider creating a new one.
...
+   `lspServers` into it.
+2. Add a matching entry in `.github/plugin/marketplace.json`, `.claude-plugin/marketplace.json`, `.cursor-plugin/marketplace.json`, and `.agents/plugins/marketplace.json`. Keep names and `plugins[].source` values consistent across all marketplace manifests. Descriptions may be host-specific when a capability is unavailable there; for example, the Codex marketplace must not advertise `.agent.md` agents or LSP integration.
    Also add a `plugins/<plugin-name>/version.json` (copy an existing one) so the plugin participates in automated versioning. Start its `plugin.json` version at `0.1.0`.
 3. Add a CODEOWNERS entry for the new plugin and its tests (see [Code ownership](#code-ownership)).
 4. Add the plugin to the **What's Included** table in the root `README.md`.
 5. Create a `tests/<plugin-name>/` directory for skill tests.
 
 See existing plugins for the expected format.
 
+This is the repository's current legacy multi-host layout, not an Agent Plugins 1.0 package layout.
```
AbhitejJohn: Could we make the intended lifetime of this compatibility layer more explicit? An Agent Plugins 1.0 migration would move portable skills and MCP configuration to the schema-qualified root manifest and fixed component paths, removing most duplicated Codex-manifest and parity validation. A smaller Codex integration smoke would still be useful to verify installation and discovery. Calling this out as a legacy-layout bridge would help prevent the custom validation from becoming a permanent parallel format.

eng/skill-validator/src/Check/PluginProfiler.cs
Evangelink: Good catch — fixed in 38f0a915e. `apps` and path-based `hooks` now share the Codex path validator, with regression cases for both reproduced invalid paths.

README.md
Evangelink: Updated in 38f0a915e. Both references now use the canonical specification URL that returns the published Agent Plugins 1.0 specification.

.github/workflows/codex-plugin-smoke.yml
Evangelink: Added in 38f0a915e. The smoke workflow now uses PR/ref-scoped concurrency with `cancel-in-progress: true`.

.github/workflows/codex-plugin-smoke.yml
Evangelink: Clarified in 38f0a915e. The workflow and contributor docs now state that the credential-free direct MCP call covers installation, discovery, and transport, but does not exercise authentication, inference, or model-driven selection.

CONTRIBUTING.md
Evangelink: Clarified in 38f0a915e. The contributor guide now describes this as a temporary legacy-layout bridge, with Agent Plugins 1.0 removing most custom parity validation while retaining a smaller Codex integration smoke.

JanKrivanek (approved):
Thanks, the updated changes address all three inline findings. Approving with the expectation that #1006 will cover the Agent Plugins 1.0 migration and remove or simplify this temporary Codex compatibility layer. The linked issue now explicitly captures that cleanup. The follow-up should favor native Codex runtime checks over maintaining a second parser that can drift over time, while preserving full skill-discovery and MCP coverage.

PR #1177, opened 2026-09-15
Unify the test engineering agent entry point

### LINKED ISSUE #1119: Confusion with code-testing-agent
Something in the metadata seems to be confusing agents into thinking the code-testing-agent skill is the name of another agent they can call.
I have had this happen a few times in VS Code, where instead of loading the skill, they try to run it as an agent instead.
It seems to be particularly bad with GPT-5.6 Luna, but other models also have the same problem sometimes.
Heck, when I first installed the plugin, I thought that was an agent name and was confused as to why I couldn't select it in the agent list.

Evangelink:
## Summary

- replace the generation-only public surface with one `test-engineer` agent for generating, repairing, running, auditing, and improving tests
- make `test-quality-auditor`, `testability-migration`, and the existing pipeline workers internal delegates of that agent
- rename the implicit model-routed skill from `code-testing-agent` to `code-testing`, while retaining the old name as a hidden compatibility alias
- route post-migration quality handoffs through `test-engineer`
- add agent scenarios for focused generation, repairing weak tests, and read-only quality audits

## Design for discussion

The intended entry points are now deliberately different:

- users explicitly select `test-engineer`
- models implicitly activate `code-testing` for requests that require test changes
- broad or multi-stage skill requests invoke `test-engineer`; focused requests can remain inline
- specialist agents remain callable by the orchestrator but do not compete in the agent picker

The compatibility alias keeps explicit uses of the former `code-testing-agent` skill name working without exposing that agent-shaped name to model routing.

This overlaps conceptually with #1165: the new `agent.test-engineer` spec is included here, while first-class execution and reporting for custom-agent evals remains in that PR.

## Validation

- `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin plugins/dotnet-test`
- `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -- check --plugin plugins/dotnet-test-migration`
- static assertion that `test-engineer` is the only user-invocable agent in `dotnet-test`
- `git diff --check`

Closes #1119


plugins/dotnet-test/agents/test-engineer.agent.md
```diff
@@ -1,12 +1,18 @@
...
+  Primary test engineering agent for generating, repairing, running, auditing,
+  and improving tests across supported languages. Handles focused work
+  directly; coordinates broad generation through specialist workers, quality
+  assessment through test-quality-auditor, and explicit .NET testability
+  refactors through testability-migration. Use for end-to-end test work. Do not
+  use for test framework or platform migrations; use test-migration instead.
+name: test-engineer
+user-invocable: true
+disable-model-invocation: false
 tools: ["agent", "skill", "read", "search", "edit", "execute", "Task", "Skill", "Read", "Glob", "Grep", "Edit", "Write", "Bash", "read_file", "replace", "write_file", "glob", "grep_search", "run_shell_command"]
```
JanKrivanek, line 12: Let's get rid of tools - as without the explicit specification it's not restricting - which should be more powerfull

JanKrivanek (approved):
Overall looks good!

I'd be very curious about evals - especially from MSBench

Evangelink:
@AbhitejJohn this is most likely a "breaking change", not sure if we want to version it somehow or we are cool with just moving on.

AbhitejJohn:
@Evangelink : Lets update the version, that would be a clean change. You could use `/version-bump` on this PR to update when you are ready.

PR #1188, opened 2026-09-18
Track full test-generation scope and safe .NET project scaffolding

0101:
## Summary
- track each in-scope source file as tested or explicitly deferred during broad test generation
- keep .NET generation on one test project and place new SDK test projects outside production default globs
- honor repository-pinned `MSTest.Sdk`/Microsoft.Testing.Platform scaffolding while keeping the existing `code-testing-agent` entry point and builder/tester agents

## Validation
- `dotnet run --project eng/skill-validator/src/SkillValidator.csproj -p:CopilotSkipCliDownload=true -- check --plugin plugins/dotnet-test --allowed-external-deps eng/allowed-external-deps.txt` (22 skills, 10 agents, 1 plugin)

Existing `dotnet-test` evals are unchanged.


Evangelink (commented):
/evaluate

PR #1195, opened 2026-09-21
Add preference-based skill value dashboard

webreidi:
## Summary

Add a Skill Value dashboard that combines side-by-side task results with token, execution-time, and reliability diagnostics.

Every headline value and suggestion comes from the same latest run and commit: preference result, activation contract, activation rate, token/time cost, reliability counts, and timeout state. Earlier runs are shown only as separate history in the drilldown and exclude the headline run, so different skill revisions are never blended into one recommendation.

Preference wins and regressions remain visible when cost telemetry is missing or has fewer than five paired observations; only cost claims are withheld in that case. Missing token/time telemetry normalized to zero is excluded from cost averages and shown as unavailable rather than as free execution.

Generated dashboard data is not checked in. Production continues to source data from `dashboard-eval-data`, avoiding an incomplete and oversized local snapshot.

## Related issue

N/A

## Validation

- `node --test eng\dashboard\skill-value.test.js` (5 tests passing)
- Regression coverage for latest-run-only headlines, low-sample cost, failed activation contracts, missing telemetry, and legacy reliability data
- PowerShell syntax validation for `eng\dashboard\generate-benchmark-data.ps1`
- Viewed the results in the offline dashboard

## Checklist


webreidi:
Note: this drastically changes the "Value" for almost all the skills.

eng/dashboard/skill-value.js
```diff
@@ -131,7 +255,13 @@
       const treat = newArm();
       let actExpected = 0, actFired = 0;
       let passTotal = 0, baseFail = 0, treatFail = 0;
+      let bothPass = 0, bothFail = 0, baselineOnlyPass = 0, treatmentOnlyPass = 0;
       let hasPass = false;
+      // Recommendation evidence is strictly the newest run's value, including
+      // null. Falling through to an older non-null result would let a legacy or
+      // incomplete latest run retain a stale "worth installing" assessment.
+      const preference = runs.length > 0 ? (runs[0].s.preference ?? null) : null;
```
Evangelink: The preference comes only from `runs[0]`, but activation, cost, and reliability below are summed across all 20 retained runs without considering `commit.id`. That lets one headline combine different skill revisions. I reproduced a latest run with 0/5 activations and 2× cost still rendering “Worth installing … 43% fewer tokens and 43% faster” because older runs dominated. Please base every input to the recommendation on the same run/commit, or present the historical diagnostics separately.

eng/dashboard/skill-value.js
```diff
@@ -197,47 +336,129 @@
...
+    const initialAssessment = valueAssessment(row);
+    if (initialAssessment.status === 'activation-contract-failed') {
+      const violated = row.activationContract.violated || 0;
+      return {
+        text: `<b>Guidance withheld</b> — the skill activated in ${violated} task(s) where it was expected to stay off.`,
+        cls: 'sv-value negative',
+        status: initialAssessment.status,
+      };
+    }
+    if (!gated(pairedN) && initialAssessment.status !== 'preference-only') {
```
Evangelink: This metric gate overrides the authoritative preference result for every status except `preference-only`. A conclusive win with no cost data says “Looks helpful,” the same win with 1–4 metric samples becomes “Insufficient,” and a conclusive regression without cost data is hidden too. Apply the sample floor only to cost claims; preserve preference guidance while marking cost unavailable or under-sampled.

eng/dashboard/data/components.json
```diff
@@ -0,0 +1,18 @@
...
+  "dotnet-msbuild",
+  "dotnet-nuget",
+  "dotnet-template-engine",
+  "dotnet-test-migration",
+  "dotnet-test",
+  "dotnet-upgrade",
+  "dotnet",
+  "dotnet11",
+  "skill-value"
+]
```
Evangelink: This checked-in manifest lists 15 plugin tabs, but none of their `<plugin>.json` files are tracked, so each tab fails in a fresh checkout. Production does not use this source snapshot; deployment copies data from `dashboard-eval-data`. Since `skill-value.json` is 2.59 MB and over 99% of this diff, either remove these generated files or replace them with a small, self-consistent local fixture.

Evangelink (commented):
The preference-based direction is valuable. I left three non-blocking comments on evidence consistency, low-sample behavior, and the checked-in dashboard fixture.

JanKrivanek (approved):

JanKrivanek (approved):

PR #1200, opened 2026-09-22
Pin DevOps Health Copilot runtime

AbhitejJohn:
## Summary

Pin the Copilot CLI runtime to `1.0.80` in the DevOps Health check, groom, and investigation workflows. This prevents those workflows from resolving to Copilot CLI `1.0.85`, which cannot list tools through the currently generated MCP Gateway `v0.4.18`.

This is a narrow compatibility fix for the failures recorded in #834 and the stale publication state visible in #695. It does not change permissions, safe-output policy, MCP tool grants, network policy, action pins, or the Gateway version.

## Why

#1173 fixed the earlier repository-controlled failures from #1166 and #1167: cache-memory miss handling, groomer output safety, worker limits, issue-695 validation, transactional publication, investigation controls, and runtime consistency. It did not claim to fix the provider authentication failures tracked in #1162 and #1172.

Direct post-merge runs `35682662360` and `35694535133` still produced empty safe output. The first run log states `No explicit Copilot CLI version requested`, downloads the live compatibility matrix, rejects cached `1.0.80`, and installs `1.0.85`. Copilot CLI then requests MCP protocol `2025-11-25`; MCP Gateway `v0.4.18` rejects tool listing with error `-32022` because that protocol version is unsupported.

The workflows still completed with a green conclusion because the agent reported the unavailable tools and exited without publishing safe output. This also explains why #695 still shows old health data: publication stopped after the runtime compatibility failure rather than receiving a new dashboard result.

## What changed

Each source workflow now declares:

```yaml
engine:
  id: copilot
  version: 1.0.80
```

The three generated locks now pass `1.0.80` directly to `install_copilot_cli.sh` and expose `GH_AW_ENGINE_VERSION: "1.0.80"`. The compiler remains `gh-aw v0.88.7`, and MCP Gateway remains `v0.4.18`.

## Impact

The check, groom, and investigation workflows use the runtime version that their current generated stack supports. The lock body hashes, permissions, tools, network controls, safe-output policy, action pins, and container pins are unchanged.

## Validation

- `gh aw compile devops-health-check devops-health-groom devops-health-investigate --strict --no-check-update` — 3 workflows compiled successfully; only the two pre-existing fixed-schedule warnings remain.
- `python eng/evaluation/test_token_failover.py` — 43 tests passed, 3 skipped.
- Explicit source/lock assertions — all three sources declare `1.0.80`; all three locks call `install_copilot_cli.sh 1.0.80`; no unpinned install invocation remains.
- Pinned `actionlint` `v1.7.7` — clean for repository-maintained workflow YAML.
- `git diff --check` — clean.

## Risk and limits

Pinning `1.0.80` can defer future Copilot CLI fixes. The pin should be advanced together with a compatible MCP Gateway/compiler stack after end-to-end validation. This PR does not address unrelated provider HTTP 401 incidents from #1162 or #1172, and it does not upgrade `gh-aw` because the newer tested compiler path has an upstream Windows strict-compilation defect.

## Rollback

Revert commit `3e04676cb` to restore live runtime resolution in all three workflows.

_(Copilot prepared this draft PR on Abhitej's behalf.)_


JanKrivanek (approved):
We should ideally track item for removing this pin once the issue with listing tools is resolved on gh-aw side

PR #1205, opened 2026-09-23
Speed up dashboard model filtering

Evangelink:
## Summary

Model checkbox changes on the Skills Evaluation Dashboard were destroying and recreating every Chart.js instance. On the deployed `dotnet-test` view, one click rebuilt 440 charts from a roughly 19.8 MB data payload.

This change keeps chart instances alive and filters their cached labels and dataset arrays in place using no-animation updates. Rapid checkbox changes are coalesced into one animation-frame refresh, while tooltips and issue legend notes remain synchronized with the selected models. A dependency-free regression test reproduces the deployed 440-chart shape and verifies that filtering does not construct or destroy charts.

## Related issue

N/A

## Validation

- `node --check eng/dashboard/dashboard.js`
- `node --test eng/vally-adapter/*.test.mjs eng/dashboard/*.test.js` - 99 tests passed
- `git diff --check`

## Checklist

- [ ] I searched existing issues and pull requests to avoid duplicates.


eng/dashboard/dashboard.js
```diff
@@ -831,15 +897,10 @@
...
-            if (timeOF === 'high' || tokenOF === 'high') legendFlags.overfittingHigh = true;
-            else legendFlags.overfittingModerate = true;
-          }
-          const timeIssues = (timeNA ? 1 : 0) + (timeTO ? 1 : 0) + (timeOF ? 1 : 0);
-          const tokenIssues = (tokenNA ? 1 : 0) + (tokenTO ? 1 : 0) + (tokenOF ? 1 : 0);
-          if (timeIssues > 1 || tokenIssues > 1) legendFlags.multiIssue = true;
+          recordIssueFlags(legendFlags, timeNA, timeTO, timeOF);
+          recordIssueFlags(legendFlags, tokenNA, tokenTO, tokenOF);
+          recordIssueFlags(entryLegendFlags[entryIndex], timeNA, timeTO, timeOF);
+          recordIssueFlags(entryLegendFlags[entryIndex], tokenNA, tokenTO, tokenOF);
```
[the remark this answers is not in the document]
Evangelink: Good catch. Fixed in 9d5e30483 by aggregating time and token severity per efficiency entry so high suppresses moderate, while retaining per-metric multi-issue detection. Added regression coverage for mixed high/moderate overfitting with multiple issues.

eng/dashboard/dashboard.js
```diff
@@ -1215,6 +1320,26 @@
...
+        const model = (entry && entry.model) ? entry.model : 'unknown';
+        if (activeModels.has(model)) indexes.push(index);
+      });
+      visibleEntries = indexes.map(index => entries[index]);
+      chart.data.labels = indexes.map(index => sourceLabels[index]);
+      chart.data.datasets.forEach((dataset, index) => {
+        applyDatasetIndexes(dataset, datasetSources[index], indexes);
+        dataset.hidden = !activeModels.has(datasetModels[index]);
+      });
+      const visibleDatasetCount = datasetModels.filter(model => activeModels.has(model)).length;
```
JanKrivanek, line 1349: **P2: Filter legend items by active model**

This enables the legend based on the selected dataset count, but Chart.js's default label generator still includes all retained datasets, including those with `dataset.hidden = true`. With Chart.js 4.4.7 and 10 models, selecting just one model produces two visible series but a 20-item legend. Filtering therefore brings back the crowded legend that `MAX_INLINE_LEGEND_SERIES` is meant to prevent.

Please filter the generated legend items by active model as well, while retaining the cached datasets. A regression case that crosses the inline-legend threshold would catch this; the current `FakeChart` does not execute legend generation.

eng/dashboard/dashboard.js
```diff
@@ -1020,14 +1087,48 @@
...
         liveCharts.push(effChart);
       });
+      }
+      chartsInitialized = true;
+    } else {
+      liveCharts.forEach(chart => {
+        if (chart && chart.applyModelFilter) chart.applyModelFilter(activeModels);
+      });
     }
     } // end draw()
```
JanKrivanek, line 1123: **P2: Hide charts with no measurements for selected models**

Updating a fixed set of chart instances leaves irrelevant charts visible when models have different test coverage. For example, if model A only covers test X and model B only covers test Y, deselecting A previously left only Y's quality and efficiency charts. This version leaves all four containers, with two blank X charts. On a large dashboard, filtering can leave many empty charts to scroll past.

Please keep the instances cached but hide containers whose test has no measurements in the selected models, and restore them when those models are reselected. A fixture with different tests per model would cover this regression.

JanKrivanek (approved):
LGTM

eng/dashboard/dashboard.js
Evangelink: Good catch. Fixed in 97ea3629d by filtering generated legend items for model-hidden datasets while preserving user-hidden active series, with a 10-model regression test that crosses the inline legend threshold.

eng/dashboard/dashboard.js
Evangelink: Agreed. Fixed in 97ea3629d by hiding cached quality and efficiency chart containers when the selected models have no measurements and restoring them when coverage returns, with a split-coverage regression fixture.

eng/dashboard/dashboard-filtering.test.js
```diff
@@ -0,0 +1,412 @@
...
+  await renderDashboard(t, pluginData, 2);
+
+  assert.equal(FakeChart.instances.length, 2);
+  const efficiencyChart = FakeChart.instances.find(chart =>
+    chart.data.datasets.some(dataset => dataset.label === 'Isolated Time (s)')
+  );
+  assert.ok(efficiencyChart);
+  const notes = efficiencyChart.canvas.parentElement
+    .querySelectorAll('.not-activated-legend')
+    .map(note => note.innerHTML);
```
AbhitejJohn, line 320: Nice coverage for the severity precedence. Could we also toggle an issue-bearing model here and assert that the notes shrink, then invoke the tooltip `afterTitle` callback for the surviving entry? Those are the two synchronization guarantees called out in the PR description, but the current filter tests use entries with no issue flags and do not exercise the tooltip remap.

AbhitejJohn (approved):
Thanks for addressing the filtering synchronization cases. The current implementation looks good to me: chart instances stay cached, hidden models no longer crowd the legend, Chart.js visibility metadata stays aligned with the model checkboxes, and charts without selected measurements are hidden and restored correctly.

I verified the live `dotnet-test` payload at 19.83 MB and 440 charts. I also ran the stated validation command at the latest head: all **102 tests passed**, with `node --check` and `git diff --check` clean. The PR body still says 99 tests, so please update that count. I left one non-blocking test suggestion for the tooltip and issue-note synchronization paths.

**Verdict: approve with nits.**

eng/dashboard/dashboard-filtering.test.js
Evangelink: Added in dacb2d735. The efficiency fixture now filters out the high/multi-issue model, verifies the notes shrink to the surviving moderate issue, and invokes `afterTitle` to confirm the tooltip remaps to the surviving model entry.

AbhitejJohn (approved):

